PR #153, opened 2026-04-20
Pin KTC divergence and harden canonical single-curve invariant

jasonleetucker-code:
## Summary

Partial fix for two of the four mathematical-soundness red flags flagged during the value-accuracy review. Scoped so production values do not change.

### Red flag #3 — no KTC reconciliation test (FIXED)

`tests/canonical/test_ktc_reconciliation.py` loads the live KTC snapshot from `CSVs/site_raw/ktc.csv`, sorts players by value, and pins our Hill curve's divergence at ranks 1/5/12/24/50/100/150/200/300/400.

Current baseline (2026-04-20):

| Rank | KTC | Ours | %Δ |
|---:|---:|---:|---:|
| 1 | 9998 | 9999 | 0.0% |
| 12 | 7794 | 8459 | +8.5% |
| 50 | 5151 | 4967 | −3.6% |
| 100 | 3619 | 3055 | **−15.6%** |
| 200 | 2422 | 1648 | **−32.0%** |
| 400 | 1008 | 815 | −19.1% |

Tolerance ±2 pp on each pin. Curve-shape invariants also pinned: rank 1 matches by construction; we sit above KTC in the early mid-top; tail past rank 100 is compressed well below KTC until the curve is intentionally re-fit.

### Red flag #2 — two-curve architecture (HARDENED)

Verified the canonical Hill engine already bypasses `calibration.py` via the `if not use_canonical_engine:` gate in `scripts/canonical_build.py`, and `valuation_result_to_asset_dicts` emits `blended_value == calibrated_value == display_value`. The bypass was a single fragile boolean — now defended two ways:

- `calibrate_canonical_values` raises `RuntimeError` if handed any asset tagged `_pick_calibration_source == "canonical_pipeline"`, preventing silent double-calibration if the gate is ever bypassed.
- `tests/canonical/test_canonical_single_curve.py` pins both sides: canonical emit equality, and the guard rejecting canonical input while still accepting pure legacy assets.

### Not included

- Red flag #1 (tier cliffs + volatility adjustments excluded from `display_value`) — this is deliberate per `player_valuation.py:547-550` and is a product decision, not a bug. Changing it would shift every user-facing value and require re-fitting the Hill constants.
- Red flag #4 (joint hyperparameter optimization) — research effort, separate session.

## Test plan

- [x] `python3 -m pytest tests/canonical/ -q` → 243 passed
- [x] New tests fail loudly when `HILL_MIDPOINT` / `HILL_SLOPE` are mutated
- [x] New guard fires on canonical-tagged input; pure legacy assets still calibrate
- [x] No production value changes — legacy engine path untouched
- [ ] Re-baseline `PINNED_DELTAS` when Hill curve is intentionally re-fit via `scripts/fit_hill_curve_from_market.py`

https://claude.ai/code/session_011hat7VDYqxbR6DKNGKtdKH

### DIFF
--- a/src/canonical/calibration.py
+++ b/src/canonical/calibration.py
@@ -237,6 +237,13 @@ def calibrate_canonical_values(
     The knee can vary per universe (IDP uses a higher knee than offense).
     Picks are calibrated separately using the legacy pick value curve.
 
+    This function is for the LEGACY engine only.  The canonical 6-step
+    Hill-curve engine produces display-scaled values in a single pass
+    (see ``src/canonical/player_valuation.py``) and must not be
+    re-calibrated here — doing so would stack a second curve on top of
+    the Hill output.  Assets tagged by the canonical pipeline with
+    ``_pick_calibration_source == "canonical_pipeline"`` are rejected.
+
     Args:
         assets: List of canonical asset dicts.
         universe_scales: Optional override for per-universe max scales.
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
CI failure (`Validate PR` → `Run unit tests`) is not caused by this PR.

**Failing test:** `tests/api/test_pick_refinement.py::TestPlayerRankingsUnchanged::test_known_player_values_unchanged`
```
AssertionError: 'medium' != 'high'
 : Brock Bowers confidenceBucket changed high→medium
```

**Root cause:** the automated data refresh at `3d12aa8` (merged directly to `main` ~1.5h before this branch) changed Brock Bowers' `confidenceBucket` and broke the pinning test.

**Verification:**
- `HEAD~2` (commit before the refresh): test passes
- `HEAD~1` (the refresh itself, pre-dating this PR): test fails
- This PR's commit `5a07847`: no changes to confidence-bucket logic or test fixtures

**Recommended action:** resolve on `main` before merging any open PR — either re-baseline the pinned value or investigate whether the data refresh exposed a real regression (the test name `test_known_player_values_unchanged` suggests it's designed to catch exactly this kind of drift). Intentionally not silently re-baselining here, since erasing the signal is the wrong default for a test like this.

Any PR targeting current `main` will hit the same red.

---

PR #227, opened 2026-04-22
Hide mobile source chips until row is tapped

jasonleetucker-code:
## Why this is a follow-up

#226 flipped the cold-start default for `showSiteCols` to `false`, but that only helps brand-new users. Anyone who already had `showSiteCols: true` stored in localStorage (including the user who filed the original mobile screenshot) still sees the per-row source chip strip rendered beneath every player — which dominates the mobile viewport and pushes the Value column off-screen.

## Change

Gate the mobile chip strip render on `isExpanded` instead of `settings.showSiteCols`. On mobile, the chips now only appear when you **tap a row to reveal them**, matching what the user literally asked for: "click and then be able to see the source columns."

Desktop per-source columns continue to honor `settings.showSiteCols` — behavior there is unchanged.

## Effect, by surface

- **Mobile, cold-start row**: headline columns only (# / Player / Pos / Consensus / Value). No chip strip. Value column fits without horizontal scroll.
- **Mobile, tapped row**: chip strip appears beneath the row alongside the existing source audit panel.
- **Desktop**: unchanged. Per-source columns still controlled by the `showSiteCols` master toggle (now reachable from both the Columns popover on `/rankings` and `/settings`).

## Test plan

- [ ] On mobile (≤768px), verify the chip strip is hidden under every player row on page load regardless of stored `showSiteCols` value.
- [ ] Tap a row and confirm the chip strip (and existing audit panel) render beneath it.
- [ ] Tap the row again to collapse — strip disappears.
- [ ] On desktop, verify `showSiteCols: true` still renders per-source columns inline; toggling from the Columns popover master still works.
- [ ] Confirm the Josh Allen row's Value column is fully visible on a 390px-wide viewport without scrolling.

https://claude.ai/code/session_01T1SY7WNCysDYBzzPwqpBV4

### DIFF
--- a/frontend/app/rankings/page.jsx
+++ b/frontend/app/rankings/page.jsx
@@ -1446,10 +1446,16 @@ export default function RankingsPage() {
                           horizontally and get hidden via `hide-mobile`.
                           Instead of dropping the data entirely, we
                           render a compact flex strip of source chips
-                          below each player row so mobile users see the
-                          same per-source value + rank they'd see on
-                          desktop.  Gated on `showSiteCols` so the
-                          toggle has the same effect on both surfaces.
+                          below each player row — but only when the
+                          user has tapped the row to expand it.
+                          Rendering the strip under every row by
+                          default dominated the mobile viewport and
+                          pushed the headline Value column off-screen;
+                          gating on `isExpanded` makes the source
+                          audit on-demand.  `showSiteCols` still
+                          governs the desktop column render; on mobile
+                          the "click to reveal" pattern is the
+                          default.
 
                           Uses a dedicated `.rankings-mobile-source-row`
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Closing as **superseded**.

Reviewing the diff against current `main`:
- The Columns popover master-toggle UI additions have since landed via #226.
- `showSiteCols: false` default has since landed via #226.
- The only remaining meaningful change is the mobile render gate: `{settings.showSiteCols && (` → `{isExpanded && (` at `frontend/app/rankings/page.jsx` ~line 1462, plus the associated comment refresh.

That last piece is being opened as a fresh small PR rather than carrying this one forward — most of this branch would otherwise try to re-insert lines that already exist on main.

---

PR #401, opened 2026-05-06
feat(te-premium-lab): implied TEP multiplier display + DraftSharks pair

jasonleetucker-code:
## Summary
Two iterations on the TE Premium Lab:

1. **Implied TEP multiplier in the comparison footer** — the user asked "what do we multiply each site's TE values by to get TEP values".  The mean boost row was already that number; this adds a second footer row that re-expresses it as a multiplier (1.25× instead of +25%).  Hover tooltip clarifies the intended use.

2. **DraftSharks as a 4th source pair** — DS's TEP behavior is driven by the league dropdown (WASM worker applies our league's scoring once selected), not a URL change.  Added a ``--capture-baseline`` flag that runs a second scrape pass without applying the league, capturing DS's public baseline (non-TEP) values.  Doubles scrape time but the lab is research-only so that's acceptable.

## Multiplier readout (current top-24 sample)
| Source | Avg boost | Multiplier |
|---|---|---|
| KTC | +24.7% | **1.25×** |
| FantasyPros Fitzmaurice | +24.6% | **1.25×** |
| DynastyNerds | +35.5% | 1.36× |

KTC and Fitz both land at the ~25% the user expected.  DN runs higher because their value scale caps top players, compressing the #1 boost and inflating mid-tier boosts (Bowers SF ≈ 8301 already near their ceiling).

## Sources investigated this round

| Source | Outcome | Note |
|---|---|---|
| DraftSharks | ✅ added (this PR) | ``--capture-baseline`` flag scrapes second pass |
| Flock Fantasy | ❌ doesn't publish | Tried 8 format-param variants; only ``superflex`` returns data |
| DLF | ❌ would need auth probe | All ``-tep-`` URL guesses 403 unauthenticated |

## Operator workflow
For one-shot research scrape including DraftSharks:

\`\`\`
python3 scripts/fetch_draftsharks.py --capture-baseline
\`\`\`

The DS pair will appear in the lab automatically once both ``draftSharksSf.csv`` and ``draftSharksSfBase.csv`` are on disk.

## Test plan
- [x] \`pytest tests/research/test_te_premium.py\` — 50 passed
- [x] \`npm run build\` — all pages under bundle budget
- [x] DS fetcher: existing single-pass behavior unchanged (no flag); ``--capture-baseline`` runs second pass through same authenticated context

### DIFF
--- a/frontend/app/te-premium-lab/page.jsx
+++ b/frontend/app/te-premium-lab/page.jsx
@@ -798,8 +798,8 @@ function MultiSourceComparison({ comparison }) {
           </tbody>
           <tfoot>
             <tr>
-              <td colSpan={4} style={{ fontWeight: 600, paddingTop: 8 }}>
-                Mean across top-{comparison.top_n}
+              <td colSpan={4} style={{ fontWeight: 600, paddingTop: 10 }}>
+                Mean boost across top-{comparison.top_n}
               </td>
               {availableSources.map((s) => {
                 const agg = aggregates[s.key];
@@ -808,7 +808,7 @@ function MultiSourceComparison({ comparison }) {
                     <td
                       key={s.key}
                       className="muted"
-                      style={{ textAlign: "right", paddingTop: 8 }}
+                      style={{ textAlign: "right", paddingTop: 10 }}
                     >
                       —
                     </td>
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
Superseded by the new direction — replacing the sandbox-driven research approach with hardcoded blanket TE multipliers (1.25× non-TEP / 1.10× TEP-native, KTC exempt) and scrapping the TE Premium Lab entirely.  See follow-up PR.

PR #544, opened 2026-07-26
fix(intel): refresh workflow repo context + auth diagnostics

jasonleetucker-code:
Fixes the two failures from the first live intel-refresh run (30191331002) after #534 deployed.

## 1. Failure-tracking step: `fatal: not a git repository`
This workflow never checks out the repo (it only curls the prod host), so `gh label`/`gh issue` had no repository context and the idempotent `intel-stale` issue was never opened. Fixed with `GH_REPO: ${{ github.repository }}` on the step's env (cheapest correct fix — no checkout needed), with a comment explaining why it's load-bearing.

## 2. The 401 — diagnosis + observability
**Code audit verdict: no functional bug in the auth path.**
- Workflow sends `Authorization: Bearer <INTEL_REFRESH_TOKEN secret>` — exactly the shape the server parses (case-insensitive `bearer ` scheme, whitespace split).
- Server reads `INTEL_REFRESH_TOKEN` (fallback `SIGNAL_ALERT_CRON_TOKEN`) at process start; the observed body `"Sign-in or bearer token required."` is the intel endpoint's own message, proving the request traversed nginx + the self-authed allowlist and failed only the bearer check.
- Env plumbing exists: the dynasty systemd unit loads `.env` via `EnvironmentFile=-__APP_DIR__/.env`.

So the live 401 is **environmental** — either the VPS `.env` line never reached the process (e.g. written as `export KEY=...`, which systemd's EnvironmentFile silently ignores, or the service wasn't restarted after the edit) or the Actions secret and VPS value differ. The code previously could not distinguish these; now it can:
- `server.py::_intel_bearer_auth_ok` logs **which branch** rejected a presented bearer — "no token configured server-side (env not loaded?)" vs "presented token does not match (len presented=N, len configured=M)". Never logs token values.
- The workflow's trigger step treats 401 as a **non-retryable configuration error** and prints a concrete checklist: secret/VPS byte parity (watch trailing newlines), plain `KEY=value` .env line, service restart, and the journal pointer.
- `.env.example` now documents `INTEL_REFRESH_TOKEN` (previously only in PR notes), including the systemd EnvironmentFile caveats.

**Next step for ops:** re-dispatch the workflow after this merges — the journal line on the VPS (`journalctl -u dynasty | grep "intel bearer auth rejected"`) will say definitively whether the env var isn't loaded or the values mismatch. No credentials touched in this PR.

## Verification
- Workflow YAML parses; all four shell steps pass `bash -n`; `GH_REPO` asserted present on the failure step.
- `tests/intel/` — 88 passed (the workflow-script tests re-extract and re-exercise the edited heredocs).
- `python -m ruff` (0.6.9, CI pin) format + check clean on touched files.

---

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -83,6 +83,20 @@ LOG_FORMAT=text
 # VAPID_PRIVATE_KEY=
 # VAPID_CONTACT=mailto:ops@example.com
 
+# ── Sharp Tracker intel cron (Phase 5) ───────────────────────────────────
+# Shared secret for the daily intel-refresh GitHub Actions workflow
+# (.github/workflows/intel-refresh.yml).  The workflow POSTs
+# /api/intel/refresh?leagueKey=all with ``Authorization: Bearer <token>``;
+# the server accepts it when the value matches INTEL_REFRESH_TOKEN
+# (falling back to SIGNAL_ALERT_CRON_TOKEN when unset).  Generate with
+# ``openssl rand -hex 32`` and store the SAME value as the repo's
+# INTEL_REFRESH_TOKEN Actions secret.
+#
+# systemd loads this file via the dynasty unit's EnvironmentFile —
+# the line must be plain KEY=value (no ``export``), and the service
+# must restart after edits (the token is read at process start).
+# INTEL_REFRESH_TOKEN=
+
 # ── IDP Calibration Lab ──────────────────────────────────────────────────
 # Controls network access for the historical stats adapter layer at
 # src/idp_calibration/stats_adapter.py.
--- a/.github/workflows/intel-refresh.yml
+++ b/.github/workflows/intel-refresh.yml
@@ -103,6 +103,24 @@ jobs:
               sleep 30
               continue
             fi
+            if [[ "${code}" == "401" ]]; then
+              # Auth failure is a CONFIGURATION problem, not a retry
+              # candidate.  The endpoint accepts
+              # ``Authorization: Bearer <INTEL_REFRESH_TOKEN>`` where
+              # the server reads INTEL_REFRESH_TOKEN (falling back to
+              # SIGNAL_ALERT_CRON_TOKEN) from the .env loaded by the
+              # dynasty systemd unit.  Check, in order:
+              #   1. the INTEL_REFRESH_TOKEN repo secret matches the
+              #      VPS value byte-for-byte (no trailing newline);
+              #   2. the VPS .env line is plain KEY=value (systemd
+              #      EnvironmentFile ignores ``export KEY=...``);
+              #   3. the dynasty service restarted after the .env
+              #      edit (the token is read at process start);
+              #   4. the backend journal — the server logs WHICH
+              #      branch failed (no token configured vs mismatch).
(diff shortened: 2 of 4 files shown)


jasonleetucker-code:

Reviewed commit: `a206d8ce4b7c73d20bf715c19a436812522cf8c3` (base `bd79d3d8`)

Verification performed: parsed the PR workflow YAML, ran `bash -n` on all four step scripts (all pass), exercised the new `_intel_bearer_auth_ok` logic against edge cases (empty/None header, `Bearer` with no token, whitespace-only, tab separator, non-ASCII) in a scratch session, compiled `server.py` at the PR head, and ran `python3 -m pytest tests/intel/ -q` on the PR head — **88 passed**. Confirmed the endpoint returns 401 (not 403) for auth failures (`server.py:10445-10450`), so the new workflow 401 branch matches what the server actually sends, and it sits correctly inside the retry loop after the 409 handler. Confirmed the `GH_REPO` fix is real: the workflow has no checkout step, `gh label create`/`gh issue list` failures were swallowed by `|| true` / `|| echo ""`, and the final `gh issue create` then died without repo context — `GH_REPO` is honored by all three `gh` commands used. The "no token configured" log wording is accurate given the `SIGNAL_ALERT_CRON_TOKEN` fallback at `server.py:10275`. No token *value* can reach the logs in any branch.

### Findings

**P3 — server.py (mismatch warning): configured-secret length is written to the journal.**
`len(INTEL_REFRESH_TOKEN)` is logged on every mismatch. This is deliberate diagnostics, but it is token metadata — and this PR's own 401 checklist and failure-issue flow actively direct operators to consult and share journal output, so that line plausibly ends up pasted into a GitHub issue. `len(presented)` (attacker-supplied) is fine; suggest dropping `len configured=%d` and logging only a length-equality boolean (e.g. `lengths_match=%s`), which preserves the diagnostic ("trailing newline in the secret" shows up as a length mismatch) without disclosing the secret's length.

**P3 — server.py (`_intel_bearer_auth_ok`): unauthenticated log-spam vector.**
Both new `log.warning` calls fire for unauthenticated requests — any client sending `Authorization: Bearer x` to `/api/intel/refresh` on the public host writes a journal line per request, unthrottled. Header content is never logged (no injection), so this is only noise/disk pressure, but worth knowing it exists on an internet-facing route.

**P3 — pre-existing, unchanged by this PR but in the touched function: non-ASCII bearer → 500.**
`hmac.compare_digest(presented, INTEL_REFRESH_TOKEN)` raises `TypeError: comparing strings with non-ASCII characters is not supported` when the presented token contains any byte ≥ 0x80 (Starlette decodes headers as latin-1, so `curl -H "Authorization: Bearer $(printf '\xe9')"` reaches this compare). Result is an unhandled 500 instead of 401. Verified the exception empirically. Existed on `main` before this PR, but since the function is being touched: encoding both sides to bytes (`presented.encode("utf-8", "surrogateescape")` vs `INTEL_REFRESH_TOKEN.encode()`) closes it. Note the cron itself can't hit this (hex tokens), and the workflow's generic non-401 error branch handles a 500 correctly.

No P1/P2 findings. The workflow changes (401 fast-fail, `GH_REPO`), the `.env.example` docs, and the auth-diagnostics branching are all correct as written; comments accurately describe the code they annotate.

---

PR #547, opened 2026-07-26
Identity sweep: recover votes lost to name drift + matching audit tooling

jasonleetucker-code:
## What

Sweep of all 22 registered source CSVs for player votes silently dropped at the `_canonical_match_key` join against the Sleeper player pool (the known case was PFK's "Kenneth Gainwell" vs Sleeper's "Kenny Gainwell", only recovered because that CSV carries sleeper_ids — this finds and fixes every other instance).

1. **New audit tool `scripts/audit_identity_matches.py`** — replays the exact CSV → pool join the contract build performs (same `_parse_source_csv_cached` parser, same `_canonical_match_key` cascade, same `row_groups_by_key` pool construction as `_enrich_from_source_csvs`) for every source in `_SOURCE_CSV_PATHS`; reports per-source match/unmatched counts and fuzzy near-misses (difflib; rapidfuzz used when installed) for triage. Also computes an **alias collision-delta**: any post-alias `canonical_player_key` absorbing more than one pre-alias pool key would mean the alias table merged two real players. Exit 0 always (audit, not a gate).
2. **18 verified alias entries** added to `CANONICAL_NAME_ALIASES` in `src/utils/name_clean.py` (the existing alias mechanism — no new system). Full evidence table in `docs/identity-audit-2026-07.md`; every entry confirmed same-player via position + age + team agreement between the vendor CSV row and the pool row (e.g. "Ahmad Gardner" → "Sauce Gardner" legal name, DB/IND/age 24-24.8 both sides; "Justin Madubuike" → "Nnamdi Madubuike" 2024 player rename still used by DLF IDP).
3. **No normalizer changes** — every near-miss was nickname/legal-name/typo drift, which belongs in the deterministic alias layer; no case was found that `normalize_player_name` should have collapsed but didn't.

## Impact (2026-07-26 board, re-verified post-rebase on current main)

Verified by a full row-by-row diff of two complete `build_api_data_contract` builds (alias table off vs on): **16 player rows gain 24 source votes; zero rows lose anything.**

| Source | CSV-join recovered | Contract votes gained |
|---|---|---|
| idpTradeCalc | +6 | 0 (payload already carried) |
| draftSharksIdp | +8 | +8 |
| draftSharks | +5 | +5 |
| idpShow | +3 | +3 |
| ktc / ktcSfTep | +2 / +2 | 0 (payload already carried) |
| dlfSf / fantasyProsSf | +2 / +2 | +2 / +2 |
| dlfIdp / dynastyNerdsSfTep / fantasyCalc / flockFantasySf | +1 each | +1 each |
| pfkDynasty | id-join → name-join | 0 |

The ktc/ktcSfTep/idpTradeCalc recoveries show no contract delta because the scraper payload already carries those values under the Sleeper spelling — for them the alias hardens the CSV **fallback** path (used when a scrape fails and the CSV persists) and fixes `sourceOriginalRanks` / `sourceNativeValues` / `sourceAudit` stamping that previously missed on the drifted names.

Remaining unmatched rows are categorized in the report: overwhelmingly players genuinely absent from the live pool (retired/FA veterans and non-Sleeper prospects on deep vendor boards — fantasyNavigatorSf's 301 unmatched are almost entirely this class; all 15 of its fuzzy near-misses verified as *different* players and deliberately NOT aliased).

## Collision safety

- Alias collision-delta on the live board: **0 collisions** (checked on every audit run).
- `tests/scripts/test_audit_identity_matches.py` pins the invariant and proves the detector fires when a merge IS present (synthetic pool containing both spellings of an aliased pair), plus an end-to-end regression run of the script against the committed CSVs + export.
- `tests/utils/test_name_clean.py::TestIdentitySweepAliases` pins each of the 18 collapses and adds table hygiene invariants (keys normalize-stable, values fixed-points, no alias chains).

## Re-scope after the 8 merged PRs

Rebased onto current main and fully re-verified: the playerctx `resolve_player`/`id_overrides` mapper and news mention matching are separate lookup surfaces — nothing in this sweep was absorbed; the `_derive_player_row` team/yearsExp additions are orthogonal. All 18 vendor spellings still present in the refreshed CSVs (fantasyCalc's refresh added one more recovery: "Robert Henry Jr."). `src/api/data_contract.py` is **untouched** (the alias table lives in `src/utils/name_clean.py`, so the additive-dict contingency never triggered).

## Noted for follow-up (not done here, out of mandate for `data_contract.py`)

`CSVs/site_raw/dynastyNerdsSfTep.csv` carries a `SleeperId` column that `_SLEEPER_ID_ALIASES` doesn't recognize (`sleeper_id`/`sleeperId`/`sleeper_player_id` only) — one token would give Dynasty Nerds the same ID-grade drift immunity as pfkDynasty.

## Validation

- `python -m pytest tests/ -q -m "not livedata"` → **3066 passed, 280 deselected** (post-rebase)
- `ruff check` + `ruff format --check` clean on all touched Python
- Audit + full contract rebuild run before/after on both the 2026-07-25 and 2026-07-26 boards; per-row stamp diff confirms gains-only

---

### DIFF
--- a/.github/workflows/audit-identity-matches.yml
+++ b/.github/workflows/audit-identity-matches.yml
@@ -1,0 +1,184 @@
+name: Audit Identity Matches
+
+# Guards the CSV -> player-pool identity join against vendor name drift.
+#
+# Why this exists: ``scheduled-refresh.yml`` pushes freshly scraped
+# CSVs and a new export to ``main`` every two hours WITHOUT running
+# pytest, so the alias collision-delta invariant pinned in
+# ``tests/scripts/test_audit_identity_matches.py`` never gets evaluated
+# against the data that actually changed.  A vendor renaming a player
+# into an existing pool name would silently merge two players' votes
+# and nothing would catch it until someone ran the audit by hand —
+# historically "never" (same gap that motivated
+# ``audit-dropped-sources.yml``).
+#
+# This job runs the audit against the refreshed data and fails when the
+# collision-delta check finds anything.  It is deliberately a SEPARATE
+# workflow from the refresh: an advisory data finding must never block
+# the data refresh itself, so the refresh stays green and this job goes
+# red (plus a rolling issue) when the board needs a human.
+#
+# Exit codes (from ``scripts/audit_identity_matches.py``):
+#   0  - no alias-introduced collisions; workflow green
+#   1  - at least one collision (``--fail-on-collision``); workflow red
+#        with the offending names in the job summary
+#
+# Unmatched rows are reported but never fail the job — deep vendor
+# boards legitimately list players outside the Sleeper pool (retired
+# veterans, non-Sleeper prospects).  Only collisions are a hard defect.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      # The refresh commits land here; run the check on the new data.
+      - "CSVs/site_raw/**"
+      - "exports/latest/dynasty_data_*.json"
+      # ...and whenever the join or the alias table itself changes.
+      - "src/utils/name_clean.py"
+      - "src/api/data_contract.py"
+      - "scripts/audit_identity_matches.py"
+  schedule:
+    # 08:17 UTC daily — backstop in case a refresh lands without
+    # touching the path filters above.
+    - cron: "17 8 * * *"
+  workflow_dispatch:
+
+concurrency:
+  group: audit-identity-matches
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  audit:
+    name: Audit CSV -> pool identity join
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: |
+          set -Eeuo pipefail
+          # Runtime manifest only — this job does not run the test
+          # suite.  No ``|| true``: a genuine install failure must
+          # surface as a red workflow, not a green-but-useless run.
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+          pip check
+
+      - name: Locate latest snapshot
+        id: snapshot
+        run: |
+          set -Eeuo pipefail
+          SNAPSHOT=$(ls -t exports/latest/dynasty_data_*.json 2>/dev/null | head -1 || true)
+          if [ -z "$SNAPSHOT" ]; then
+            # A missing snapshot is itself the alert: the export
+            # pipeline has broken.  Returning 0 here would make
+            # "audit didn't run" look identical to "board is healthy".
+            echo "::error title=Missing snapshot::no dynasty_data_*.json in exports/latest/ — export pipeline may be broken"
+            exit 1
+          fi
+          echo "Using snapshot: $SNAPSHOT"
+          echo "path=$SNAPSHOT" >> "$GITHUB_OUTPUT"
+
+      - name: Run identity match audit (fails on alias collision)
+        run: |
+          set -Eeuo pipefail
+          python scripts/audit_identity_matches.py \
+            --json-path "${{ steps.snapshot.outputs.path }}" \
+            --json "${RUNNER_TEMP}/identity_audit.json" \
+            --limit 10 \
+            --github-summary \
+            --fail-on-collision
+
+      - name: Upload full report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: identity-audit-report
+          path: ${{ runner.temp }}/identity_audit.json
+          if-no-files-found: warn
+          retention-days: 30
+
+      - name: Open / update rolling issue on collision
+        if: failure()
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -Eeuo pipefail
+          RUN_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+          REPORT="${RUNNER_TEMP}/identity_audit.json"
+          # jq, not inline python: an indented python -c heredoc inside a
+          # YAML block scalar is an IndentationError waiting to happen.
+          DETAIL="(no report produced — the job failed before the check ran)"
+          if [[ -f "$REPORT" ]]; then
+            DETAIL=$(jq -r '
+              (.aliasCollisions // [])
+              | if length == 0 then
+                  "(no collisions recorded — the job failed outside the check)"
+                else
+                  map("- " + (if .crossFamily then "**CROSS-FAMILY** " else "" end)
+                      + "`" + .canonicalName + "` merges: " + (.mergedNames | join(", ")))
+                  | join("\n")
+                end' "$REPORT" || echo "$DETAIL")
+          fi
+          # Body written flush-left on purpose: an indented heredoc would
+          # render the whole issue as a markdown code block.
+          BODY=$(cat <<EOF
+          The identity-match audit found an **alias-introduced collision** on the live board.
+
+          A collision means two DISTINCT player-pool rows now resolve to one canonical
+          name, so their source votes merge and both players' values are corrupted.
+          Cross-family collisions are worse: the name-only CSV join replicates a matched
+          entry across every position group sharing the name.
+
+          **Collisions:**
+
+          ${DETAIL}
+
+          - Workflow run: ${RUN_URL}
+          - Commit: \`${GITHUB_SHA}\`
+          - Detected: $(date -u +"%Y-%m-%dT%H:%M:%SZ")
+
+          Likely causes:
+          1. A vendor renamed a player into a spelling that already exists in the pool
+             (most common — check the newest \`CSVs/site_raw\` diffs).
+          2. A newly added entry in \`CANONICAL_NAME_ALIASES\` (\`src/utils/name_clean.py\`)
+             merged two real players.
+          3. The Sleeper pool gained a second row for a player we alias.
+
+          Triage: run \`python scripts/audit_identity_matches.py\` locally; see
+          \`docs/identity-audit-2026-07.md\` for the verification standard (an alias is
+          only correct when position + team + age agree, or an explicit ID matches).
+          EOF
+          )
+          # Strip the YAML block-scalar indentation from every line so the
+          # issue body is valid markdown rather than one big code block.
+          BODY=$(printf '%s\n' "$BODY" | sed 's/^          //')
+          gh label create identity-collision \
+            --color B60205 \
+            --description "Auto-opened when the identity-match audit finds an alias collision" \
+            2>/dev/null || true
+          # Idempotent rolling issue: comment on the open one if it
+          # exists, else open a single labelled issue and reuse it.
+          EXISTING=$(gh issue list --state open --label identity-collision \
+                       --json number --jq '.[0].number' 2>/dev/null || echo "")
+          if [[ -n "$EXISTING" && "$EXISTING" != "null" ]]; then
+            gh issue comment "$EXISTING" --body "$BODY"
+          else
+            gh issue create \
+              --title "Identity audit: alias-introduced collision on the live board" \
+              --body "$BODY" \
+              --label identity-collision
+          fi
--- a/docs/identity-audit-2026-07.md
+++ b/docs/identity-audit-2026-07.md
@@ -1,0 +1,166 @@
+# Identity Audit — July 2026
+
+Sweep of every source CSV in `_SOURCE_CSV_PATHS` for player votes lost
+to name drift between vendor spellings and the Sleeper player pool.
+Tooling: `scripts/audit_identity_matches.py` (new — replays the exact
+`_canonical_match_key` join the contract build performs and fuzzy-scores
+every unmatched row against the pool).  Payload audited:
+`exports/latest/dynasty_data_2026-07-26.json` (1,095 pool rows after
+far-future pick injection); numbers re-verified after the 2026-07-26
+data refreshes and the Phase 3-6 merges (search/filters, playerctx,
+news) — none of which changed the CSV → pool join this sweep fixes.
+
+## Per-source match rates
+
+Before → after adding the 18 alias entries below to
+`CANONICAL_NAME_ALIASES` (`src/utils/name_clean.py`).  "id" = rows
+recovered only by the sleeper_id join (pfkDynasty is the only CSV
+carrying ids the parser recognizes today).
+
+| Source | CSV rows | Matched before | Matched after | Recovered | Rate after |
+|---|---|---|---|---|---|
+| ktc | 500 | 497 | 499 | +2 | 99.8% |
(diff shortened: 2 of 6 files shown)


jasonleetucker-code:

Reviewed commit: `ed8c592c` (branch `claude/identity-sweep`, base `main`)

Everything below was verified empirically against the committed CSVs, the committed export payload, and live A/B contract builds — not by reading the doc table. **No blocking findings.** All 18 aliases are correct same-player collapses and the direction is right on every one. Three follow-ups worth filing, none of which should hold the merge.

---

### What I verified

**1. All 18 aliases are same-player (checked all 18, not a sample).**
For every alias I resolved the KEY and the VALUE against the pool (`exports/latest/dynasty_data_2026-07-26.json`, 1,095 rows) and against every vendor CSV row that produces the key:

- Every alias VALUE resolves to **exactly one** pool row. No alias VALUE is ambiguous in the pool.
- **No alias KEY currently matches any pool row** — so no alias silently re-points an existing pool player. This is the specific trap in item 5 and it is clean today.
- Position / team / age agree between the vendor row and the pool row in every case.

The strongest individual confirmations (beyond position+age):

| Alias | Independent evidence found |
|---|---|
| `kenneth gainwell` → `kenny gainwell` | ID-grade: **both** `pfkDynasty.csv` and `dynastyNerdsSfTep.csv` carry `sleeper_id/SleeperId = 7567` on the row spelled "Kenneth Gainwell"; pool `Kenny Gainwell` is sid 7567. DS narrative text independently confirms RB, ex-PIT, signed TB, age 27.3. |
| `alim mcneil` → `alim mcneill` | Only one `Alim McNeil*` exists anywhere in the CSV corpus (`idpTradeCalc` single-l; `draftSharksIdp` DL age 26.2 double-l). The only other "McNeil" is Emmanuel McNeil-Warren, clearly distinct. |
| `justin madubuike` → `nnamdi madubuike` | Exactly one Madubuike league-wide. `dlfIdp` is the sole file still on "Justin"; `draftSharksIdp` (DL, 28.7), `idpShow` (DT), `idpTradeCalc` all already say "Nnamdi". Rename, not two people. |
| `cameron ward`, `cameron skattebo` | Pool spelling appears in **14** other CSVs; only `draftSharksSf` uses "Cameron". Direction is unambiguous. |
| `josh palmer` → `joshua palmer` | Pool spelling in 10 CSVs; `dlfSf` is the lone outlier. |
| `nickolas martin` → `nick martin` | `idpShow` lists it as LB; pool `Nick Martin` is LB/SF/age 23. The retired IND center is not in the pool at all, so there is nothing for it to collide with. |

**2. Audit script.** `python3 scripts/audit_identity_matches.py` → exit **0**, `Alias collision-delta check: clean (0 alias-introduced pool collisions)`. The per-source match table it prints matches the doc's "after" column **exactly** on all 22 sources.

**3. Tests.** `tests/utils/test_name_clean.py` + `tests/scripts/` → **265 passed**. Widened to `tests/identity tests/utils tests/scripts` → **421 passed**. `tests/api/test_source_registry_parity.py` + `test_source_overrides.py` → **77 passed**.

The hygiene invariants genuinely execute (I confirmed they are parametrized/looped over the live table, not stubs):
- `test_alias_keys_are_normalize_stable` — renormalizing each key is a no-op, so no dead unreachable key.
- `test_alias_values_are_fixed_points` — no `a→b, b→c` chains.
- `test_vendor_spelling_collapses_to_pool_spelling` — asserts `resolve_canonical_name(vendor) == normalize_player_name(pool)`, which is what actually **pins the direction**.

**4. Alias direction.** `resolve_canonical_name` is `CANONICAL_NAME_ALIASES.get(norm, norm)` (`src/utils/name_clean.py:300`), and both sides of the join go through it — the pool row via `_canonical_match_key(nm)` (`src/api/data_contract.py:3491`) and the CSV row at parse time. So keys must be the vendor spelling and values the pool spelling. Verified correct for all 18. Worth noting the table now legitimately contains **both** directions of the nickname convention (`pat mahomes`→`patrick mahomes` vs `patrick surtain`→`pat surtain`); that is correct-by-construction since the value must equal the pool spelling, and the test pins it.

**5. Collision detector logic is genuine, not weaker.** `alias_collision_delta` computes `pre_key = normalize_player_name(nm)::group` and `post_key = canonical_player_key(nm, pos)` (which applies the alias), then flags any post-key absorbing >1 distinct pre-key. That is a real pre- vs post-alias comparison over pool rows. I confirmed it fires on a synthetic same-group merge rather than just passing vacuously.

**6. Contract-level impact reproduced exactly.** I ran two full `build_api_data_contract` builds (alias table on vs off) and diffed `canonicalSiteValues` per row. Result: **16 rows gain 24 source votes, 0 rows lose anything, pool row count unchanged at 1,095** — matching the doc's claim precisely, including the per-player breakdown.

> Note for anyone re-running this: you must clear `dc._SOURCE_CSV_PARSE_CACHE` between builds. It is keyed on path+mtime only, and the `csv_lookup` it stores is **already alias-resolved**, so a naive A/B run reports a false "0 delta". Not a product bug (an alias-table edit implies a process restart), but it will mislead a reviewer.

---

### P2-1 — Collision detector is blind to cross-position-group merges, and a test enshrines that blindness

`scripts/audit_identity_matches.py` (`alias_collision_delta`) keys collisions by `name::position_group`, so two pool rows that share a post-alias name but sit in different groups are reported clean. `tests/scripts/test_audit_identity_matches.py::TestAliasCollisionDelta::test_cross_group_spellings_do_not_collide` asserts exactly this pair — `Michael Jackson` (WR) + `Mike Jackson` (CB) — returns `[]`.

Concrete scenario, empirically demonstrated. I injected a hypothetical WR named "Michael Jackson" into the payload and rebuilt the contract:

```
'Michael Jackson'  pos='WR'  sources=['draftSharksIdp', 'ktc']
                   draftSharksIdp=1  origRanks={'draftSharksIdp': 143.0}
'Mike Jackson'     pos='DB'  sources=['draftSharksIdp']
                   draftSharksIdp=1  origRanks={'draftSharksIdp': 143.0}

audit detector on this pool: []
```

The WR absorbs the CB's `draftSharksIdp` rank-143 vote, and the detector reports clean. Mechanism: once aliased, both rows share canonical key `mike jackson`, so `row_groups_by_key` has 2 groups and `_enrich_from_source_csvs` (`src/api/data_contract.py:3592-3608`) **replicates the best CSV entry across both groups** with `ambiguous: True`.

Fair nuance, stated plainly: this is not created from nothing by the PR. Without the alias, that same DS-IDP vote landed on the WR *only* — also wrong, just differently wrong — because the CSV-side join is name-only and never uses the vendor row's own position. So the underlying weakness is pre-existing. What the PR changes is that (a) the blast radius widens from one wrong row to two, and (b) the **new safety net specifically declares this case clean**, so it will not warn.

Why it matters here specifically: several of the 18 new keys are generic, high-recurrence names — `michael jackson`, `robert henry`, `josh palmer`, `nickolas martin`, `cameron ward`, `gabriel davis`. A future rookie or UDFA at a different position group named any of these silently inherits votes.

Suggested follow-up: have the detector additionally report a **name-level** (group-ignoring) merge as a warning tier distinct from the hard collision, and relabel that test to say "cross-group is tolerated by the join today" rather than asserting it as correct. Longer term, the real fix is using the vendor CSV's own position column (most of these files have one) to pick the group instead of replicating across groups.

### P2-2 — The invariant never runs on the path that actually changes the data

The collision check is pinned only by `TestCommittedDataRegression`, i.e. it runs when a human opens a PR. But `.github/workflows/scheduled-refresh.yml` commits and pushes refreshed `exports/**` and `CSVs/**` straight to `main` (`git commit -m "chore: automated data refresh ..."` / `git push`, lines ~455-461) **without running pytest** — and the export is refreshed several times a day (three commits to that file on 2026-07-26 alone). The audit script is also deliberately exit-0-always.

Net effect: a scrape that introduces a colliding pool row ships to production silently and is only caught whenever someone next opens an unrelated code PR.

There is already a clean precedent in this repo: `.github/workflows/audit-dropped-sources.yml` is a weekly cron that runs an audit script and **fails red** on a threshold, explicitly motivated by "without this cron, a regression would only be caught whenever someone manually runs the audit, which historically meant never." Recommend the same shape here — add an `--exit-nonzero-on-collision` flag and a weekly cron. Cheap, and it closes the gap for a value-affecting table.

### P3-1 — `idpTradeCalc.csv` carries both spellings of one aliased pair

`idpTradeCalc.csv:113` is `Andru Phillips,1233.0` and `:331` is `Dru Phillips,1233.0`. Post-alias these collapse to one canonical key, so the entry is stamped `ambiguous: True` with `candidates: ['Andru Phillips','Dru Phillips']` and surfaces that way in `sourceAudit` (`src/api/data_contract.py:7011`).

This is **benign** and does not change any number: both rows carry the identical value 1233.0, so the best-of-value pick is a no-op, and `ambiguous` is diagnostic-only with no effect on the blend. The identical value is in fact the strongest possible confirmation that the alias is right — it is one vendor double-listing one player, not two players. Worth a one-line note in the doc since it is the only alias where a single vendor emits both spellings, and a reviewer who greps for that pattern will otherwise be alarmed by it.

### P3-2 — Doc's age column has drifted from the CSVs it cites

`docs/identity-audit-2026-07.md` presents ages as "pool vs DraftSharks CSV" evidence, but ~10 of the DS figures are off by 0.1 from the current files — e.g. Andru Phillips doc `24.5` vs CSV `24.6`; Skattebo `24.3` vs `24.4`; Ward `24.0` vs `24.2`; Landman `27.6` vs `27.7`; Surtain `26.2` vs `26.3`; Hill `25.7` vs `25.8`; McCulley `23.4` vs `23.5`; Michael Jackson `29.4` vs `29.5`; Robert Henry `24.4` vs `24.5`.

Immaterial to correctness — every one still corroborates the same player, and this is almost certainly just a CSV refresh landing after the doc was written. Flagging only because the table is explicitly framed as the verification evidence, so exact-looking numbers that no longer reproduce undermine its value on the next audit.

### P3-3 — Audit scope is narrower than the alias table's blast radius

`CANONICAL_NAME_ALIASES` has three live consumers: `src/api/data_contract.py`, `src/identity/matcher.py:29`, and `src/ros/mapping.py:34` (where an alias hit is a 0.9-confidence match tier). The audit script only replays the `_SOURCE_CSV_PATHS` join, so the "0 collisions" result is scoped to the contract pool and does not speak for the ROS mapping or the identity matcher.

Low risk in practice — all 18 are verified same-player, which is the property those consumers need too, and the ROS CSVs (`draftSharksRosIdp.csv`, `draftSharksRosSf.csv`) already use the pool spellings, so the aliases help there. Worth one sentence in the doc scoping the claim rather than any code change.

---

### Verdict

The value-affecting part of this PR — the 18 aliases — is **correct**. Direction is right on all 18, no alias merges two real players, the claimed +24/-0 contract delta reproduces exactly, and the tests are real assertions over the live table rather than decoration. The tooling and docs are unusually thorough for an identity sweep.

The two P2s are both about the *safety net* rather than the change itself: the detector has a cross-group hole that a test currently blesses, and it never runs on the automated data-refresh path where new collisions would actually appear. Neither blocks this merge; both are worth a follow-up issue before the next sweep adds more generic-name aliases.

---

jasonleetucker-code:
Both P2s and the P3s are fixed in `34e27cc7`. The 18 aliases are unchanged — `src/utils/name_clean.py` and `src/api/data_contract.py` are untouched by this round.

## P2-1 — cross-family collisions are now detected

You were right, and the fix is at the granularity the join actually uses. `alias_collision_delta` keyed on `canonical_player_key` (`name::group`), but `_enrich_from_source_csvs` keys `csv_lookup` by the **name-only** `_canonical_match_key` and *replicates* a matched entry across every position group sharing that name (the `len(row_groups) > 1` branch). So a group-keyed check could never see the case that matters.

The check now compares pre-alias `normalize_player_name` against post-alias `resolve_canonical_name` at name granularity, and stamps `crossFamily` on each finding.

- The test that blessed the blind spot (`test_cross_group_spellings_do_not_collide`) is replaced by `test_cross_family_merge_is_detected`, asserting your exact Michael Jackson WR / Mike Jackson CB scenario **is** reported with `crossFamily: True` and `positionGroups: ["IDP", "OFFENSE"]`.
- Added `test_preexisting_name_collision_is_not_alias_introduced` to pin the other side: "DJ Turner" WR vs "DJ Turner II" CB collapse via the *suffix stripper*, not the alias table, so a delta check must not attribute them to aliases.
- Verified on the **real pool**, not just synthetic fixtures — injecting a cross-family bad alias (`dru phillips` → `cam ward`) reports `crossFamily: True`; injecting a same-family one (`joshua palmer` → `gabe davis`) reports `crossFamily: False`; removing them returns to clean.

**Live board is still 0 collisions under the stricter check, so no whitelist was needed.**

On your nuance: agreed and worth stating plainly — the name-only join was already position-blind, so pre-alias that vote landed on the WR alone, which was also wrong. The alias widens the blast radius from one wrong row to two. The detector now catches the widening; the underlying position-blindness of the CSV join is pre-existing and out of scope here.

## P2-2 — the invariant now runs on refreshed data

New `.github/workflows/audit-identity-matches.yml`, following the `audit-dropped-sources.yml` precedent:

- Triggers on pushes to `main` touching `CSVs/site_raw/**` or `exports/latest/dynasty_data_*.json` (where the refresh actually lands), plus a daily cron backstop and `workflow_dispatch`.
- Runs the audit with a new `--fail-on-collision` flag (the script stays exit-0-always by default), writes a `$GITHUB_STEP_SUMMARY` table, uploads the JSON report, and opens/updates a rolling `identity-collision` issue.
- **Separate workflow by design** — per your constraint, an advisory finding must never block the data refresh, so `scheduled-refresh.yml` is untouched and stays green while this job goes red.
- Unmatched rows never fail the job; only collisions do.

Both gate paths were smoke-tested locally: clean board → exit 0; injected bad alias → exit 1 with the `::error` annotation, and the `jq` extraction renders `- **CROSS-FAMILY** \`cam ward\` merges: Cam Ward, Dru Phillips` into the issue body. (I used `jq` rather than inline `python -c` — an indented heredoc inside a YAML block scalar is an `IndentationError` waiting to happen — and strip the block-scalar indentation off the body so the issue isn't one big code block.)

Note this relaxes the original "don't touch `.github/workflows/**`" constraint, per your explicit request; it adds a new file rather than editing any existing workflow.

## P3s

- **DraftSharks ages refreshed** against the 2026-07-26 CSVs. Also found the cause: DS ages are decimal and re-derived every refresh, so they drift ~0.1 per snapshot. Documented so the next reader doesn't treat it as an error.
- **Separately noticed while doing that**: DraftSharks now ships an **empty `Team` column** (header still declares it). Nothing reads it today, but it was the strongest corroborating field for alias verification — flagged as a recommendation.
- **Blast radius documented**: `src/ros/mapping.py` (imports `CANONICAL_NAME_ALIASES` directly as `_ALIASES`) and `src/identity/matcher.py` (`_identity_canonical_key` builds master identity records from it) also consume the table, so the audit's scope is narrower than the table's reach.
- **`_SOURCE_CSV_PARSE_CACHE` gotcha recorded** in both the script docstring and the report: it's keyed on `(path, mtime)` and caches an *already alias-resolved* lookup, so an in-process A/B contract build reports a false "0 delta". Run each arm in a fresh interpreter. All before/after numbers in this PR were produced that way.

## Validation

- `python -m pytest tests/ -q -m "not livedata"` → **3067 passed, 280 deselected, 0 failures**
- `ruff check` + `ruff format --check` clean on touched Python
- Workflow YAML parses; both gate paths exercised locally

---

PR #549, opened 2026-07-26
Redesign R2: rankings board + player profiles

jasonleetucker-code:
## What this is

R2 of the redesign: **the rankings board and the player profile** — the two highest-traffic surfaces and the heart of the war-room × terminal brief — rebuilt on the R0 design system and the R1 shell. Plus the first UI surface for the `src/playerctx` data layer (#539), which previously had none.

## Rankings board (`app/rankings/`)

Rebuilt on `PageHeader` / `Panel` / `DataTable` / `Tabs` / ds controls. **Every existing behavior is preserved verbatim** — lenses, pos + confidence filters, the Phase 3 advanced-filter rail + token search, sortable columns including the dynamic per-source columns, tier grouping, row expansion with the source-audit panel, watchlist toggles, news chips, Copy/CSV export, row limit + filter bypass, IDP gating, `?pos=` deep-link seeding. `buildRows` stays pure, backend stamps render verbatim, and no ranking math moved client-side.

- **Sorting split**: the page memo keeps the sort pipeline (the lens system and `src:` columns are page semantics); `DataTable` renders `presorted` rows and owns only the header state machine + `aria-sort` stamps. The audit found **zero** `aria-sort` in the app — every sortable header now has it, headers are real buttons, and rows are keyboard-expandable.
- **Density as the point**: compact density, tabular-nums everywhere, tier bands as structural separators (`renderBeforeRow`) rather than decoration; trust strip is a StatTile row; the two signal rails are Panels with `Movement` glyphs replacing the ASCII arrows.
- **Decomposition**: `board-utils.js` (formatSourceCell / exportSourceCells, extracted verbatim), `board-sections.jsx` (methodology, rails, audit panel, mobile source strip), `board.module.css` (token-driven structure). Page drops from 2,214 lines to ~1,050.
- **Performance**: measured before rebuilding — the 200-row default + filter bypass keeps render cost flat on the ~1.1k-row pool, so virtualization is **not** warranted (documented inline; revisit if the default limit passes ~500). `/rankings` page chunk **59.9 → 55.9 KB**.

## Player profile (`components/PlayerPopup.jsx`)

Now the scouting card, on the ds `Drawer` — which brings the focus trap, Escape, focus restore, and overlay-stack semantics the audit found entirely missing from the old overlay div. Identity header (headshot, pos/team/age/experience/rookie/rank, ownership + roster slot), Our Value + injury-adjusted StatTiles, ROS context, value chain, Sharp Tracker intel (#534), edge signal, pick projection, rank history, news digest + items (#540), source breakdown, consensus narrative — **all data derivations preserved verbatim**.

**NEW — `PlayerContextSection` (#539's first UI surface):** contracts (APY, guaranteed, end year, with the OTC signing-team caveat surfaced as a tooltip), snap share (season %, games, side, and the 3-game trend as a `Movement`), and depth-chart standing (RB1 + starter/backup badge). Cached per player, every block optional, silent degrade when the snapshot has nothing.

## Backend (additive)

`GET /api/playerctx/player` — mtime-cached snapshot read keyed by Sleeper id through `sleeperIndex` per the `docs/playerctx.md` consumer rules (record keys are gsis ids *or* `sleeper:<id>` fallbacks). Global player metadata, so it follows the scoring-profile side of the CLAUDE.md split: no league resolution, session-gated like the intel reads. Missing snapshot or unknown player is a clean 404, never a 5xx.

## ds primitive extensions (additive, all tested)

`DataTable`: `presorted`, `rowClassName`, `renderBeforeRow`/`renderAfterRow`, `headerTitle`, `firstDirection`, `align: center`. `Dialog`: `closeLabel`, `className`. `Icon`: star / star-filled. `ds.css`: PageHeader actions now wrap, and `.ds-panel` gained a `min-width: 0` guard — both fixed **real mobile overflow bugs** (see below).

## E2E

SEL registry reconciled to R1+R2 (board rows, controls, dialogs, palette results, shell search); mobile-smoke's bottom-nav assertions reconciled to the R1 tab bar (Home/Ranks/Trade/News/Menu); palette label regex updated.

Ran locally against a real stack: **journey-rankings 6/6 desktop + mobile-smoke 3/3 mobile-chromium PASS**. The run caught two genuine regressions I then fixed — the PageHeader action row overflowing a 390px viewport by 185px, and the signal rails' flex children widening the mobile layout viewport (which also made the drawer's close button unclickable at fixed coordinates). Worth noting the E2E net did its job on the first surface it was pointed at.

## Verification

- **vitest: 53 files / 1159 tests green** (adds PlayerContextSection specs + DataTable extension-prop specs)
- **`npm run build` green, all bundle budgets pass** — no page moved except `/rankings` shrinking
- **`pytest tests/api`: 1222 passed, 3 skipped, 366 subtests** (full suite; includes the 6 new playerctx endpoint tests)
- **ruff 0.6.9** (repo-pinned) clean on touched Python
- Latest `main` merged; only automated data-refresh commits landed since

---


jasonleetucker-code:

Reviewed commit: `bf15133b02f8f6bed343001dc07a759a1f732afc` (base `origin/main`)

I did not write this PR. I diffed the pre-R2 `page.jsx` (2214 lines) and `PlayerPopup.jsx` against their rebuilds line-by-line to test the "every existing behavior preserved verbatim" claim, audited the new endpoint against its siblings, and grepped every existing consumer of the extended ds primitives.

**Headline: the behavior-preservation claim largely holds.** The load-bearing logic (`formatSourceCell`, `exportSourceCells`, the filter/sort pipeline, the source-audit panel, the value chain, the popup's `siteDetails` derivation) is a verbatim extraction, and the sharp edges I went hunting for survived. The findings below are two process/completeness gaps and a set of small silent deltas.

### Tests run

| Suite | Result |
|---|---|
| `cd frontend && npx vitest run` | **1159 passed / 53 files**, 0 failed |
| `python3 -m pytest tests/api/ -q` | **1222 passed, 3 skipped**, 366 subtests passed (13 min) |
| `python3 -m pytest tests/api/test_playerctx_endpoint.py -q` | 6 passed |

---

## P2

### P2-1 — `PlayerContextSection` ships with no producer for the data it reads

`frontend/components/PlayerPopup.jsx:299` renders the new playerctx block, and `server.py:10684` serves it from `data/playerctx/snapshot.json`. Nothing schedules the script that writes that file:

```
$ grep -rn "refresh_playerctx" --include=*.yml --include=*.sh --include=*.service --include=*.timer .
(no hits — only docs/playerctx.md)
```

`docs/playerctx.md:190-192` names this exact precondition:

> No GitHub Actions workflow is added in this PR — schedule `python3 scripts/refresh_playerctx.py` alongside the existing weekly jobs **when the UI starts consuming the snapshot**.

This *is* that PR, and the workflow/timer still isn't there. `data/` is gitignored and not a build artifact, so on the Hetzner box the file will not exist: every `/api/playerctx/player` call returns the (correct, clean) 404, `PlayerContextSection` returns `null`, and the headline new feature of the profile redesign is invisible in production. It degrades silently, which is the right failure mode — but it means the section can never be verified in prod. Compounding it, `_ctxCache` (`PlayerPopup.jsx:283`) stores the `null` for `_CTX_TTL_MS` = 30 min per player, so the first backfill won't surface for up to half an hour per already-viewed player.

**Fix:** add the weekly workflow (or a systemd timer in `deploy/`) in this PR, or state explicitly that the section is expected to be dark until a follow-up lands.

### P2-2 — The rewritten board has no component-level test, and the E2E net it relies on was already red

The two new frontend test files cover a leaf section (`PlayerContext.test.jsx`) and the primitive (`ds/DataTable.test.jsx`). Nothing renders `RankingsPage`. `frontend/__tests__/custom-mix-badge.test.js` only exercises the pure `describeCustomMix` export, which is unchanged.

So the entire "verbatim preservation" claim for a 2214→1525-line rewrite of the highest-traffic surface rests on `tests/e2e/specs/journey-rankings.spec.js` — and this PR had to repair that suite's selectors for **R1** shell changes, not R2 ones:

- `tests/e2e/helpers/journey.js:41` `.nav-search-btn` → `.shell-search-btn`
- `:38` `.picker-sheet .asset-row` → `.shell-palette-option`
- `tests/e2e/specs/mobile-smoke.spec.js:70` `"More"` → `"News"` / `"Menu"`

Those are R1 regressions, which means the journey suite was failing on `main` before this branch and has therefore **never** validated the R2 board. The claimed safety net is being built in the same commit it is being trusted for.

**Fix:** add a jsdom render test over `RankingsPage` pinning the four behaviors most likely to silently rot: (a) tier separator emitted only when `tierGroupingActive`, (b) row click expands / interactive-descendant click does not, (c) first-click sort direction per column class, (d) `hasActiveFilter` bypassing `rowLimit`.

---

## Behavior-preservation audit — verified clean

Each of these I diffed against the pre-R2 implementation and confirmed identical:

| Behavior | Verdict |
|---|---|
| **`src:` column sort fallback rule** | Verbatim. `page.jsx:503-520` preserves `allowRawFallback = !src?.isRankSignal`, `Number.isFinite(aMeta)` gate, and the `|| 0` sink — rank-signal sources still never read the synthetic `canonicalSites` encoding. |
| **First-click sort direction** | Preserved *exactly*, via a non-obvious route. Old: `setSortAsc(["rank","name","pos"].includes(col))`. New: `DataTable.handleSortClick` defaults `numeric → desc, text → asc`, and the page marks `rank` `firstDirection:"asc"`, `name`/`pos` text, `score`/`value`/`src:*` `numeric`, `confidence` `firstDirection:"desc"`. Every column resolves to the old direction. Same-column re-click flips in both. |
| **Row-limit bypass** | `page.jsx:543-546` byte-identical predicate. |
| **IDP snap-back on league switch** | `page.jsx:314-318` identical, incl. the `DL`/`LB`/`DB` list. |
| **`?pos=` deep-link seeding** | `page.jsx:325-337` identical, incl. the lower/upper normalization split and mount-only deps. |
| **Tier grouping** | `renderBeforeRow` (`page.jsx:1471-1483`) is logically equivalent to the old `showTierBreak` (`idx > 0 && tierId !== prevTierId && tierId != null`), and `i` is the index into the presorted `displayRows`, so `displayRows[i-1]` is correct. |
| **`totalCols` colspan** | Old `10 + (showSiteCols ? visibleSources.length : 0)`; new `columns.length` = 6 fixed + N src + 4 tail = same number, same column order. |
| **Copy / CSV export** | Same headers, same cells (`exportSourceCells` is the extracted original), same filename pattern. The TSV path's `escape` default is identity so `Array.join` coercion of `null` still yields `""`. |
| **Source-audit panel** | `board-sections.jsx:189+` is a verbatim move, including all five `audit.reason` strings and the matched/missing/expected/n-a badge ladder. |
| **`stopPropagation` on nested controls** | Correctly *replaced*, not dropped. Old used explicit `e.stopPropagation()` on the name span, value span, star, and chip anchors; new relies on `DataTable`'s `fromInteractiveDescendant` (`DataTable.jsx:88`) matching `a,button,input,select,textarea,[role=button],[role=link]`. All four sites became `<button>`/`<a>`, so they are covered. |
| **Mobile source strip** | Still emits `<tr className="rankings-mobile-source-row">`, so `globals.css:1937/2002` still governs it. |
| **Watchlist / news chips / lenses / conf filter / adv rail + token search** | All routed through the same `rowMatches` / `applyLens` / `useUserState` / `lookupPlayerNews` calls with unchanged arguments. |
| **Removed imports** | `getRetailLabel`, `source`, `selectedLeagueKey`, `trustStats.withAnomalies` were all computed-but-unused in the old page. Dead-code removal, not a regression. |

**Incidental bug fix worth noting:** the old `SortHeader` (`page.jsx:842` pre-R2) destructured only `{col, children, style, className}` and set its own `title`, so the descriptive tooltips passed to the Consensus and per-source headers were **silently discarded**. `headerTitle` on the new `DataTable` `<th>` actually delivers them. Net: those tooltips now work; the generic "Sort by X" hint is gone from all sortable headers.

## buildRows purity / no client-side math — verified clean

`frontend/lib/dynasty-data.js` is untouched by this PR. Nothing in `page.jsx`, `board-utils.js`, `board-sections.jsx`, or `PlayerPopup.jsx` recomputes ranks or values: `val` reads `rankDerivedValue ?? values.full`, the sort comparators read `resolvedRank` / `blendedSourceRank` / `sourceRankMeta[k].valueContribution`, `movers` reads the `rankChange` stamp, and `positionRankByName` / `siteDetails.pct` / `consensusText` CV are display derivations carried over unchanged from the old page. No fallback blend was reintroduced.

## New endpoint — verified clean

- **Auth gating matches siblings.** `/api/playerctx/player` is absent from `_PUBLIC_API_EXACT`, `_SELF_AUTHED_API_EXACT`, and `_PUBLIC_API_PREFIXES`, so `_private_api_gate` (`server.py:2557`) 401s anonymous callers. That is exactly the posture of `/api/intel/player` (`server.py:10408`), which is likewise not allowlisted. ✅
- **Missing snapshot / unknown player → clean 404.** `_playerctx_snapshot` catches `OSError` on `stat()`; `store.load_snapshot` catches `(OSError, ValueError)` and shape-checks `players`; `_playerctx_lookup` returns `None` on a missing `sleeperIndex` entry or a non-dict record. Every path lands on the 404 with `error: "no_context"`. No 5xx surface. ✅
- **CLAUDE.md scoping is correct.** No `_resolve_league_for_request`, no `leagueKey` param, no registry read — playerctx is global player metadata and is treated as scoring-profile-side. ✅
- **mtime cache cannot serve stale data across a snapshot rewrite.** `store._atomic_write_json` writes to `path.with_suffix(...tmp-<ms>)` then `tmp.replace(path)`, so the published inode's mtime is always the fresh write time — strictly increasing, and no torn read is possible. I traced both interleavings of `stat()`/`load_snapshot()`: the worst case caches *fresher* content under an older mtime, which self-corrects on the next request. ✅

One residual (P3): the cache keys only on mtime, so a snapshot restored with `cp -p` / `rsync --times` / a tar extract that preserves an older-or-equal mtime would be invisible for the process lifetime. `(st_mtime, st_size)` would close it cheaply. Not reachable via the documented refresh path.

## ds primitive extensions — verified clean

Every new prop is additive with a default that reproduces prior behavior, and I grepped the full consumer set:

- **`DataTable`**: only consumer on `main` is `frontend/app/design/DesignGallery.jsx` (plus tests), and it passes none of the new props. `presorted=false` → `sortRows` as before; `rowClassName=null` → `extraClass=""` → the className join produces the identical string or `undefined`; `renderBeforeRow/AfterRow=null` → no-op; `headerTitle` undefined → React omits `title`; `firstDirection` falls back to `col.numeric ? "desc" : "asc"`; `align` is gated on `!col.numeric` so numeric still wins. The `<tr key>` → `<Fragment key>` move preserves key identity semantics.
- **`Dialog`**: `closeLabel="Close"` and `className=""` defaults are byte-equivalent (`` `${base} `.trim() === base ``). `Modal`/`Drawer` spread props, so the three existing call sites (`DesignGallery`, `CommandPalette`, `MobileChrome`) are unaffected.
- **`Icon`**: `star` / `star-filled` are pure additions.

---

## P3 (silent deltas — none are correctness bugs, but they weren't called out)

1. **`rankings-audit-row` class dropped.** `page.jsx:1494` renders the audit panel in a bare `<tr>`. `globals.css:2460` `.rankings-audit-row td { padding: 0 !important; border-top: none; background: rgba(15,23,42,.6) }` no longer applies, so the expanded panel gains `.ds-table td` padding and loses its dark inset. Cosmetic but visible on every expansion.
2. **Source-breakdown bars lost their magnitude encoding.** Old `PlayerPopup.jsx:1119` colored the fill `green ≥90% / cyan ≥50% / red <50%`; `player-card.module.css:219` is now a flat `var(--chart-1)`. If that was decorative, fine — but it read as signal.
3. **Empty filtered result renders no table at all.** `page.jsx:1450` passes no `emptyState`, and `DataTable.jsx:216` returns the default `null` when `rows.length === 0`. Previously the headers stayed put over an empty `<tbody>`. Only the "0 shown" line remains. Passing an `<EmptyState>` (already imported) would be a one-liner.
4. **Quarantined StatTile now always renders**, showing `0`. `page.jsx:1143` — the old trust bar hid the stat entirely at zero (`page.jsx:1268` pre-R2).
5. **Per-lens tooltips dropped.** Old lens buttons carried `title={lens.description}`; `Tabs` has no per-tab `title` and `page.jsx:1223` only passes `{id,label}`. The description still renders below for non-consensus lenses.
6. **`.ds-page-header__actions` change is global, not page-scoped.** `ds.css:335-341` drops `flex-shrink: 0` and adds `flex-wrap: wrap` + `min-width: 0`, affecting all **25** `<PageHeader>` consumers. Almost certainly an improvement (the rankings header has 6 actions), but it's a system-wide layout change riding in a page rebuild — worth a spot-check on a narrow-title page like `/waivers` or `/trending`.
7. **Breakpoint semantics shifted by one pixel.** `hide-mobile` is `@media (max-width: 768px)` (inclusive); `ds-col-hide-md` is `@media (width < 768px)` (exclusive). At exactly 768 CSS px the previously-hidden columns now render. Noise-level.
8. **`tests/api/test_playerctx_endpoint.py:5-6` docstring contradicts the implementation** — it says "no auth gate (public NFL data, same posture as intel reads)" while the `_authed` fixture at :29 correctly notes the endpoint sits behind the session gate. The code is right; the docstring is wrong, and there's no test asserting the 401 for an anonymous caller.

---

PR #550, opened 2026-07-26
League Intelligence LI-1..LI-5: canonical config, registry fix, exact scorer, exact best-ball optimizer, value schema, replacement levels

jasonleetucker-code:
League Intelligence workstream (WS-E), batched per `docs/ORCHESTRATION.md` §2: LI-1 through LI-7 on one branch, one PR at the mid-week integration window.

**Nothing user-visible changes.** League-adjusted values stay unpublished until LI-7 validates them end to end; the only live-behavior changes are correctness fixes to team strength (before/after below).

---

## Headline: flex allocation must be measured on weekly outcomes, not means

The spec demanded *endogenous* flex allocation instead of assumed splits. Making the optimizer exact (LI-3) is what made it measurable. But **how** you measure it turns out to matter as much as measuring it at all.

Optimizing on `rosValue` (a season-long **mean**) erases the weekly spikes best ball actually pays for. Re-solving the same 21-slot vector over **actual 2025 weekly scoring** (158 team-weeks):

| | projection path (means) | weekly actuals |
|---|---|---|
| FLEX → TE share | **0.0%** | **10.4%** |
| FLEX → RB / WR | 50% / 50% | 50.6% / 37.6% |
| SUPER_FLEX → QB | 75% | ~87% |
| TE started / team | **2.00** | **2.215** |

An earlier revision of this PR reported `FLEX: WR 12, RB 12, TE 0` and "TE demand overstated 46%" as findings. **Both were artifacts of the point estimates** and are retracted. The honest statement is that a preassigned even split is still too *high* for TE (2.92 assumed) and the mean-based endogenous figure is too *low* (2.00) — the measured value is 2.215, between them.

### The dominant error was neither: asymmetric endpoints

The TE premium compares a 1-TE reference against our 2-TE league. Every earlier figure measured the **league** endpoint from data while **assuming** the reference was 1.0 TE/team. Re-solving the 1-TE vector over the same weekly scores: TE won **27.2% of FLEX** and teams started **1.608 TE/team**.

| basis | ref | league | median premium |
|---|---|---|---|
| assumed 1.0 / naive 2.0 | TE12 | TE24 | 1.239 |
| assumed 1.0 / actual 2.215 | TE12 | TE27 | 1.316 |
| assumed 1.0 / rostership 2.71 | TE12 | TE33 | 1.416 |
| **symmetric actuals 1.608 → 2.215** | **TE19** | **TE27** | **~1.12** |
| *KTC measured (external)* | | | *1.320* |

The structural demand change is **1.378×**, not 2.215×; an assumed reference overstates it by **1.61×**. **Operative structural premium ≈ 1.12** — a finding, not a published value.

> **⚠ The 1.316 row lands 0.004 from KTC and must never be cited as validation.** It pairs a measured league endpoint with an assumed reference, so the agreement is the assumption reflected back. This is the third such false corroboration in this workstream (see ADR-009).

Full derivation, caveats and retractions: `docs/league-intelligence/DECISIONS.md` ADR-009.

---

## LI-1 — Canonical league config + registry fix

`src/league_intel/config.py` — versioned canonical config (`configVersion` 1) from the provenance snapshot: all 141 scoring keys, the 21-slot structure (QB1 RB2 WR3 TE2 FLEX2 SFLEX1 K1 DL3 LB3 DB3), bench 37 (roster 58), teamCount 12, bestBall, playoffs 7@wk15, taxi/reserve 0, draft rounds 6, FAAB $100. `refresh_snapshot()` does one polite GET, diffs behavior-relevant fields, and on drift writes a **new dated snapshot + returns the diff** — the stored snapshot is never mutated.

**`config/leagues/registry.json` corrected to live truth**: TE 1→2, K added, DL/LB/DB 2→3, IDP_FLEX 2→0, rosterSize 30→58, taxiSize 5→0.

Consumers verified (`tests/league_intel/test_registry_consumers.py`): `league_registry` loader; `ros/scrape.py::_flatten_starter_slots` → `lineup.py`; `playoff_sim.py::_load_starter_slots`; `trade/team_impact.py` (K deliberately excluded from fit analysis); `server.py` draft-capital `teamCount`; frontend `useTeam.js`/`LeagueSwitcher.jsx` (teamCount only). `src/trade/suggestions.py::DEFAULT_STARTER_NEEDS` was a hardcoded mirror of the stale lineup — **fixed** (TE 1→2, DB 2→3).

## LI-2 — Exact scorer + golden validation

`src/league_intel/scorer.py` — pure `score_stat_line()` with per-key explainable components.

**Sleeper scoring is a pure dot product over shared stat keys.** 1,415/1,415 rostered 2025 player-weeks reconcile within 0.011, plus two full team totals. There are **no stacking rules to encode** (ADR-006 supersedes ADR-005's phrasing).

| Open question | Verdict |
|---|---|
| Pick-six | **Stacks** — `pass_int` + `pass_int_td` both charge (−6 at 2026 rates) |
| `bonus_fd_*` | **Is itself a precomputed stat key** = first downs gained |
| `rec` base + band | **Stacks mechanically**; nonzero-rate confirmation pends 2026 wk 1 |
| IDP multi-event | **All events stack** (sack + sack-yd + QB hit + TFL + solo) |
| `idp_blk_kick` vs `blk_kick` | **No double-count** — `blk_kick` is TEAM/DEF-only |
| Kicker | Pure per-yard (`fgm` rate 0) |

## LI-3 — Best-ball optimizer exactness (ADR-004 → ADR-007)

**Audit verdict:** a slot-ordered greedy claiming optimality "because per-slot decisions are independent". False in general — optimal only while slot eligibility sets form a **laminar** family. True here today, so it was **correct by an unstated, unenforced precondition**.

**The bigger bug: Sleeper evaluates eligibility against `fantasy_positions`, not `position`.** `_hydrate_overlay_players` read the NFL dump and kept only `position` — so **every hybrid IDP was locked out of half its legal slots in production**. Found empirically: reconstruction scored *below* the host on 5/10 weeks, and the diff was hybrids the host started.

Fixed: exact maximum-weight assignment (matroid greedy + augmenting paths, dependency-free) behind the same interface; `fantasy_positions` wired end-to-end; canonicalization pass keeps slot *labels* intuitive without changing scores.

**Validation:** brute-force equivalence, an explicit non-laminar counterexample, and **historical reconstruction — 10/10 real Sleeper best-ball team-weeks reproduce the host's awarded total**. 8/10 identical starter sets; the 2 that differ score identically, resolving the tie-handling question.

## LI-4 — Parallel value schema + single selector (backend only)

`src/league_intel/values.py` — `marketValue` / `consensusValue` / `leagueAdjustedDynastyValue` with schema+model+config+dataThrough stamps, and `get_active_value(player, mode, context)`.

The **no-op guarantee is enforced in construction**: `build_player_values` raises if anyone flips `LEAGUE_ADJUSTED_IS_NOOP` without supplying a validated model. Consensus is *read* from `rankDerivedValue` — `data_contract.py` untouched.

## LI-5 — Coverage fix + replacement levels (ADR-008)

`_positional_coverage` returned **exactly 100.00 for all 12 teams** — a constant contributing a flat 5 points to every composite. Now slot-derived, demand-weighted, eligibility-aware.

All 12 teams (`startLI3` = exact optimizer + hybrid eligibility):

| team | startBASE | startLI3 | Δstart | covBASE | covLI5 | BASE | LI5 | Δtotal |
|---|---|---|---|---|---|---|---|---|
| Type 2 DiaBaddies | 851.04 | 851.04 | 0.00 | 100.00 | 100.00 | 660.54 | 660.54 | 0.00 |
| I'm Not Afraid Anymore! | 720.33 | 731.52 | +11.19 | 100.00 | 98.41 | 552.68 | 560.48 | +7.80 |
| Russini Panini | 690.51 | 704.26 | +13.75 | 100.00 | 98.41 | 532.23 | 541.67 | +9.44 |
| Pop Trunk | 665.44 | 665.44 | 0.00 | 100.00 | 98.41 | 516.03 | 515.95 | −0.08 |
| CollinFoz | 643.10 | 643.10 | 0.00 | 100.00 | 98.41 | 492.91 | 492.83 | −0.08 |
| Rage Against The Achane | 636.87 | 636.87 | 0.00 | 100.00 | 94.05 | 490.01 | 489.71 | −0.30 |
| TyBWell | 561.89 | 565.70 | +3.81 | 100.00 | 100.00 | 429.74 | 432.49 | +2.74 |
| killaKich00 | 542.85 | 548.54 | +5.69 | 100.00 | 95.63 | 413.94 | 417.82 | +3.88 |
| Jasons archenemy | 522.95 | 522.95 | 0.00 | 100.00 | 98.41 | 405.37 | 405.29 | −0.08 |
| Chargers Team Doctor | 449.02 | 450.09 | +1.07 | 100.00 | 95.63 | 349.40 | 349.30 | −0.10 |
| ughb | 430.97 | 433.37 | +2.40 | 100.00 | 98.41 | 332.55 | 333.76 | +1.22 |
| jstuedle | 402.92 | 402.92 | 0.00 | 100.00 | 90.87 | 316.64 | 316.18 | −0.46 |

**6 of 12 teams get a better starting lineup** from the hybrid fix. Coverage now spans 90.87–100.00 instead of a constant.

**Honest limitation:** at 5% weight the coverage fix moves composites by ≤0.46. It is now correct rather than a lie, but still a weak signal.

**Downstream:** `teamRosStrength` feeds `pick_projection.py` and the playoff/championship sims. Composite values move, but **rank order is unchanged across all 12 teams**, so projected pick slots are unchanged today. Sim odds shift slightly. `test_pick_projection.py` injects synthetic values and pins no live ordering, so no test needed updating.

### Replacement levels + scarcity

Four tiers (starter / bestBallStarter / roster / waiver) off the real 12×58 pool (666 rostered), with smoothed ±2-rank bands.

**Six separate scarcity components**, not one score — QB `waiverScarcity` **0.75** vs RB **0.21** is the defining fact of a superflex league. Flagged as a surfacing candidate for LI-9.

Two deliberate deviations (ADR-008): unpriced players excluded from level pools; `waiverScarcity` measured against the best-ball starter floor rather than the noisy roster tail.

## LI-7 — Guardrails, confidence, explanation decomposition

`src/league_intel/adjustment.py`. Evidence tiers, not a scalar confidence: an axis with no admissible evidence contributes **exactly zero** and is arithmetically inert regardless of the factor supplied.

Three guardrails: **evidence gate** (ABSENT axes contribute 0), **magnitude cap** (±25%), **monotonicity** (`check_position_monotonicity` over a batch — order preservation is a set property; a per-row version could never fire).

**Confidence reflects reality**: with no permitted raw-category source (LI-6), *every* league-adjusted value rests on roster structure alone, so `STRUCTURAL_ONLY` is the explicit labelled default and `projection_corroborated` is False.

**Non-duplication pinned on exact arithmetic**, never a tolerance. The TE axis is `ABSENT` specifically so it cannot stack on the blend's existing ×1.15 alignment multiplier.

---

## Data-quality findings — handed off, NOT fixed here

**Owned by the identity / data-quality workstream (WS-H):**

- `data/ros/aggregate/latest.json` contains **6 duplicate player rows** — `nate landman`, `cam skattebo`, `cam ward`, `tank dell`, `cam bynum`, `mitch tinsley`, each present as both lowercase and Title Case, with ROS values split across two rows.
- 16 rows carry non-lowercase `canonicalName`; **any join against it is lossy** — 40 of 666 rostered players fail to match.

## Reproducibility caveat

The exact optimizer and the `fantasy_positions` eligibility fix exist **only on this branch**. Any measurement depending on them — including the flex-allocation and TE-demand figures above — is **not reproducible on `main`** until this merges.

## Verification

- `pytest tests/league_intel/` — **249 passed**; `tests/ros/` green (pre-existing slot-labelling tests pass unmodified)
- Full suite `-m "not livedata"` — **3234 passed, 280 deselected, 4 subtests** on the LI-5 tree
- ruff 0.6.9 `format --check .` clean repo-wide; `check` clean on all changed files
- ADRs 006–009 recorded; STATUS + TASK_REGISTRY current


jasonleetucker-code:
## Merge blocker: the headline flex-allocation finding is a measurement artifact

CI is green and the LI-1..LI-5 work is sound, but **the number this PR leads with is wrong**, and it's load-bearing — every replacement level and scarcity figure below it rests on the measured allocation. Flagging before the Jul 29 integration window rather than after.

### The defect

`measure_endogenous_starters` runs the exact optimizer on `rosValue` — a **season-long 0-100 mean**. On a point estimate a TE can only win a flex slot if its *average* beats the best spare RB/WR, which essentially never happens. Hence `TE 0`. But best ball pays for weekly spikes, not averages: a TE averaging 8 PPG who posts 22 twice wins those two weeks outright and wins nothing on a season mean. The optimizer is exact; the input collapses the variance the format monetizes.

`replacement.py`'s docstring currently states this artifact as a finding — *"FLEX never takes a TE at all"* — and the PR body converts it into a claimed 46% overstatement in the even-split model.

### Measured on actual scoring

170 real 2025 team-weeks (all 17 weeks, league `1180092661344120832`), same exact optimizer with the `fantasy_positions` fix, re-solved under the current 21-slot vector:

| slot | filled by |
|---|---|
| FLEX ×2 | RB 172 (50.6%) · WR 128 (37.6%) · **TE 40 (11.8%)** |
| SUPER_FLEX ×1 | QB 140 · WR 18 · TE 7 · RB 5 |

- **TE started per team-week: 2.28** (median 2, max 5) — not 2.00
- **20.6% of team-weeks start 3+ TEs**
- TE rostered per team: 5.10

The same data under the **old 1-TE vector** gives TE **29.1%** of flex — that's the isolated rules effect (TE2 was flex-eligible, now slot-locked), and it's why 2025 lineups can't be read directly without re-solving.

So the even-split model's TE figure of 2.92 is still too high, but the endogenous figure of 2.00 is too low. The real number is 2.28, and TE flex-worthiness is ~12%, not zero.

### Roster-era bias — now empirical

Separately settled: live 2026 rosters (built knowing the 2-TE rule) carry **5.42 TE/team** vs **5.10** in 2025 — a +0.32 (~6%) behavioral response. Direction of the bias is **up** (the 2025 pool is slightly thinner, so TEs look scarcer than they now are). Magnitude is small and within noise at n=12; treat the sign as settled, the size as minor.

### Caveats on my own measurement

Stated so they aren't lost: single season; 2026 rules applied counterfactually to 2025 rosters and scores; **170 of ~204 team-weeks used** — entries without enough rostered players for a 21-slot solve were skipped, and that exclusion needs checking for correlation with roster construction (if thin rosters drop out, the sample skews deep); missing `players_points` treated as 0. Also note the exact optimizer + `fantasy_positions` fix exists **only on this branch** — the measurement is not reproducible on `main` today.

### What's needed before merge

1. Correct the PR headline table and the `replacement.py` docstring.
2. Reconcile `starters_per_team` for TE against 2.28 and restate the downstream replacement/scarcity numbers.
3. ADR the durable fix for the projection-vs-actuals gap — either calibrate the depth constant from actual weekly outcomes, or sample weekly outcomes around projections using the Gaussian model already in `playoff_sim.py:248-265`.

The harness and the full 2025 fixture set are already in the workstream worktree (`scripts/measure_flex_allocation_actuals.py`, `tests/league_intel/fixtures/matchups_2025/`). The LI agent has this and is working the reconciliation now.

---

jasonleetucker-code:
## Follow-up: defect confirmed, but my reconciliation ask was wrong on two counts

Addressed in `b6ec0ab6`. Correcting my own comment above, since the record on this PR is what ships.

**1. The artifact is real and independently reproduced.** Re-solving the 21-slot vector over actual 2025 weekly scores: FLEX TE share **0.0% → 10.4%**, TE started/team **2.00 → 2.215** (my pass measured 11.8% / 2.28 — close, and the two were derived independently).

**2. My `3.79 → 2.28` mapping was invalid, as flagged.** `3.79` was never `starters_per_team` — it was *marginal-weighted effective depth* from the marginal best-ball probe (how many TEs carry value, not how many start), and it had already been retired for a 2.08× churn confound. What actually feeds `replacement.py:576` is `starters_per_team`, which for TE was **2.00**. The real before/after is **2.00 → 2.215**. Anyone still quoting 3.79 (including my comment above) should stop.

**3. The dominant error was neither of the above.** The premium compares a 1-TE reference against our 2-TE league. Every figure to date — mine included — measured the *league* endpoint from data while **assuming the reference endpoint was 1.0 TE/team**. It isn't. Re-solving the 1-TE vector over the same weekly scores: TE won **27.2%** of FLEX, teams started **1.608 TE/team**.

| basis | ref | league | median | TE1-12 |
|---|---|---|---|---|
| assumed 1.0 / naive 2.0 | TE12 | TE24 | 1.239 | 1.175 |
| assumed 1.0 / actual 2.215 | TE12 | TE27 | 1.316 | 1.214 |
| assumed 1.0 / rostership 2.71 | TE12 | TE33 | 1.416 | 1.252 |
| **symmetric 1.608 → 2.215** | **TE19** | **TE27** | **1.121** | **1.082** |
| *KTC measured* | | | *1.320* | *1.227* |

Structural demand change is **1.378×**, not 2.215×. An assumed reference overstates it by **1.61×**. **Operative premium ≈1.12** — below every prior figure.

The `1.316` row lands 0.004 from KTC's 1.320. **That is not validation** and shouldn't be cited as such: it pairs a measured league endpoint with an assumed reference, so the agreement is an artifact of exactly the asymmetry being corrected.

One thing this does enable: if structure warrants ~1.12 and KTC charges 1.32, the residual ~1.18× is plausibly the **scoring** component of KTC's TEP — the first quantitative support for the axis ambiguity we've been unable to separate. Suggestive, not established.

### Bias checks, both closed

- **Roster-era**: direction **up** (2026 5.42 vs 2025 5.02 TE/team) — premium marginally overstated, within noise at n=12. The earlier "downward" claim in this workstream is retracted.
- **Exclusion**: the skipped short-roster team-weeks carry *more* TEs (6.17 vs 5.02) and are short at K/IDP, so the sample skews TE-shallow and demand is if anything **understated** — the bias runs the safe way.

### Durable fix

Option (a) adopted and ADR'd: calibrate the depth constant from actual weekly outcomes. Depth is a league-structure constant, history is the right input, and it avoids stacking a second fitted layer — the Gaussian model in `playoff_sim.py:248-265` is itself an approximation, and deriving a structural constant through it would embed its error. Option (b) stays right for forward-looking per-player variance. Constant is recomputed, not frozen.

### Still caveated

Single season; 2026 rules applied counterfactually to 2025 rosters; the exact optimizer + `fantasy_positions` fix is branch-only, so **none of this is reproducible on `main` today**. The largest remaining lever is no longer a measurement but an assumption: the reference endpoint presumes KTC's standard board targets a league like our 2025 one (superflex, 2 FLEX). A generic 1-TE league with a single flex slot would give TEs less flex opportunity, lower the reference, and raise the premium again.

249 tests pass, ruff clean, nothing published. The headline table and `replacement.py` docstring in this PR still need correcting before the Jul 29 window.

---

PR #551, opened 2026-07-26
Redesign R3: dashboard, news, and market surfaces

jasonleetucker-code:
## What this is

R3 of the redesign: **the dashboard, the news wire, and the two market surfaces** — rebuilt on the R0 design system and the R1 shell, on top of R2&#39;s base (this PR targets `main` but branches from `claude/redesign-r2-rankings`; merge R2 first).

Four surfaces, four commits. Behaviour is preserved — with **two exceptions the fresh-eyes review caught, both since restored**, and **one regression I introduced while fixing the second and then fixed properly**; see [Review follow-ups](#review-follow-ups-fresh-eyes-review-of-622117cb) before taking the preservation claim at face value.

## `/news` — digest-first wire

Opens on **By player** rather than the raw feed, because that&#39;s the data&#39;s own shape: `src/news/digest.py` emits a digest once a player has 2+ stories inside the hard 7-day window, so digest-first isn&#39;t a skin, it&#39;s the model surfaced.

- Header states the 7-day window once; a five-tile coverage strip (stories, players covered, alerts, on-your-roster, sources live) answers the &#34;at a glance&#34; question before any scrolling.
- One flush `Panel` holds controls + both feeds. `Tabs` for the view (real tabpanel), `SegmentedControl` for scope (3 values), `Select` for team/position/source (&gt;5 values each) — per the DS&#39;s own control-choice rule.
- **Two deliberate feed idioms**, not one card list: digest rows use a fixed identity rail (player · pos/team · story count · age) so the eye tracks a column; story rows are dense meta/headline/body/mention stacks with a gold inset edge for roster-relevant items.
- Preserved: scope relevance scoring, search, the conjunction-per-mention facet rule, ambiguity-safe player linking, digest facet parity, the 100-item cap, fail-fast with no fixture fallback.

## `/edge` — market intelligence

The pre-R3 page stacked eight equally-weighted sections in one wall of cards with nothing telling you where to look. The signals actually fall into three families, so they&#39;re grouped that way behind a tablist and the actionable one leads: **Market gaps** (offense vs KTC + IDP vs IDPTC) → **Agreement** (consensus anchors, widest disputes) → **Data caution** (flags, single-source).

Six StatTiles, the value-vs-spread scatter in a proper Panel with its edge-zone explanation above the plot, each section a Panel whose body is an edge-to-edge DataTable with methodology in the footer.

One genuine improvement: the IDP buy/sell sections gain a visible **Gap** column — that magnitude already drove their ordering, it was just buried in a `title` attribute.

## `/finder` — arbitrage blotter

Workflow presets become a real tablist; the results table becomes a DataTable whose columns follow each preset&#39;s own disclosure flags; the expand control moves into a footer that states what the limit means. `defaultSort` mirrors each workflow&#39;s comparator so initial order is identical to the pre-R3 table while headers become genuinely sortable. IDP league gating and its stale-selection snap-back preserved exactly.

Shared column specs live in `edge/edge-columns.jsx`; `/finder` imports them and adds only its three genuinely different cells.

## `/` — the war room

Restructured so the page answers one question in reading order: **what changed, and what do I do about it?**

1. **Command bar** — identity + team value + the four deltas + tier mix + value trend
2. **Ticker** — the market moving underneath you
3. **Signals rail** — the specific &#34;act on this&#34; list for YOUR roster; nothing sits above it but context
4. **Working grid** — diagnostics left, market movement centre, intel + jumping-off right

Column assignment is by information role, not by what fits: explicit at every breakpoint (3 → 2 → 1), with the right column folding under the pair at tablet rather than orphaning. No auto-fit card wall.

**The third container system dies here**: `components/terminal/Panel.jsx` is deleted and all eight consumers move to the ds `Panel`, along with its `.panel--*` hook classes. `TeamCommandHeader` owns the page&#39;s single `h1` via `PageHeader`; `StaleBanner`&#39;s two hand-rolled states become ds `Banner`s with real live-region wiring; `TopSignalsRail` becomes a Panel of signal cards with `Badge` + `Movement`.

## Cross-cutting

- **Eight hand-rolled tables gone.** The audit found **zero** `aria-sort` in the app; every sortable header on these surfaces now has it. These were static tables, so no sort semantics changed — only added.
- **Every ASCII direction glyph on these surfaces is gone** — the five ▲/▼ sites (movers, market movement, watchlist, value chart, signals rail) and the › caret now use ds `Movement`/`Icon`, so direction carries an announced label instead of colour plus a character.
- **Zero inline colour/spacing/font/radius** across all four surfaces. The eight remaining inline styles in the terminal are computed layout math (percentage widths, flex ratios, the Sparkline&#39;s prop-driven dimensions), which the design system explicitly permits.
- Structure moved to token-driven CSS modules (`news.module.css`, `edge.module.css`, `finder.module.css`, `terminal.module.css`). Per-panel content classes and the pos/conf badge classes stay global until the R5 purge — the same call R2 made for the board.
- **One ds API addition, plus a new primitive** (was: &#34;no ds API changes&#34; — corrected, see follow-ups): `Panel` gains a **controlled** disclosure (`collapsible` / `collapsed` / `onToggleCollapsed` / `bodyId`), and a new `CollapsiblePanel` owns the state. Purely additive; every existing call site is untouched. Nothing else was added or mutated. See [the server-safety fix](#server-safety-fix) for why the state lives in a wrapper.
- `nav-model.js` untouched — all four routes were already registered correctly.

## Review follow-ups (fresh-eyes review of `622117cb`)

The review verified the preservation claim held broadly, and found two places where it did not. Both are behaviour the CSS-module migration lost silently, and both are restored in `a7f2f0d1` / `4cccfd84`.

**The dashboard&#39;s mobile stacking order regressed.** Below 768px the three-column grid collapses to one. `main` dissolved the column wrappers (`display: contents`) so panels could reorder *across* columns, then gave each an explicit `order`. The migration kept the single-column media query but dropped both halves — so panels stacked in raw DOM order, putting Portfolio and Scouting (the two least urgent surfaces) first and pushing Movers/Signals two panel-heights down. That inverts the reading order this whole page is built around. Restored `display: contents` inside the `&lt;768px` block plus `main`&#39;s exact scale (signals 10, movement 20, news 30, portfolio 40, scouting 50, actions 60), and re-tagged the six components with the `panel--*` hook classes those rules target — `PortfolioSummary` at all three call sites, `ScoutingIntel` at both. Movers/Watchlist stay unordered on purpose: an unset `order` is 0, so they lead the stack.

**…and then the fix itself was inert, which self-review caught (`4cccfd84`).** A media query adds **no specificity**, so `.col { display: contents }` inside `@media` and the bare `.col { display: flex }` are both (0,1,0) and source order alone decides — and the base rule was written *after* the media block. Flex won at every width, so the restoration would have rendered exactly like the regression it was meant to fix while reading correctly in the diff. Same trap one level down: `.colRight` takes `display: grid` at &lt;1320px and that rule is still in force below 768px, so `.col` alone left the right column an intact box its three panels could not escape. The section is now ordered base → tablet → mobile, and the mobile block names `.colRight` explicitly.

**`ScoutingIntel` lost its collapse toggle.** It shipped `collapsible defaultCollapsed` on `main`; moving it onto the ds `Panel` dropped the toggle silently because the primitive had no such prop. The behaviour is restored: the toggle is a real `<button>` wrapping the heading with `aria-expanded` + `aria-controls`, collapsing sets `[hidden]` so the body leaves the a11y tree rather than merely disappearing, and it no-ops without a `title` since a disclosure with no label is unusable. The **state** lives in a new `CollapsiblePanel` rather than in `Panel` — see below.

<a name="server-safety-fix"></a>
### …and my first version of that fix broke `Panel`&#39;s server-safety (`0b36074c`)

Adding `collapsible` put `useState` + `useId` into `ds/Panel.jsx` **without a client directive**. On `origin/main` Panel is a pure function, so it can be rendered from a React Server Component; that change silently made it client-only.

Why it matters concretely: `app/league/shared-server.jsx` is server-safe *by contract* (&#34;no hooks … so they can be rendered either from a React Server Component or a Client Component&#34;), **seven Server Components import it**, and four render its `Card` — the same legacy `Card` that ds `Panel`&#39;s own docblock names as one of the three container systems it replaces. The R5 `/league` migration would have walked straight into it.

The fix moves the **state**, not the markup:

| | |
|---|---|
| `Panel` | renders the disclosure, **controlled** (`collapsible` / `collapsed` / `onToggleCollapsed` / `bodyId`); stays hook-free |
| `CollapsiblePanel` | `&#34;use client&#34;`, owns `useState` + `useId`, composes `Panel` |

I chose this over the one-line `&#34;use client&#34;` on `Panel` deliberately. That would make the symptom disappear while pushing client JS onto seven currently-server `/league` routes — and `/league` has **2.2 KB of headroom** against its 170 KB budget, the tightest in the app after `/trade`. It would also quietly convert *the* container primitive into a client-only component, which is a real architectural change to a frozen shared contract.

**Reviewers should know: I could not reproduce the runtime failure.** Those seven routes are dynamic and data-backed, this container&#39;s backend has no data, so they never render during `next build` — which compiles either way. The defect is **reasoned** from React&#39;s server/client module rules plus the fact that `Panel` on `main` is hook-free; it is **not observed**. I treated it as real because the cost of being wrong is asymmetric, but it has not been reproduced and should not be described as if it had.

New `frontend/__tests__/components/ds/panel-server-safe.test.js` pins the contract that previously lived only in a docblock — which is precisely why nothing caught this. It walks `Panel`&#39;s **server** module graph: a dependency carrying a client directive is a boundary and stops the walk, since an RSC may freely render a client child (`Icon` is one, and that is fine). I got that rule wrong on the first pass — asserting no client component anywhere in the graph — and the test failing is what corrected it. Verified non-vacuous against **both** escape routes: re-adding the hooks fails 2 assertions, and adding `&#34;use client&#34;` to `Panel` fails 3.

**Coverage for the follow-ups.** `frontend/__tests__/components/ds/Panel.test.jsx` (11 tests) pins the split — the default path unchanged, the disclosure as a real announced control, `Panel` behaving as a *controlled* component, and `CollapsiblePanel` owning the state. `frontend/__tests__/terminal-mobile-order.test.js` (6 tests) asserts the CSS contract directly: jsdom evaluates neither media queries nor `order`, so what can be pinned is that the rules exist, target the right panels, carry the right relative priority, **and win the cascade** — which is exactly what regressed, twice. The two cascade assertions were verified non-vacuous by running them against the pre-fix stylesheet: they fail there and pass here.

**Not fixed, deliberately:** the ~90 lines of orphaned `.panel*` CSS in `globals.css`. Verified the shape of it rather than assuming: `.terminal-grid` / `.terminal-col` are genuinely dead (no JSX consumer outside the stylesheet), but `.panel-tabs` / `.panel-tab` are **live** in `PlayerMarketMovement`, `BuySellHold` and `TeamNewsFeed`, and the block&#39;s `.panel--*` `order` rules are live global class names paired with an `order: 0` reset at ≥720px that keeps desktop correct. So it is a block that must be deleted as a *unit*, not swept — a purge, not a review fix. It belongs to R5 where the global-class sweep already lives, and is planned in `docs/redesign/R5-PANEL-CSS-PURGE.md` so it isn&#39;t rediscovered as a surprise. The module-scoped rules added here are self-contained and do not depend on any of it surviving.

## Verification

- **vitest: 57 files / 1186 tests green** (was 53/1159 when this body was first written; +27 from the follow-ups above)
- **`next build` green; all 13 bundle budgets green**
- `/edge` 22.8 → 22.3 KB · `/finder` 9.1 → 11.5 KB · `/rankings` unchanged

### Flag 1 — E2E specs not run (honest gap)

`@playwright/test` isn&#39;t installed in this container, and installing it would touch the coordination-controlled lockfile, so I declined to work around it. I did boot a full stack on private ports (8100/3100 — 3000/8000 were occupied by a concurrent agent) and confirmed `/news` and `/edge` render their R3 structure server-side (`ds-panel`, `ds-page-header`, `news-controls`, tabs). `/` and `/finder` are auth-gated, so SSR shows only the shell — expected, not a defect. **I&#39;d rather report this honestly than claim a green run I didn&#39;t get.** journey-news + a dashboard pass are worth running at the integration window.

The E2E SEL registry is extended additively (`news-controls`, `news-digest-row`, `news-story-row`, edge/finder tab + row selectors, dashboard command bar + signal cards). The earlier note that R4&#39;s `NAME` export &#34;isn&#39;t on this branch, so there&#39;s nothing to collide with&#34; is now **stale**: the collision was real, and the E2E workstream — which owns that registry — resolved it in `21222e76`, landing one SEL registry both redesign branches carry. That commit is on this branch.

### Flag 2 — dashboard bundle got smaller

**`/page` 72.2 → 69.8 KB** (20.2 KB headroom against the 90 KB budget). The rebuild came in *under* the pre-R3 size — the legacy Panel dying plus ASCII glyphs and inline styles collapsing into shared ds primitives more than paid for the new structure. **No budget bump needed.**

---


jasonleetucker-code:

Reviewed commit `622117cbef501146c6729f8949c2c016a3c4f697` (R3: finish the dashboard token migration) against `origin/main` @ `9ccdecea`. I did not write this PR. I tested the preservation claim rather than accepting it.

**Verdict:** the "preserved verbatim" claim holds for all the data logic I checked. Two real behavior losses on the dashboard that the claim doesn't cover, and one merge hazard shared with #552.

---

### Verified clean (tested, not assumed)

**`/finder` `defaultSort` — correct for all five presets.** This was the flagged off-by-one risk, so I checked the null-handling boundary specifically. `DataTable.sortRows` sinks nulls last in *both* directions (`components/ds/DataTable.jsx:128`), which is not what `?? 0` / `?? Infinity` do — so equivalence depends on nulls never reaching the comparator. They can't:
- `isEligibleForAnalysis` requires `!!row.rank` (`lib/display-helpers.js:66`), so every rank-sorted preset (`stable-idp`, `single-risk`, `all`) has a truthy rank.
- Both spread-sorted presets filter `(r.sourceRankSpread ?? 0) > 15` / `> 10` before sorting, so `sourceRankSpread` is non-null in the sorted set.

Ties fall through to pre-sort index, and `results` is already sorted by `workflow.sort` before slicing — so the rendered order is identical to pre-R3 for every preset.

**`/edge` — all eight sections survived the regroup intact.** Every filter predicate, threshold, comparator and limit is byte-equivalent (`EDGE_SECTION_LIMIT` 15, `EDGE_PREMIUM_LIMIT` 10, `EDGE_CAUTION_RANK_LIMIT` 300, `EDGE_PREMIUM_RANK_LIMIT` 250, `PREMIUM_SUMMARY_SPREAD` 20, `IDP_TOP_LIMIT` 200). Per-panel `defaultSort` matches each section's own comparator. The IDP panels deserve a note: `idpGapOf()` returns `0` on regex miss, and `0` is not null, so it sorts last under `desc` — exactly where `b.diff - a.diff` put it before. Correct.

**Deleted terminal `Panel` — all 8 consumers migrated.** BuySellHold, PortfolioSummary, PlayerMarketMovement, WatchlistPanel, ScoutingIntel, MoversPanel, QuickActions, TeamNewsFeed. No `terminal/Panel` imports remain anywhere.

**Dashboard contracts intact.** `useTerminal` cache priming uses an identical key in `TerminalLayout.jsx:54-58` and `TeamCommandHeader.jsx:113-117`, so the priming no-op still holds. Signal-dismissal precedence (per-league bucket over legacy flat map, `TopSignalsRail.jsx:110-118`) is untouched. `TeamCommandHeader`'s server-first / local-fallback precedence for `totalValue`, `tierCounts`, `delta7/30/90/180` is unchanged.

**`/news`** — identical lib imports, identical filter pipeline order (scope → player facets → source → search), `MAX_ITEMS` still 100. The dropped `usePlayerMeta` was a one-line wrapper, now inlined identically.

**No business logic moved client-side.** No rank/value recomputation in any new component; backend stamps render verbatim.

**Tests + build:** `npx vitest run` → **54 files / 1165 tests passing**. `npm run build` clean, **all bundle budgets pass** (`/page` 69.6/90 KB, `/edge` 22.3/30 KB, `/finder` 11.5/20 KB).

---

### P2 — mobile + tablet dashboard panel order regressed

`components/terminal/terminal.module.css:121-131`, `components/terminal/TerminalLayout.jsx:67-82`

Pre-R3, `.terminal-col` was `display: contents` below 720px and the panels carried explicit `order` values, deliberately floating the actionable surfaces above the diagnostics. main's own `TerminalLayout` docstring said so:

> mobile (<720px): single column; panels re-ordered via CSS `order` so the most actionable surfaces (signals, market movement) land above portfolio/scouting diagnostics.

New `.col` is `display: flex` at every breakpoint, so at `<768px` the columns stack in DOM order and the `order` rules are gone:

| | old mobile order | new mobile order |
|---|---|---|
| 1 | Movers | **Portfolio** |
| 2 | Watchlist | **Scouting** |
| 3 | Signals (BuySellHold) | Movers |
| 4 | Market Movement | Market Movement |
| 5 | News | Signals |
| 6 | **Portfolio** | Watchlist |
| 7 | **Scouting** | News |
| 8 | Quick Actions | Quick Actions |

Failure scenario: a manager opens the dashboard on a phone to see what to act on today. Portfolio allocation and Scouting diagnostics — the two longest, least actionable panels — now occupy the entire first screen, and Movers/Signals are pushed roughly two panel-heights down. Tablet changed too: `<1320px` now leads with the left diagnostics column, where the old `grid-template-areas: "center right" / "left right"` put market movement top-left.

`TopSignalsRail` still sits above the grid, so the very top of the page is still actionable — that's what keeps this P2 rather than P1. If the reordering was intentional, worth saying so in the PR body, since it reverses a documented decision.

### P2 — `ScoutingIntel` silently lost its collapse affordance

`components/terminal/ScoutingIntel.jsx:152-155`

main passed `collapsible defaultCollapsed={false}` to the terminal Panel (`ScoutingIntel.jsx:152-157`), which rendered a labelled `aria-expanded` toggle. The ds `Panel` has **no** `collapsible` prop (`components/ds/Panel.jsx:32-44` — I grepped the whole `ds/` directory; the concept doesn't exist there), so the prop was dropped and the toggle is gone with no replacement. ScoutingIntel was the only consumer using it.

This compounds with the finding above: Scouting is simultaneously the longest diagnostics panel, newly promoted to position 2 on mobile, and no longer collapsible. A phone user has to scroll past all of it, every visit, with no way to fold it away.

### P2 — `tests/e2e/helpers/journey.js` will conflict with #552, and the obvious resolution breaks the whole e2e suite

Both PRs insert into `SEL` at the same anchor (immediately after `navSearchButton`). I ran the actual three-way merge:

```
$ git merge-file -L R3 -L base -L R4 merged.js base.js theirs-r4.js
MERGE EXIT=1   → conflict at lines 42-102
```

The dangerous part is *where* the conflict boundary falls. R4's side of the conflict block contains both its selectors **and** the closing `};` of `SEL` **and** the entire `const NAME = {...}` declaration. But R4's separate edit to `module.exports` (adding `NAME,`) is a different hunk and **auto-merges cleanly** — I confirmed `NAME,` lands at line 237 of the merged file regardless.

So if whoever merges second resolves the conflict by keeping the R3 block and dropping what looks like a stray duplicate brace, they get a file that exports `NAME` without defining it → `ReferenceError: NAME is not defined` at `require()` time → **every Playwright spec that imports the journey helper fails**, not just R4's. Unit tests won't catch it; only the regression job will.

No key collisions between the two selector sets, so the correct resolution is purely mechanical — just keep both blocks and R4's `NAME` object. Worth a note in whichever PR merges second.

### P3 — ~90 lines of orphaned `.panel*` CSS left behind, with two live rules buried inside

`app/globals.css:4734-4844` and `5234-5260`

Nothing emits these any more: `.panel`, `.panel:hover`, `.panel--bare`, `.panel--collapsed`, `.panel-head`, `.panel-head-text`, `.panel-title`, `.panel-subtitle`, `.panel-head-actions`, `.panel-collapse-btn`, `.panel-body`, `.panel-body--flush`, the six `.panel--portfolio/scouting/movement/signals/news/actions` order rules, and `.terminal-grid` / `.terminal-col*`.

The landmine: `.panel-tabs` (4822) and `.panel-tab` (4830, 6545) **are still live** — `PlayerMarketMovement.jsx:137` and `TeamNewsFeed.jsx:52` both use them. They sit in the middle of the otherwise-dead block, so a future "purge `.panel*`" sweep will silently break the window/scope tab strips on two dashboard panels. Either rename those two to a ds-owned class now, or leave a comment marking them live.

### P3 — `MoversPanel` rows are still non-focusable clickable divs

`components/terminal/MoversPanel.jsx:49-52`

`<div className={styles.moverRow} onClick={...}>` — no `role`, no `tabIndex`, no key handler. Expanding a mover to see its source breakdown is keyboard-unreachable. Identical to main, so not a regression — flagging only because the PR invests in a11y elsewhere and this row sits in the middle of the rebuilt surface.

---

PR #552, opened 2026-07-26
Redesign R4: draft war room + trade surfaces

jasonleetucker-code:
## What this is

R4 of the redesign: **the draft war room and every trade surface** — `/draft`, `/trade`, `/trades`, `/angle`, `/waivers` — rebuilt on the R0 design system and the R1 shell. Branched from R2 (`bf15133b`) for its `DataTable`/`Dialog` extensions; merge after R2.

**Scope of the preservation claim — read this before the diff.** Existing business logic is untouched: `trade-logic`, `waiver-logic`, `draft-logic`, `league-analysis`, `trade-retro-value` and every `/api/*` response render verbatim. No value math moved client-side, and values are read exactly the way the existing pages read them, so the LI-4 `getActiveValue` swap stays a clean substitution.

**But this PR is not purely a rebuild.** One section is brand-new user-facing surface with no pre-R4 behaviour to diff against, and it deserves reviewer attention rather than a skim:

> **`/waivers` FAAB v2 contention is a NEW FEATURE** — ~130 lines (`components/waivers/FaabRecommendation.jsx:98-230`) rendering a clearing price, a projected top rival, and a per-opponent bid table. The backend has returned `contention` since #538, but no main-branch frontend file referenced `contention`, `perOpponent`, `topRival` or `clearing`. This PR changes zero Python files; it is the first UI over an engine that had none. It renders bid guidance a manager will spend real FAAB against — review it as new code, not as a migration.

## Surfaces

| Surface | Bundle | Headline |
|---|---|---|
| `/waivers` | **unmeasured → 30.3 KB** | FAAB v2 contention gets its first UI (**new feature**) |
| `/trades` | 15.6 → **14.2 KB** | Winners/losers card wall → sortable table |
| `/angle` | 14.5 → **12.9 KB** | Ranked packages → sortable arbitrage table |
| `/trade` | 73.0 → **67.1 KB** | 3,044 L decomposed; headroom 2.0 → 7.9 KB |
| `/draft` | 124.9 → **122.4 KB** | 4 hand-rolled modals → ds Modal; headroom 3.1 → 5.6 KB |

**No budget bump needed anywhere.** Both pages flagged as near-budget came back with *more* headroom than they started with, matching R2&#39;s outcome.

### `/waivers` — the claim desk

- **Contention block** (new): clearing price + projected top rival as StatTiles, then a per-opponent DataTable (estimated bid, their FAAB, positional need, aggression) with `balanceUnknown` / `lowSample` / `intel` flags. Rivals excluded from the clearing math are dimmed; the recommender&#39;s own `notes` carry the prose.
- **Factors are disclosed, not dumped** — the 11-row breakdown and league FAAB history both sit behind toggles.
- **Freshness footer** from `inputsAsOf`/`staleInputs`, so a bid resting on day-old rosters admits it.
- Rival flags are plain Badges, not Tooltips: a tooltip needs a focusable child, and three per row across a league would add ~30 tab stops to a read-only table.

### `/trade` — the terminal

3,044 lines split into `page.jsx` (state + orchestration), `trade-sections.jsx`, and `trade.module.css`, mirroring R2&#39;s rankings decomposition. Simulation impact moves onto StatTiles + Movement; suggestion category buttons get real tab semantics; rail cards become focusable buttons with visible focus rings (they were unfocusable divs).

### `/draft` — the war room

Four hand-rolled modals → ds `Modal`; StatsStrip → StatTile grid; all 11 remaining legacy `.card` containers → `Panel` (zero remain on the page). Workspace/localStorage contract, live Sleeper sync, inflation + bid-ceiling math, alert thresholds, triage behavior and all keyboard shortcuts preserved.

## Half-wired seams found

1. **FAAB v2 contention had no UI** — engine shipped in #538, frontend still on v1 pills. Addressed here as a new feature (see scope note above).
2. **`/waivers` had no bundle budget** — shipping unmeasured since it was created. Added (30.3 measured → 36 pinned).
3. **`/draft`&#39;s four modals had `role=&#34;dialog&#34;` but no focus trap, no focus restore, and inconsistent Escape.** A genuine a11y defect that was invisible in code review because the ARIA attribute made it *look* correct.
4. **13 dead imports in `/trade`** — nine were dead *before* this PR (`MonteCarloButton`, `ValueBandBadge`, `PlayerImage`, `ResilientSection`, `SHARE_PARAM`, `meterVerdict`, `percentageGap`, `tradeGapAdjusted`, `multiTeamAnalysis`, `valueAdjustmentFromSideArrays`, plus three from `dynasty-data`).

Two of my own mistakes, caught and fixed in-branch and called out because they&#39;re the kind a large decomposition loses quietly:

- I **dropped the sticky tray&#39;s verdict bar** during the `/trade` extraction (the gradient track + marker driven by `verdictBarPosition`). Restored on tokens, marker offset passed as a CSS custom property since that&#39;s layout math.
- I **weakened a spec**: replacing `/trades`&#39; `&#34;Loading trade data...&#34;` sentinel with a skeleton made its `waitForFunction` vacuous, and since the page title now renders immediately the follow-up assertion would pass *before any data landed*. Retightened to wait on a real ledger entry or the explicit empty state.

## Review follow-ups (fresh-eyes review of `79bde6f6`)

The review found two of this PR&#39;s own claims didn&#39;t survive checking. Both are now corrected — one in code, one in this body.

**`aria-sort` on `/draft`&#39;s main board — the claim was false, and is now true.** The board&#39;s nine sortable columns went through a local `th()` helper rendering a bare `<th>` with `cursor: pointer` and an `onClick`: no `<button>`, no `tabIndex`, no `role`, no `aria-sort`, with direction conveyed only by a `▲`/`▼` glyph. A keyboard-only user could not sort the largest table on the page at all — Tab skipped the header and Enter/Space did nothing. The helper was byte-identical to main, so R4 preserved it faithfully and introduced nothing; the defect was the *blanket* a11y claim, which held for every surface that moved onto `DataTable` and not for this one. `th()` now renders a real `<button>` inside the `<th>` with `aria-sort` on the `<th>` and the shared `.ds-table__sort` glyph rotating off it. `frontend/__tests__/components/DraftBoardSort.test.jsx` (4 tests) pins keyboard operability and the `aria-sort` state machine.

**FAAB contention scoping + coverage.** The review&#39;s point that ~130 lines of new surface arrived inside a PR framed as &#34;preserved verbatim&#34; is fair — this body now leads with that instead of burying it. Coverage added: `frontend/__tests__/components/FaabContention.test.jsx` (8 tests) covering the `skipped` branch, `balanceUnknown` / `lowSample` / `intel` flag rendering, the `need|surplus|neutral` domain, and the clearing/top-rival tiles.

**P3 — Dialog focus restore on a detached opener.** Reachable on `/draft`: submit the mark-drafted modal and the player leaves the pool, so the opener unmounts before the restore fires and focus falls to `document.body`. `Dialog.jsx` now guards with `isConnected` and falls back to the `#main` landmark; 2 tests added to `frontend/__tests__/components/ds/Dialog.test.jsx`.

**P3 — `waivers-smoke` first assertion was vacuous.** `/Waivers/i` against `body` matched the nav label on every authenticated page, so it passed even if the route rendered nothing but chrome. Now anchored to the page&#39;s own `<h1>`.

**P2 — the `journey.js` SEL/NAME merge trap.** Confirmed and handed to the E2E workstream, which owns that registry; resolved in `1563e527`, which lands one SEL registry both redesign branches carry. The trap was real and asymmetric: R4&#39;s conflict side contained its selectors, `SEL`&#39;s closing `};` *and* the whole `const NAME = {…}`, while the `module.exports` edge auto-merged — so a plausible resolution exports an undefined `NAME`, a `require()`-time `ReferenceError` that kills every Playwright spec while vitest stays green. Hoisting `NAME` above `SEL` makes that structurally impossible.

**Not fixed, deliberately:** `MoversPanel`&#39;s non-focusable clickable rows and the unrendered `capped` contention flag. Both are pre-existing and unchanged by this PR; folding them in would widen an already-large diff without a regression to justify it.

## ds primitives

**No API gaps — nothing added or mutated.** Everything composed from existing primitives. One constraint recorded for R3/R5: `Tooltip` requires a focusable child, so it can&#39;t annotate read-only table badges without adding tab stops. I used self-describing labels + backend-supplied notes instead.

## E2E

SEL registry gains an additive R4 block (waivers/trade/trades/angle/draft hooks) plus a `NAME` export for accessible-name patterns; `waivers-smoke` reconciled to the new copy through it, and its first assertion retightened (above). 89 specs parse.

**The E2E suite did not run for real, and the reason is environmental, not a property of this branch:** this container&#39;s backend has no data — its last scrape failed. `/api/health` returns 503, `/api/data` returns 401, and `/api/status` reports `last_success_at: null`. The journey specs are deliberately data-driven, so a run here would either skip or time out on missing data rather than exercise the changes. Chromium is available at `/opt/pw-browsers`; what&#39;s missing is a seeded stack. **These specs need a real run against data before the integration window** — flagging it rather than reporting a green I didn&#39;t get.

## Verification

- **vitest: 56 files / 1179 tests green** (was 53/1159 when this body was first written; +14 from the review follow-ups above, re-measured at `1563e527`)
- **`npm run build` green; all 14 bundle budgets pass**
- No page regressed; five improved

---


jasonleetucker-code:

Reviewed commit `79bde6f64af85625c4a69d7b2e96a4eca215e672` (R4: fix /draft horizontal overflow at 390px) against `origin/main` @ `9ccdecea`. I did not write this PR. I tested the preservation claim rather than accepting it.

**Verdict:** the preservation work is solid — the sticky tray and the FAAB call sites really are equivalent, and I couldn't find any value math that moved client-side. But two of the PR's own claims don't survive checking: `/waivers` ships a **new feature**, not a rebuild, and the `aria-sort` claim is false for `/draft`'s primary table.

---

### Verified clean (tested, not assumed)

**`/trade` sticky-tray verdict bar — genuinely equivalent.** The author flagged this as dropped-and-restored, so I diffed the inputs rather than the markup. The entire gap computation block is byte-identical to main (`app/trade/page.jsx:653-659` vs main's `895-901`): same `pwTotalA/pwTotalB` off `sideTotals[].adjusted`, same `pwGap`, same `pctGap` rounding. Same `colorFromGap` / `verdictFromGap` / `verdictBarPosition` helpers, same `sides.length === 2` gate. The marker moved to a custom property, and `left: calc(var(--marker-pos, 50%) - 6px)` (`trade.module.css:227`) reproduces the old inline `calc(X% - 6px)` exactly, including the neutral state (`--neutral-signal` for old `--cyan`) and the `transition: left`. `aria-hidden` on the decorative track is a correct addition — the verdict text carries the meaning.

**`/waivers` FAAB contention matches the recommender's actual response shape.** I checked the rendering against `server.py:4767-4774` / `4802-4809` and `src/trade/faab_contention.py:335-337` rather than against the component's docstring. Every field the table reads exists on the wire: `ownerId`, `teamName`, `expBid`, `faabRemaining`, `needLevel`, `aggression`, `lowSample`, `intelLevel`, `balanceUnknown`. The `needLevel` domain is `need|surplus|neutral` (`faab_contention.py:246-272`) and `NEED_TONE`/`NEED_LABEL` cover all three with no fallthrough. The `skipped: true` branch is handled and reads `notes[0]`, which the server always populates with a skip reason. `capped` is on the wire but not rendered — a real omission (a rival pinned at their FAAB ceiling can't go higher, which is useful) but not a correctness bug.

**`computeFaabHint` preserved.** Same helper, same `{ leagueBudget: 100, topValueInPool: topAddValue }`, same `m.addValue || m.add?.rankDerivedValue || m.add?.values?.full || 0` fallback chain (`app/waivers/page.jsx:210-220`).

**`/angle` preserved,** including the subtlety the comment calls out — main already drove result rendering off `result.mode` rather than the UI selection (main `angle/page.jsx:297-299`), so that's carried over, not newly fixed. Request payload identical.

**Dropped imports are genuinely dead.** `/trade` lost `RANKING_SOURCES`, `SOURCE_VENDOR_LABELS`, `vendorForSource` — all three were imported but never referenced on main. `posBadgeClass` dropped from `/trade` and `/waivers` in favour of ds `Badge`. No loss.

**`/draft` modals — the focus trap is real.** `components/ds/Dialog.jsx:98-144` does what it claims: `role="dialog"` + `aria-modal` + `aria-labelledby` on the panel (main had `role`/`aria-modal` on the *backdrop* div with no accessible name), Tab wrapping in both directions, Escape gated on a document-order topmost-overlay check, reference-counted scroll lock, and `restoreRef.current?.focus?.()` on cleanup. Strictly better than what it replaced.

**Mobile scroll containers** (`globals.css:7107-7117`) use the right pattern — `overflow-x: auto` on the list plus `min-width: max-content` on the rows. I checked for absolutely-positioned descendants that the implied `overflow-y: auto` would clip; there are none.

**Tests + build:** `npx vitest run` → **54 files / 1165 tests passing**. `npm run build` clean, **all bundle budgets pass**, including the newly-pinned `/waivers` at **30.3/36 KB**. R4 also measurably shrinks `/trade` (73.0 → 67.1 KB), `/draft` (124.9 → 122.4), `/angle` (14.5 → 12.9) and `/trades` (15.6 → 14.2).

---

### P2 — the `aria-sort` claim is false for `/draft`'s main board

`app/draft/page.jsx:1118-1134`, headers wired at `1188-1218`

```
$ grep -c "aria-sort" frontend/app/draft/page.jsx
0
```

The draft board's nine sortable columns (`#`, `Player`, `PreDraft`, `Fair`, `Enforce`, `Win at`, `Max Bid`, `Final`) go through a local `th()` helper that renders a bare `<th>` with `cursor: pointer` and an `onClick` — no `<button>`, no `tabIndex`, no `role`, no `aria-sort`. Sort direction is conveyed only by a `▲`/`▼` text glyph appended to the label.

Failure scenario: a keyboard-only user on `/draft` cannot sort the board at all — the `<th>` never receives focus, so Tab skips it and Enter/Space do nothing. A screen-reader user gets a plain column header with no sort state and a stray arrow character. This is the largest and most interactive table on the page.

To be fair: this helper is **byte-identical to main**, so R4 preserved it faithfully and introduced nothing. The issue is the PR's blanket claim of "`aria-sort` on all sortable headers" — true for every surface that moved onto `DataTable`, not true here. Either move the board onto `DataTable` or scope the claim in the PR body so the next a11y audit doesn't treat `/draft` as done.

### P2 — the FAAB v2 contention section is a new feature, not a rebuild

`components/waivers/FaabRecommendation.jsx:98-230`, rendered at `516`

The backend has returned `contention` for some time (36 references in `origin/main:server.py`), but **no main-branch frontend file references `contention`, `perOpponent`, `topRival`, or `clearing`** — I grepped all of `frontend/`. R4 changes zero Python files. So this is ~130 lines of brand-new user-facing surface — a clearing price, a projected top rival, and a full per-opponent bid table — arriving inside a PR framed as "every behavior preserved verbatim."

Two consequences worth weighing before merge:

1. **Review attention.** A reviewer told everything is preserved will skim this section. It's the one part of the diff with no pre-R4 behavior to diff against.
2. **Zero test coverage.** No test in `frontend/__tests__/` references `FaabRecommendation` or `contention`, and the vitest counts are identical between this branch and #551 (1165 both) — neither PR adds a single unit test. For a section that renders bid guidance a manager will spend real FAAB against, at minimum the `skipped` branch and the `balanceUnknown`/`lowSample` flag rendering deserve a test.

The code itself looks right (see verification above). This is a scoping and coverage concern, not a correctness one.

### P2 — `tests/e2e/helpers/journey.js` will conflict with #551, and the obvious resolution breaks the whole e2e suite

Both PRs insert into `SEL` at the same anchor (immediately after `navSearchButton`). I ran the actual three-way merge:

```
$ git merge-file -L R3 -L base -L R4 merged.js base.js theirs-r4.js
MERGE EXIT=1   → conflict at lines 42-102
```

The hazard is asymmetric and specific to this PR. Your side of the conflict block contains the selectors, the closing `};` of `SEL`, **and** the whole `const NAME = {...}` declaration. But your edit to `module.exports` (adding `NAME,`) is a separate hunk that **auto-merges cleanly** — I confirmed `NAME,` lands at line 237 of the merged file no matter how the conflict is resolved.

So if the conflict is resolved by keeping #551's block and dropping what looks like a stray extra brace, the result exports `NAME` without defining it → `ReferenceError: NAME is not defined` at `require()` time → **every Playwright spec importing the journey helper fails**, not just R4's. Vitest won't catch it; only the regression job will.

No key collisions between the two selector sets, so the fix is mechanical — keep both blocks plus the `NAME` object. Just needs to be called out for whoever merges second.

### P3 — `waivers-smoke` first assertion is now vacuous

`tests/e2e/specs/waivers-smoke.spec.js:24-26`

```js
await expect(authedPage.locator("body")).toContainText(/Waivers/i, ...)
```

`"Waivers"` is the nav label for `/waivers` (`lib/nav-model.js:59`), so it's present in `body` on every authenticated page. If the `/waivers` route threw and rendered nothing but chrome, this assertion still passes — where the old `/Waiver Add\/Drop/i` tested real page content.

The second assertion (`Best add\/drop moves|Addable|Droppable|Pick your team`) is data-driven and does provide cover, which keeps this P3. Anchoring the first one to the page's `<h1>` rather than `body` would restore the intent cheaply.

### P3 — focus restore lands on `document.body` when the opener is gone

`components/ds/Dialog.jsx:142`

`restoreRef.current?.focus?.()` calls `focus()` on the stored element unconditionally. Calling it on a detached node is a silent no-op, and focus falls back to `document.body`.

This is reachable on `/draft`: open the mark-drafted modal from a player row's Draft button, submit, and the player leaves the available pool — so the opener unmounts before `setModalPlayer(null)` triggers the restore. A keyboard user lands back at the top of the document and has to Tab all the way to where they were. Guarding with `isConnected` and falling back to a stable container would close it. Pre-existing behaviour was worse (no restore at all), so this is polish rather than a regression.

---

PR #554, opened 2026-07-26
E2E: one-command runnable suite + verified pass on the stacked redesign

jasonleetucker-code:
## The headline

The suite now has a **verified end-to-end pass against the full stacked redesign** — `main` + R2 (#549) + R3 (#551) + R4 (#552) merged locally:

```
149 passed · 0 failed · 29 skipped · exit 0 · 9.3 min · zero server deaths
```

All nine rebuilt surfaces are covered by that run. Two agents in a row could not get a real run; this PR fixes why, and the fixes are infrastructure, not assertion-loosening.

## One command, fresh container

```bash
npm ci && npm --prefix frontend ci && npm run e2e
```

No env setup, no running stack, offline, no credentials. It boots the stack, waits for readiness, runs the specs, and tears down. **Verified cold** (stack down, every `E2E_*` unset) — not asserted from reading the config.

## Why it was unrunnable

Both reports were misdiagnosed downstream, because the obvious signals lie.

**R4's "backend has no data."** The Playwright `webServer` booted `python server.py` with only `UPTIME_CHECK_ENABLED`, so on a fresh container the backend **died at import** — `server.py` raises without `JASON_LOGIN_PASSWORD`. Everything observed afterwards is *also* what a correctly-configured offline stack looks like, which is exactly why it read as a data problem:

| Signal | What it actually means |
|---|---|
| `/api/health` → 503 | Normal offline. The startup scrape can't reach ranking sites, so the backend is permanently "degraded". |
| `/api/data` → 401 | Correct — auth-gated. Use a session. |
| `last_success_at: null` | Expected. The suite runs on the committed snapshot, never a live scrape. |

The config now passes the backend everything it needs: `ALLOW_DEFAULT_LOGIN_DEV`, `E2E_TEST_MODE` + a shared secret (without which **every signed-in spec skips and the run is green while testing nothing**), `RATE_LIMIT_BYPASS_IPS` (the 60/min + 1000/hour public limiter drains mid-run and its 429s surface as bogus auth/render failures), and an empty browsers dir so the startup scrape fails fast instead of reaching live ranking sites.

**R3's "can't install `@playwright/test` without touching the lockfile."** Both lockfiles are committed, so **`npm ci` is exact and never rewrites them** — verified byte-identical, 613 ms. No lockfile edit is needed by anyone. Preflight now checks runner and frontend deps separately and prints that command.

## Verify, don't assume

New `tests/e2e/global-setup.js` runs before any spec and checks the things that actually matter — snapshot loaded (`/api/status has_data`), test sessions genuinely unlocked, an **authenticated** `/api/data` populated, frontend serving — each failing with the fix, not a wall of timeouts. It also gates on the frontend explicitly: Playwright was observed starting tests while `next build` was still running, producing a screenful of `ERR_CONNECTION_REFUSED` that looks like a product catastrophe.

It also fails fast when `E2E_BASE_URL` points at a dead port. That gap cost me a 10-minute run that reported 149 failures against nothing — the guard I'd written skipped itself in exactly that case.

Preflight seeding is no longer "is `data/` non-empty?": it validates the snapshot the backend will actually pick (newest by name), reseeds when it's empty or truncated, and quarantines an unusable newer file that would otherwise shadow a good one — the state a failed scrape leaves behind.

Chromium is auto-detected from `PLAYWRIGHT_BROWSERS_PATH` when the pinned revision isn't installed, so sandboxes where `playwright install` can't run need no flag.

## Product findings the run surfaced

- **Page navigations must bypass the backend's page proxy.** Through it, `/` renders the **anonymous** shell for a signed-in session — `/api/auth/status` says `authenticated:true` while the page shows "Sign In". Specs asserting signed-in chrome were testing logged-out pages. All page navigations now go through `pageUrl()` (production's nginx topology). `/waivers` and `/news` aren't in the proxy's route list at all, and `/league` SSR exceeds its 5 s timeout.
- **`critical-smoke`'s anonymous `/` check has been red on `main`** since R1 renamed the wordmark to "Brisket" and moved anonymous `/` to the sign-in surface. Nightly-only E2E hid it. Now asserts the brand, which is present in every route and auth state — this check exists to prove the route renders, not to pin marketing copy.
- **The methodology toggle rename** ("How this works" → "Methodology") silently no-opped the chart specs' opener, so three chart specs failed as "chart didn't render" rather than "the button moved". Matches both labels now.
- **`useAuth` caps `/api/auth/status` at 5 s and resolves to *unauthenticated* with no retry** — a slow backend can strand a valid session on the anonymous shell for the page's whole life. This is a real product fragility, reported rather than papered over with a retry.

## Next dying under load — reported, not masked

It reproduces **intermittently: 1 of 4 managed full runs**. Characterisation:

- **Not the suite's concurrency.** Identical default worker count across the clean runs and the killed one; `fullyParallel: false` already.
- **Not memory.** Bare `[WebServer] Killed` — SIGKILL from outside, no Node stack trace, no heap-OOM message. RSS ~180 MB with 8.7 GB free; no OOM in `dmesg`.
- **Not `/league`-specific.** Hammering all ten `/league` tabs on a healthy stack: all 200 in ~10 ms.
- **Correlates with process ownership.** Every manually-backgrounded stack (`nohup`/`setsid`) was reaped; the Playwright-managed path — which owns the children and bounds their lifetime — was clean in 3 of 4 runs including the gate run.

**Recommendation: no serialization change and no retries.** Use the managed path, which is now the documented command. Cost when it does hit: ~15 cascading `ERR_CONNECTION_REFUSED` that read as a product catastrophe. Worth a follow-up that aborts the run when the frontend disappears mid-flight, so the signal stays honest.

## Cross-PR hazard (#551 ↔ #552) — pre-resolved on both branches

R3 and R4 both append to the `SEL` registry at the same anchor, and the conflict was a **trap**: R4's side held its selectors, `SEL`'s closing brace, *and* the whole `const NAME` declaration, while the `module.exports` hunk adding `NAME,` auto-merged separately. Keeping one block and dropping the duplicate-looking brace exports `NAME` without defining it → **ReferenceError at `require()`, killing every spec, not just the losing workstream's — while vitest stays green.**

Fixed structurally rather than resolved once: `const NAME` hoisted above `const SEL` (sitting against the closing brace is what let it get swallowed), R4's block first inside `SEL`, and the **identical registry shipped on both branches** so the merge is a no-op in either direction. Verified by trial merge in both orders plus loading the module (34 `SEL` keys, `NAME` defined and exported). Mechanism and the hoist fix were characterised by the design custodian; landed from the registry owner so both branches carry identical bytes.

## Scope note

This PR is deliberately **not** a re-do of R2's selector reconciliation (#549 owns that). It carries infrastructure, the topology fix, and the residue #549 doesn't cover. Against `main` alone it still shows 3 failures that #549 fixes (`.nav-search-btn` → `.shell-search-btn`, the pre-R1 mobile-nav locator) plus the `useAuth` fragility above; against the stack **with** #549 the suite is green.

Selectors that changed in the rebuild are matched **both ways** (`.table-wrap` and `.ds-table-wrap`) so these specs span the rebuild instead of needing a flag-day edit timed to a merge.

## Verification

- **Stacked (main + #549 + #551 + #552 + this): 149 passed / 0 failed / 29 skipped, exit 0.** The 29 skips are deliberate project gating (fixed-viewport visual suites on mobile, desktop journeys on mobile) and uncommitted pixel baselines.
- Cold one-command run verified twice from a fully torn-down stack.
- `npm ci` lockfile-identical; preflight recovery verified by simulating a failed scrape (truncated newer snapshot → quarantined and reseeded).

---


jasonleetucker-code:
**Product defects filed separately: #555.**

The five defects this PR's runs surfaced are in `server.py` / frontend-auth territory and are **not** fixed here — #555 carries them with the evidence:

1. The backend page proxy serves the **anonymous shell to signed-in sessions** (`/api/auth/status` says `authenticated:true` while the proxied page shows "Sign In"; the same cookie against Next directly renders the authed dashboard).
2. `/waivers` and `/news` are missing from the proxy's hand-maintained page-route list (404 through `:8000`, 200 through `:3000`).
3. `/league` SSR exceeds the proxy's 5 s timeout (503 cold; 7-19 s direct).
4. `useAuth` caps `/api/auth/status` at 5 s and resolves to *unauthenticated* with **no retry** — a slow moment strands a valid session on the logged-out shell for the page's whole life.
5. `critical-smoke`'s anonymous `/` assertion has been red on `main` since R1 renamed the wordmark.

This PR works around 1-3 in test code only, by routing page navigations to the Next origin — which is what production does via nginx. #4 is reported rather than retried around, deliberately: a retry in the test would hide a user-facing defect. #5 is the one genuinely test-side item and is fixed here (`84b28a51`).

Note for anyone reading the nightly: **until this PR lands, `main`, #551 and #552 all still carry the failing `/Risk It/i` assertion**, so main's nightly E2E is expected-red for that reason — not a new regression.

---

jasonleetucker-code:
Closing as **contained**, not abandoned — every commit on this branch is on `main`.

`claude/e2e-r1-reconcile` is a strict ancestor of `claude/e2e-abort-guard` (`git merge-base --is-ancestor` confirms it, and `git log pr554..pr559` is a single commit going the other way). So merging #559 as `e55f791b` landed this PR's work in full, and running a second ~20-minute gate for it would have added nothing but queue time while the runner pool was saturated.

Verified on `main` after the squash merge rather than inferred from the ancestry:

| file | on `main` |
|---|---|
| `tests/e2e/global-setup.js` | present |
| `tests/e2e/preflight.py` | present |
| `tests/e2e/README.md` | present |
| `tests/e2e/playwright.config.js` | present |

`git diff origin/main pr554` now shows only `main` being *ahead* — it carries `stack-death-reporter.js` and the `e2e.yml` duplicate-key fix, which this branch predates. Nothing here is unmerged.

One thing that landed alongside and is worth recording here, since this PR is where the suite was made runnable: `e2e.yml` declared `E2E_PAGE_ORIGIN` twice. GitHub Actions rejects a workflow with a duplicate mapping key, while YAML parsers silently keep the last one — so **all eight runs of that file since it was introduced were startup failures with zero jobs**, including the scheduled run. Nothing ever reported red because nothing ever ran. Fixed in #559; a duplicate-key scan across all fourteen workflows finds no others.

---

PR #557, opened 2026-07-26
docs: session audit handoff for independent technical review

jasonleetucker-code:
Adds `docs/CLAUDE_SESSION_AUDIT_HANDOFF.md` (1,651 lines) — a comprehensive handoff intended to be given to an independent AI or senior engineer for a technical and strategic audit. **Docs only. No code, no config, no behavior change.**

Grounded in the repository state at `c534280d`.

## What it contains

All 15 requested sections: product objective · work chronology · current architecture with file paths · every implemented formula with constants and `file:line` references · roster-analysis system · trade-recommendation system · competitor-parity provenance · the agent/workflow fleet · file-level inventory of completed vs. in-flight work · prioritized technical debt · contradictions with their controlling interpretation · adversarial self-audit · a 7-day execution order · reviewer questions · source appendix.

## Two limitations stated in the document rather than papered over

1. **Section 2 is reconstructed archaeology, not recall.** This session carries no prior conversation — the handoff request was the first message. The chronology is rebuilt from git history, merged/open PR bodies, ADRs and the roadmap docs, with provenance and confidence tabulated per section. No dialogue was invented.
2. **No tests were executed here.** `pytest` is absent from the session container (the SessionStart health hook reports the same). Every test figure quoted in the document is attributed to the PR body it came from and explicitly marked unverified.

## Findings the reviewer is pointed at first

Ranked in the document; the top of the list:

- `config/leagues/registry.json` `rosterSettings` are stale on 8 fields on `main`, and `src/trade/suggestions.py::DEFAULT_STARTER_NEEDS` is a hardcoded mirror of the same stale lineup — so trade needs/surpluses, FAAB need factors and ROS lineups are all computed against a lineup the league does not use. Fix exists on #550, unmerged.
- Sleeper evaluates slot eligibility against `fantasy_positions`, not `position`; hybrid IDPs are locked out of legal slots in production. Also #550.
- Two contradictory single-source penalties (`×0.30` in the blend vs `×0.88` in the finder) and two contradictory fairness scales (`769` vs `350/900/1800`) are live simultaneously.
- The IDP market-corridor clamp bounds IDP values to ±15% of IDPTC, which constrains the "21-source consensus" claim on IDP and is not disclosed in the UI.
- The measured TE-premium (≈1.12, symmetric endpoints) contradicts the applied constant (1.15).

The self-audit section is deliberately adversarial about the document's own reliability, including the three places where its author could not verify a claim and said so.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UvqwYWWM9689T21dyGDRto)_

### DIFF
--- a/docs/CLAUDE_SESSION_AUDIT_HANDOFF.md
+++ b/docs/CLAUDE_SESSION_AUDIT_HANDOFF.md
@@ -1,0 +1,1651 @@
+# Claude Session Audit Handoff — Risk It To Get The Brisket
+
+**Prepared:** 2026-07-26
+**Repository:** `jasonleetucker-code/riskittogetthebrisket`
+**Branch this document was written on:** `claude/session-audit-handoff-tvxfc1`
+**Repo HEAD at time of writing:** `c534280d` (identical to `origin/main`)
+**Intended audience:** an independent AI or senior engineer performing a technical + strategic audit.
+
+---
+
+## 0. Provenance warning — READ THIS FIRST
+
+This document must not be read as a verbatim transcript reconstruction, because it cannot be one.
+
+**The Claude Code session in which this document was produced began with the handoff request itself.** There is no prior visible conversation in this session's context: no earlier user messages, no earlier assistant turns, no earlier tool calls. The container was started fresh, the repository was cloned fresh, and the first user message was the request to write this file.
+
+Therefore:
+
+| Section | Provenance | Confidence |
+|---|---|---|
+| 1 (objective) | Inferred from `CLAUDE.md`, `docs/ROADMAP-competitor-parity.md`, `docs/league-intelligence/MASTER_PLAN.md`, PR bodies | Medium-high — these documents state user intent explicitly and quote user decisions |
+| 2 (chronology) | **Reconstructed from repository artifacts only** — git log, merged PR titles/bodies, open PR bodies, ADRs, roadmap "User decisions (confirmed)" blocks | Medium — sequence and content are real; the *conversational* framing (what was asked vs. what was proposed) is inferred |
+| 3, 4, 5, 6 (architecture, formulas, roster, trade) | **Read directly from source code at `c534280d`** with file:line references | High — verifiable by inspection |
+| 7 (external inspiration) | `docs/ROADMAP-competitor-parity.md` + registry comments | High |
+| 8 (agents/workflows) | `.github/workflows/`, `.agents/skills/`, `docs/ORCHESTRATION.md`, open PRs | High for files; medium for "is it working" |
+| 9 (work completed) | Git diff of this branch vs `origin/main` | High — and the answer is uncomfortable; see §9 |
+| 10–14 | Analysis by me over the above | Explicitly labelled as judgement |
+
+**Where I do not know something, this document says so.** I have not invented dialogue, commit history, test results, or deployment outcomes. Every claim about repository state below was verified by reading the repository during the production of this document; every claim about *intent* is sourced to a document that records it.
+
+An auditor should treat §2 as "how the system got here, as evidenced by what it left behind" — not "what Jason said and when."
+
+---
+
+## 1. Original Objective
+
+### 1.1 The product
+
+**Risk It To Get The Brisket** is a private dynasty fantasy football valuation and decision-support platform, built for a specific 12-team Superflex + TE-Premium + IDP dynasty league on Sleeper. It ingests ~20 external ranking/valuation sources, normalizes them onto one canonical 0–9999 value scale, and serves a web application for rankings, trade analysis, waiver/FAAB bidding, draft preparation, league intelligence, and news.
+
+### 1.2 The problem it solves
+
+Dynasty managers evaluating a trade today have to consult several disagreeing markets (KeepTradeCut, DynastyDaddy, FantasyCalc, DLF, IDP Trade Calculator, …), each with its own scale, scoring assumptions, and blind spots — and none of which know anything about *their* league's scoring, roster construction, or league-mates. The stated ambition is to close both gaps at once:
+
+1. **Consensus instead of one market.** Blend many sources into a single defensible value with explicit confidence, disagreement, and provenance stamps — so a value is auditable, not a black box.
+2. **League-specific instead of generic.** Reprice assets against this league's exact 141 scoring keys, 21-slot best-ball lineup, superflex + 2-TE demand, and IDP requirements — where the generic market is systematically wrong.
+3. **Edge instead of information.** Turn the value board into action: trades that look fair to the counterparty on KTC but win on our board, FAAB bids that clear the actual expected top rival, waiver claims priced against replacement level, and intelligence on how league-mates behave across all their other Sleeper leagues.
+
+`docs/ROADMAP-competitor-parity.md` states the driving requirement plainly:
+
+> "The user wants the site to give them every possible edge in their league (trades, waivers/FAAB, player evaluation, league-mate tendencies)."
+
+### 1.3 Target users
+
+Primarily a **single power user (the owner) and his league-mates by invitation.** Evidence:
+
+- Auth is an allowlist of usernames (`PRIVATE_APP_ALLOWED_USERNAMES`, default `jasonleetucker`) plus a guest-pass system (`src/api/guest_passes.py`).
+- The roadmap explicitly declares "Public/commercial polish (site is private personal use)" as a **non-goal**.
+- A separate public, unauthenticated surface exists at `/league` (`src/public_league/`, 24+ sub-sections) that is league-facing content — standings, matchup recaps, awards, rivalries, player journeys — i.e. a league "hub" for the other 11 managers.
+
+So there are effectively **two products in one repo**: a private analytics terminal, and a public league-history/engagement site.
+
+### 1.4 Expected final experience
+
+From `docs/ORCHESTRATION.md`:
+
+> "Target: **comprehensively functional, integrated, polished product in ~1 week** (by ~2026-08-02). Optimize for the final integrated system, not constant main-branch stability."
+
+The intended end state is a single coherent app where every surface (rankings, player profile, trade, draft, waivers, news, intel, league hub) reads from one canonical contract, renders on one design system, and can explain every number it shows. Notably, the user explicitly **de-prioritized continuous production stability in favour of end-state quality** — a decision that is echoed in the instruction for this very document ("Do not prioritize keeping the production site continuously functional unless it is necessary to avoid data loss or blocked development").
+
+---
+
+## 2. Chronological Session History (reconstructed from artifacts)
+
+**Reminder: reconstructed. See §0.** What follows is the sequence of work as evidenced by the repository, with the *decisions* quoted from documents that record them. Where I infer the conversational shape, I say "inferred".
+
+### Stage A — Pre-existing baseline (before the reconstructable window)
+
+The repository already contained, before any of the work below: the canonical valuation pipeline (`_compute_unified_rankings`), Hill curves, Hampel filtering, ~18 ranking sources, Sleeper overlay, trade suggestions + arbitrage finder, ROS engine, public league hub, news backend, and a large automation fleet. This is important context: **almost nothing in this project was built from scratch in the reconstructable window.** The dominant activity has been *extension and correction of an already-large system.*
+
+Evidence: `docs/` contains dated status reports going back to 2026-03 (`scoring_adjustment_audit_2026-03-09.md`, `deploy-failure-20260323-analysis.md`, `rankings_identity_fix_2026-04-14.md`), and the contract version is pinned at `2026-03-10.v2`.
+
+### Stage B — 2026-07-25: the competitor-parity mandate
+
+**What was requested (as recorded):** a 37-item roadmap benchmarking the site against two competitors — **FantasyNavigator.com** and **PlayForKeepsDynasty.com (PFK)** — spanning player search/filtering, competitor audits, self-audit, two new ranking sources, TEP verification, FAAB optimization, PFK-style Sleeper intelligence, and a real news system.
+
+**How it was interpreted:** not as a single change but as a phased multi-week program, with audit deliverables produced *alongside* implementation rather than gating it.
+
+**Approach proposed:** `docs/ROADMAP-competitor-parity.md`, "approved 2026-07-25", sequencing Phase 0 → 2+2b → 3 → 4 → 5 → 6 → 7.
+
+**Alternatives considered and rejected (recorded in that document):**
+
+| Alternative | Verdict | Recorded reason |
+|---|---|---|
+| Audits (Phase 1) as a blocking gate before implementation | **Rejected** | User decision: "Audit reports (Phase 1) are produced incrementally alongside, not as a blocking gate." |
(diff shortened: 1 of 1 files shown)


jasonleetucker-code:
## Superseded by #561 — disposition of every finding

Closing this as superseded, not as duplicate work. This PR and the one on `main` (#553) target the same path and cannot both merge, but they were written from genuinely different positions and this one found things the merged version could not. The merged document says outright that it did not read `data_contract.py`, `player_valuation.py`, the trade engines or the ROS/FAAB formulas, and that reproducing them from memory would produce something that *looks* authoritative and misleads the auditor. **This session went and read them.** That is the material that has been carried over.

The salvage is in **#561**, as a new §16 attributed throughout to this audit.

**Method:** every unique finding was re-verified against current `main` (`253568bc`) before porting. This branch was based on `c534280d`; `main` has moved several merges since, most importantly **#556** (trade-finder per-market gate) and **#551**. Nothing was carried on this document's word. Verification was by reading source and `git show` against `pull/550/head` — **no tests were run**, so where this PR quoted a test figure from another agent's PR body, that provenance and the "unverified" marking are preserved.

**Totals:** ~60 findings ported · 13 dropped or reframed · 14 recorded as open questions · 3 of your open questions resolved · 2 errors in the merged document corrected on your evidence.

---

## Carried over

**§4 formula detail — the largest contribution.** Nearly all of it, with every constant re-confirmed at `file:line`. The merged §4D table now points at §16 per row instead of saying "unread."

- Value scale, rank→percentile, all four Hill masters (incl. ROOKIE `routed: false`), the two parallel rank-form/percentile-form parameterizations, the worked OFFENSE example
- Value-direct voting + **the full removal provenance with Hampel drop rates** (dynastyDaddySf 61%, yahooBoone 47%, fantasyProsFitzmaurice 19%, fantasyCalc 55–58%, otcffbSf 56%→86%) + the DraftSharks negative-scale carve-out + the import-time invariant. All exact.
- Hampel K/floor/min_n and the recorded reason the floor is 1000 not 500
- Count-aware blend table, read line by line
- **α-shrinkage and its tuning history** — that the joint backtest optimum was α = 0, that it was the degenerate "use IDPTC alone" solution, that it was rejected on declared-objective grounds at a recorded ~2× stability cost. This is the most decision-relevant thing you recovered and the merged document did not have it. It is now §16.9 #4 as an open question, because it is a values question wearing a metric.
- λ retirement + `madPenaltyApplied` dead-field debt
- TEP constants **plus** `_TEP_DERIVATION_SLOPE = 0.30` and the `bonus_rec_te` auto-derivation, which the merged doc missed entirely
- Pick tethering + the discount table + the three-step `current_rookie_draft_year()` resolution — including your credit for the offset schema never going stale, and your catch that 0.82/0.66/0.53 is near-geometric at 0.82 while `fallbackBase` is 0.80
- Confidence buckets, trimmed spread, depth allowance — **with the caution-saturation numbers** (72% of top-200 flagged; 143→~40 and 82→~15 post-fix; median spread 0.068 top-100 vs 0.30 at 201–400). Excellent find; flags rising *with* coverage is a clean example of a metric degrading as the product improved.
- Injury adjustment in full, and all three of your criticisms (unconditional 5% cap, enum severity, the Aug 31/Sep 1 discontinuity)
- Tiering constants + the "priors" status
- Two-way boost verified as a one-directional `max()`, optimistic by construction
- FAAB v2 in full. **Your read that this module is exemplary holds up and is carried as such** — `balanceUnknown` rivals excluded from `topRival`, contention skipped below half-coverage, never guessing which team is the user's, and the irreducible selection bias stated in the returned `notes`. It is the missing-data discipline the rest of the codebase should be measured against.
- `trend_score` — nested windows **verified**, 30-hour event gets effective weight 6, weights unsourced
- `power_v2` weights, `direction.py` bands and veteran-age thresholds
- **`team_ros_strength_percentile` is a silently inert input** — verified exactly: assigned at line 78, used only at line 120 inside the summary f-string, never touches the label. Precisely right.

**§6 trade internals.** The arbitrage score and `rank_score` compositions, both verified. Your `f_positive_bonus = 10` criticism stands and is escalated — a flat step exceeding both graded terms makes the finder substantially a binary "is board_delta positive" sort. `rank_score`'s unbounded base likewise: **the suggestion feed is primarily a "biggest trades" feed wearing a quality score.** The `__dict__.get` structural smell on `edge`/`opponent_fit` is verified and is exactly the "fix that reads correctly but cannot take effect" class — sitting latent in live code. The two fairness vocabularies (769 vs 350/900/1800) are verified, and the code comment already admits 769's origin is "undocumented legacy tuning." No 2-for-2 generator: confirmed. No roster-legality checking: confirmed. No acceptance-probability model on `main`: confirmed by an independent route.

**§10 debt and hygiene.** `Jenkinsfile`, `codex_loop.py`, the 3.4 MB PDF, 22 unindexed `docs/status/` reports, `server.py` at 10,707 lines, `data_contract.py` at 9,009 — all verified. Your line on the concurrency strategy is carried close to verbatim: *"not an architecture, a queue discipline; it works only while agents cooperate."*

**The race condition the merged §8 does not cover:** three independent writers push to `main` — `scheduled-refresh.yml` plus the two VPS shell loops. Nothing enforces serialization across machines. Concrete, cheap, absent from the merged Day-3 plan.

**§7.1 legal/ToS.** Carried nearly whole — the private-use basis, the politeness cadence, PFK's embedded publishable Supabase key and the "readable ≠ licensed" question, and your framing of Sharp Tracker as surveillance of named league-mates with "keep it inside this private app" being a **policy, not a control**. Plus `PRIVATE_APP_ALLOWED_USERNAMES` failing open on the name. The merged §7 flags Draft Sharks and IDP Show only.

**#550 branch-only material** — exact scorer + ADR-006, the greedy-vs-laminar optimizer verdict, `_positional_coverage` as a constant with the PR's own honest limitation, LI-5 scarcity, LI-4's enforced no-op and the batch-scoped monotonicity guardrail. And **`fantasy_positions` vs `position`**, which you asked to be verified rather than assumed: confirmed as branch-only, and it is the strongest evidence in either document because it was found empirically — the reconstruction scored *below* the host on 5/10 weeks before anyone knew why.

---

## Corrections you made to the merged document

Both fixed inline in #561, with pointers, because burying them in an appendix would have left false statements standing.

**1. "No database."** The merged doc said flatly *"No database. Everything is JSON snapshots… treat database migrations as N/A."* You were right: three SQLite stores, two on disk right now, with schema application and a migration module. `user_kv.sqlite` is the only genuinely irreplaceable state in the system. The merged doc's underlying concern survives — no analytics store, nothing champion-challenger could query — but its phrasing was wrong and its advice to the auditor was wrong.

**2. `grant-ssh-access.yml`.** The merged doc listed it as a **Critical** standing production-shell risk "not yet deleted," with "delete it" as Day-1 item 2. You were right that it was deleted — `c534280d`, 33 minutes after creation, **before that document was written**. The merged document listed as an open Critical risk something the repo had already closed. That is the same class of error its own §12 catalogues, committed inside the document cataloguing it. Now replaced with the question that *is* open: did it ever execute.

Worth noting the symmetry: your "I do not know the reason. An auditor should ask" is **answered** by the merged §2 Stage 9, which you had no way to see — a one-shot `workflow_dispatch` to restore SSH after the VPS rebuild wiped all four of the user's keys, created via the web UI because a safety classifier blocked Claude from committing it twice. Neither document had the whole picture; §16.7 now does.

---

## Dropped or reframed — 13 items, with reasons

All recorded in §16.7 rather than deleted, so the record shows what happened.

**Reframed (3) — real observations whose framing leads somewhere harmful:**

**1. "Two contradictory single-source penalties… at least one is wrong."** The observation is real, the conclusion is not. The two are **not stacked**: `suggestions.py` reads `rankDerivedValue`, which *has* had the 0.30 retention applied (so it correctly adds nothing); `finder.py` reads `_finalAdjusted`, which `data_contract.py` deep-copies verbatim from the raw scrape and which the retention never touches. So 0.88 is the **only** haircut on the finder's path. The obvious remediation — delete it, or unify on 0.30 — would leave single-source assets **undiscounted**, introducing a live distortion via a fix. That exact proposal was made and retracted earlier the same day; it is item 4 in the merged §12 correction table, and the constant is now pinned by a regression test. **What survives is real and is carried:** the 0.88 is unanchored, since its "matches the frontend" justification no longer holds. F-6 is the fix, and it deletes the constant rather than retuning it.

**2. "The ±15% IDP corridor clamp means IDP arbitrage is bounded by construction."** First half verified exactly, including your worked Vikings-LB example. Second half **false for the engine you attribute it to**: the clamp operates on `rankDerivedValue`, and `finder.py` does not read `rankDerivedValue` — per your own F-6 material, it reads `_finalAdjusted`, which never passes through the clamp, the IDP calibration pass, hierarchical anchoring, pick tethering or the year discount. Two smaller corrections: the 0.15 is a **ceiling** on the band (`band = min(bucket P90, 0.15)`), so the effective corridor is usually tighter than 15%; and the gather loop skips only `assetClass == "offense"`, so **picks are clamped too** — your table omits them. **The concern relocates and gets better:** the corridor genuinely binds the board `/rankings`, the player popup and `suggestions.py` all serve, it is undisclosed in the UI, and it means the two engines work from systematically different IDP values. Nobody has measured that divergence — now §16.9 #3, and it is cheap to answer.

**3. "No liquidity or package-complexity discount — a 3-for-1 prices identically to a 1-for-1."** `f_simplicity = -(len(give) + len(receive) - 2) × 3` is a live term in the arbitrage score; a 3-for-1 takes **−6**. Your formula listing in §6.5 omits it (the worked example still comes out right, since `pkg_size = 2` zeroes the term). **The narrower claim you also make does survive and is ported:** package *value* is a plain sum with no consolidation superadditivity or fragmentation subadditivity. Complexity is priced in the **ranking**, not the **valuation**.

**Arithmetic corrected (2):**

**4. "`team_ros_strength` at 0.38 effectively becomes 0.69."** The mechanism concern is resolved — `power_v2` renormalizes explicitly and documents it. The figure is also wrong: surviving weight is 0.38+0.04+0.03 = **0.45**, so 0.38/0.45 ≈ **0.844**. 0.69 divides by the *missing* weight rather than the surviving weight. **Closed, not open.**

**5. "`.env.example` is missing at least 6 variables," listing `UPTIME_CHECK_ENABLED`.** Five of six confirmed; `UPTIME_CHECK_ENABLED` **is** declared. Ported in corrected form — and the underlying point stands, given #554's root cause.

**Narrowed (1):**

**6. "The 0.20–0.25 playoff band is entirely unreachable as Selective Seller."** Conditional on championship odds < 0.02. At playoff 0.22 / championship 0.03 the `Seller` clause fails and `Selective Seller` is reached. Partially, not entirely, shadowed. Still worth a test.

**Stale — superseded by merges after `c534280d` (4):**

**7. "Both engines enforce a KTC top-150 gate."** Superseded by **#556** before this PR was written. `suggestions.py` gates on `BOARD_TOP_N_FILTER` against our blended board; `finder.py` gates on `MARKET_TOP_N_FILTER` **per market**, ranked within each market's own population. Old names survive as deprecated aliases.

**8. Finder gates "every outgoing asset has a KTC value ≥ 500" and the IDP dilution guard.** First is now per-market (`MIN_MARKET_VALUE`, on the asset's own board). **The dilution guard was deleted in #556** — its premise died when IDP assets gained a real anchor, and it had been unreachable in production anyway.

**9. Line numbers for `finder.py` (735 L) and `suggestions.py` (1,625 L).** #556 grew them to **892** and **1,699** and shifted every symbol. Not reproduced anywhere in §16. Your `data_contract.py` and `player_valuation.py` references were re-checked and **are accurate**.

**10. Status claims** — "five PRs open, none merged," `main` at `c534280d`, the WS-A/B/C table. `main` is at `253568bc`; #556 and #551 have merged.

**Stale — wrong on `main` for reasons predating this branch (2):**

**11. "Hetzner VPS … nginx + systemd + Let's Encrypt SSL."** Production is a Contabo VPS on a **bare IP over HTTP with no TLS**, following the domain-loss incident recorded in the merged §2 Stage 9 — which your session had no visibility into. Worth knowing where it came from: **`docs/status/` still carries Hetzner references** in at least two files. A small argument for the retention policy you recommend.

**12. "`config/tiers/thresholds.json` … to be replaced by `scripts/fit_tier_thresholds.py`."** Quoted accurately from the config, but the config is wrong: **that file does not exist.** The real fitter is `refit_tier_thresholds.py` and **no workflow invokes it.** Your suspicion was right and understated — see below.

**13. "`grant-ssh-access.yml` … no recorded reason."** Answered by the merged §2 Stage 9, as above.

---

## Three of your open questions, resolved

You flagged these and in one case said you should have resolved it rather than flagging it. Resolved in #561:

- **`run_valuation` is dead in production.** Every reference outside `player_valuation.py` is either the `src/canonical/__init__.py` re-export or `tests/canonical/test_player_valuation.py`. Dead code kept alive by its own tests. **But** `rank_to_value`, `percentile_to_value` and `detect_tiers` in the same module **are** live — imported by `data_contract.py`, `terminal.py` and `rank_history.py`. So the module is *partially* live and cannot simply be deleted; the dead surface is `run_valuation` plus `W_MEDIAN`/`W_MEAN`/`CLIFF_*`/`VOL_*` and the two adjustment functions. It needs a decision now, not an investigation.
- **`power_v2` renormalizes correctly** (above).
- **The tier fitter has never run.** Stronger than your suspicion: the config names a script that does not exist, the real one is unwired, so the thresholds are *definitionally* still hand-set priors.

**One finding neither document had**, surfaced while checking yours: the docstring above `HILL_ROOKIE_PERCENTILE_C` in `player_valuation.py` still says the ROOKIE master is *"Used for every rookie-only source's contributions."* That is stale — it describes the pre-2026-04-21 routing that `_curve_for_source` retired. Your §4.2 table (`routed: false`) is correct; the docstring contradicts it. A reader trusting the docstring over the routing gets the wrong model.

---

## On the TE premium

Your H1 and Day-2 item 6 read the ≈1.12-vs-1.15 gap as a contradiction awaiting a decision. §16.9 closes with the argument that it is better read as **a measurement awaiting an independent reference**: ≈1.12 was derived under the assumption that KTC's standard board targets a 2-FLEX superflex league, and no check sharing that assumption can confirm it. A generic 1-TE league with one flex slot would lower the reference endpoint and *raise* the premium. That is exactly why the merged §4A.2 calls it "a finding, not a published value" and holds the TE axis `ABSENT` in `adjustment.py` so it cannot stack on the existing ×1.15. **Adopting 1.12 on the current evidence would be the same error as the retracted 1.316-vs-1.320 "corroboration," in the opposite direction.**

---

## Assessment

Two of the sharpest things in the reconciled document came from here: the **α = 0 backtest history** and the **registry/`DEFAULT_STARTER_NEEDS` staleness stated as two distinct states** — still broken on `main`, fixed only on unmerged #550, **with the duplication surviving the fix**. Your point that `DEFAULT_STARTER_NEEDS` remains a hardcoded dict after #550 corrects its values, and will go stale again, is the correct read and was not in the merged plan. The parity test you propose is now §16.9 #9.

The provenance warning in §0 also did its job. Everything sourced to "read directly from source at `c534280d`" held up under re-verification; the failures clustered exactly where you flagged lower confidence — status claims, chronology framing, and one arithmetic slip. **Your own §12.1 admission that you read ~3,000 of ~27,000 lines and that `run_valuation` "should have been resolved rather than flagged" was accurate self-assessment**, and it is why the salvage went to source rather than trusting the text.

Closing as superseded by **#561**.

---

Issue #555, opened 2026-07-26
Auth/proxy defects surfaced by E2E: proxy serves logged-out shell to signed-in sessions, missing page routes, /league SSR over timeout, useAuth gives up after 5s


jasonleetucker-code:
Five defects the E2E work surfaced. All are **server.py / frontend-auth territory** and were deliberately *not* fixed inside #554 — that PR only works around them in test code so the suite can run. Filed together because four of the five share one root: the backend's page proxy and the client auth check disagree about who is signed in.

Evidence below is from real runs against a locally-booted stack on the committed snapshot (no mocks).

---

## 1. The page proxy serves the ANONYMOUS shell to signed-in sessions

**Highest severity here.** With a valid session cookie, `GET /` through the backend (`:8000`) renders the logged-out landing/sign-in shell, while the *same cookie* against Next.js directly (`:3000`) renders the authenticated dashboard.

Reproduced side by side in one browser context:

| Origin | `/api/auth/status` | Rendered `/` |
|---|---|---|
| `:8000` (backend proxy) | `{"authenticated":true,"username":"jasonleetucker"}` | `… ▪ BRISKET  Trade  League  Login … Sign In` |
| `:3000` (Next directly) | `{"authenticated":true,"username":"jasonleetucker"}` | `… Pick your team ▾  Search  System …` |

So the API agrees the user is authenticated while the proxied page shows "Sign In".

Production routes page traffic through nginx straight to Next, so users are unlikely to hit this today — but it means **the proxy path is not a faithful mirror of production**, and anything relying on it (synthetic checks, uptime probes, anyone hitting `:8000` directly) sees a logged-out site. Worth deciding whether the proxy should be fixed or explicitly declared non-production-representative.

## 2. `/waivers` and `/news` are missing from the proxy's page-route list

`server.py` registers page routes explicitly (`/rankings`, `/trade`, `/settings`, `/edge`, `/finder`, `/trades`, `/rosters`, `/draft-capital`, `/more`, `/admin`, `/login`, …). `/waivers` and `/news` are not among them, so through `:8000` they return:

```
GET :8000/waivers  → 404 {"detail":"Not Found"}
GET :8000/news     → 404
GET :3000/waivers  → 200
```

Same practical impact as #1 (nginx bypasses it in prod), but it's a list that has to be hand-maintained and has already drifted twice. A catch-all forward, or generating it from the Next route manifest, would stop the drift.

[...]

PR #558, opened 2026-07-26
fix(auth): transient status timeouts no longer sign users out; complete the page-route list

jasonleetucker-code:
Fixes the two defects from #555 that are actionable now. The proxy's auth behaviour is **deliberately left alone** pending the "declare non-production-representative" decision.

## 1. A transient status timeout is no longer a sign-out

`useAuth` capped `/api/auth/status` at 5 s with an `AbortController` and, on timeout with no cached flag, resolved `authenticated = false` **with no retry**. One slow backend moment stranded a *valid* session on the anonymous shell for the entire life of the page — no nav, no search, no team switcher — until a manual reload.

**The 5 s cap stays.** It exists so a hung fetch can't wedge the UI on `checking = true`, which is a real failure mode. What changes is the *terminal* resolution.

`authenticated` is now honestly three-valued:

| value | meaning |
|---|---|
| `true` | the backend said so |
| `false` | the backend said so — a real signed-out user |
| `null` | unknown: still probing, or every probe so far failed |

Only a **200 is authoritative**. `/api/auth/status` is public and reports signed-out users as `200 + {authenticated:false}`, so a 5xx, an abort, or a network error tells us nothing about this user and must not be recorded as one. Those leave the state unknown and schedule another probe with backoff (1s, 2s, 4s, 8s, then every 30s), so a session unreachable at mount recovers on its own. `focus` and `online` also re-probe immediately, so a tab backgrounded or offline through the whole backoff recovers the moment it returns.

Pre-existing behaviour preserved and pinned: a cached "was signed in" still paints optimistically through a blip, and a real `no` still signs a stale cache out.

`authUnknown` is exposed so callers can distinguish "we never got an answer" from "the answer was no" without re-deriving it.

## 2. Page routes: forward what's unclaimed instead of 404ing

The hand-maintained route list had drifted **by more than was reported**. `/waivers`, `/news` and `/draft` were the known cases; the real count on this tree is **twelve** — `/angle`, `/trending`, `/players/compare`, `/league-comparison`, `/idptc-rookies`, `/design`, `/tools/ros-data-health` and the `/league/*` subroutes (`/league/activity`, `/league/phases`) all 404'd through the backend while serving fine from Next.

Replaced with a catch-all registered as the last route in the file. **Chosen over generating from Next's build manifest** because it's smaller, needs no build artefact at import time, and can't drift again by construction.

- Ordering is load-bearing (FastAPI matches in registration order), so the `/api/` and `/_next/` guards inside are belt-and-braces for the day someone appends a route below it.
- Auth gating preserved: pages default to requiring a session, `/league` and its subroutes stay public — matching the existing hand-written handlers.

## On declaring the proxy non-production-representative

**You're right, and the evidence is stronger than "nginx bypasses it."** `_proxy_next` forwards neither the request's cookies nor its query string. That's not a bug to patch, it's the absence of a reverse proxy: making it faithful means implementing cookie passthrough, query strings, header forwarding, streaming and status semantics — reimplementing nginx for a path production never takes. Declaring it non-production-representative is the correct call, and I'd add that the route list is still worth completing regardless, since dev and synthetic paths do use it and a 404 there costs real debugging time (it cost two agents exactly that).

So this PR makes the route list complete, **not** the proxy faithful. The cookie/query gap is documented inline at the catch-all so the next reader doesn't mistake it for an oversight.

## Deliberately not touched

- **`/league` SSR performance** (7-19 s) — real, but a separate perf investigation.
- **The proxy's auth mismatch** — pending the decision above.
- **Three-valued state in the shell.** `AppShellWrapper` passes `auth.authenticated === true` down, so `null` still renders as logged-out chrome; the user-visible win here is that it now *recovers in ~1 s* instead of never. Rendering a distinct "unknown" chrome is a design decision that belongs with the shell owner, not smuggled in here.

## Verification

- **`useAuth`: 8 new tests, 6 fail against the previous implementation.** Driven by simulated aborts and 5xx with fake timers rather than wall-clock assertions, since the defect is timing-dependent.
- **Route coverage: 32 tests, 14 fail against the previous server.** The test derives the Next route table from the App Router file tree instead of hardcoding it, so a new page joins the check automatically — the property the hand-maintained list lacked — and it guards itself against vacuously passing on an empty tree.
- Frontend vitest: **55 files / 1173 tests green**.
- Backend auth + route suites (`test_private_auth`, `test_page_route_coverage`, `test_cache_control_privacy`, `test_test_session_endpoint`): **70 passed**.
- `ruff format` + `ruff check` clean on `server.py`.

---

### DIFF
--- a/frontend/__tests__/components/useAuth.retry.test.jsx
+++ b/frontend/__tests__/components/useAuth.retry.test.jsx
@@ -1,0 +1,167 @@
+// A transient failure on /api/auth/status must NOT be reported as a
+// signed-out user.
+//
+// The bug these pin: the hook capped the probe at 5s with an
+// AbortController and, on timeout with no cached flag, resolved
+// authenticated = false with no retry.  A valid session was then
+// stranded on the anonymous shell for the entire life of the page —
+// no nav, no search, no team switcher — until a manual reload.
+//
+// These are deliberately driven by SIMULATED aborts/failures and fake
+// timers rather than real elapsed time: the defect is timing-dependent,
+// and asserting on wall-clock behaviour would be slow and flaky.
+import { afterEach, beforeEach, describe, it, expect, vi } from "vitest";
+import { act, renderHook } from "@testing-library/react";
+import { useAuth } from "@/components/useAuth";
+
+function authResponse(authenticated) {
+  return { ok: true, status: 200, json: async () => ({ authenticated }) };
+}
+
+/** What fetch does when an AbortController fires — what the 5s cap produces. */
+function abortError() {
+  const err = new Error("The operation was aborted.");
+  err.name = "AbortError";
+  return err;
+}
+
+async function flush() {
+  await act(async () => {
+    await vi.advanceTimersByTimeAsync(0);
+  });
+}
+
+beforeEach(() => {
+  vi.useFakeTimers();
+  try {
+    sessionStorage.clear();
+  } catch {
+    /* jsdom always has it; guard anyway */
+  }
+});
+
+afterEach(() => {
+  vi.useRealTimers();
+  vi.unstubAllGlobals();
+});
+
+describe("useAuth — a timeout is not a sign-out", () => {
+  it("stays UNKNOWN (not false) when the probe aborts", async () => {
+    vi.stubGlobal("fetch", vi.fn(async () => { throw abortError(); }));
+
+    const { result } = renderHook(() => useAuth());
+    await flush();
+
+    // The load-bearing assertion.  Pre-fix this was `false`, which
+    // consumers render as a genuine signed-out user.
+    expect(result.current.authenticated).toBeNull();
+    expect(result.current.authUnknown).toBe(true);
+    // ...and the UI must never be left wedged on "checking".
+    expect(result.current.checking).toBe(false);
+  });
+
+  it("recovers a valid session on its own after transient aborts", async () => {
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
**Empirical confirmation that this fixes a live E2E flake — measured, not argued.**

I ran the full suite against `main` as it actually stands now (with R3 #551 merged), on a healthy stack: **146 passed / 3 failed**. One of the three was this exact defect:

```
signed-in-smoke.spec.js › home page renders with team switcher hydrated
Expected pattern: /Pick your team|Rossini|JasonLeeTucker|Team/i
Received string:  "…▪Brisket  Trade  League  Login … Sign In …"
```

A valid session rendering the logged-out shell — the anonymous-shell symptom, reproduced on merged `main` without me provoking it.

I then merged this branch onto that same tree, rebuilt the frontend, and re-ran the affected spec **8 times against the same running stack**:

```
REPEAT_RESULT pass=8 fail=0
```

Repetition rather than a single green run, because the defect is timing-dependent — one pass would have proved nothing.

So this isn't only a code-reading argument: the fix removes a failure that is currently happening on `main`, and it's the flake I originally reported as the suite's single largest source of noise. Each occurrence was always a real user-visible failure rather than a test artifact.

The other two failures in that run (`journey-settings-overrides` toggling, `public-league` franchise deep link) are unrelated and remain under watch.

---

PR #562, opened 2026-07-26
WS-J Roster Intelligence Engine: marginal contribution, per-position profiles, competitive window

jasonleetucker-code:
The roster half of WS-J (Roster & Trade Intelligence). Measures **what a roster is** by re-solving the exact best-ball lineup: marginal contribution per position, absence fragility, tradeable surplus, urgent need, and the competitive window as a five-state probability distribution.

Consumes `src/league_intel/` (exact scorer, exact optimizer, replacement levels, value schema) and `src/ros/` **as-is**. It never re-implements them, per the repo's one-live-path rule.

---

## Why this PR is based on `claude/league-intel-foundation`, not `main`

This branch is **not** independent WS-J work sitting on `main`. It merged `claude/league-intel-foundation` and builds directly on it. Basing it on `main` would replay ~19 League-Intelligence commits already under review in **#550**, producing a duplicate diff and a guaranteed conflict when both land.

So this is stacked: **#550 → this PR → #563**.

**Brought up to #550's head.** `claude/league-intel-foundation` had moved four commits ahead of what this branch contained (`928a1fa8`, `e1481162`, `1768519d`, `4bb81004` — the LI-8 work), which left this PR unmergeable against its own base. Merged in at `be23a42c`, **zero conflicts**. No generated artifact needed hand-resolution: `data/ros/*`, `data/scrape_state/*` and `CSVs/site_raw/*` all merged automatically, so no data value was chosen by me.

`928a1fa8` is the one worth knowing about — it makes `allow_scalar_fallback` a real flag in `src/league_intel/cross_market.py`, where it was previously declared and never read. **Nothing in `src/roster_intel/` on this branch calls `cross_market`**, so that change is inert here. It is live on #563, where `packages.py` does pass the flag; measured and documented there.

**Diff noise.** The two branches criss-crossed merge history, so `git merge-base` reports multiple bases. **Review `src/roster_intel/`, `tests/roster_intel/`, `src/ros/lineup.py` and `src/trade/suggestions.py`** — the rest is #550's and disappears once #550 merges and this PR is retargeted to `main`. Scoped to this workstream's own files: **17 files, +4,248 / −42**.

---

## What is here

| Module | What it does |
|---|---|
| `marginal.py` | Marginal value of a position = `S(full roster) − S(roster without that position)`, both sides an exact lineup re-solve |
| `profiles.py` | Per-position profiles built on the marginal engine — fragility, surplus, deficit, urgent need |
| `window.py` | Competitive window as a 5-state probability distribution via softmax, not a single label |
| `roster_source.py` | Builds the roster pool with the real eligibility index and ages |
| `engine.py` | `analyze_roster()` composes the three into one payload and **consumes** `src/ros/playoff_sim.py` rather than simulating a second time |

`src/ros/lineup.py` gains the `fantasy_positions` slot-eligibility join (ADR-007, reintroduced from the data side). `src/trade/suggestions.py` gets a 15-line touch.

## Corrections made inside this branch, kept in the history

Each is a place where an earlier commit on this branch shipped something that looked right and was not.

**Marginals were measured on the wrong population.** First computed on starter-sized rosters; re-measured on the FULL 53-58 man rosters, the only population where "remove this position" means anything.

**`deficit` measured concentration, not shortfall.** Deriving need as `fragility × marginal` reported **QB as the biggest need on a QB-strong roster** — a great room is concentrated by definition. Reproduced on the real league: the naive formula names QB the top need for the team with the highest QB contribution of all 12. `deficit` and `concentrationRisk` are now separate fields with their semantics stamped in the payload, and `deficit` is measured against replacement-level occupancy of the position's dedicated slots, never from fragility.

**The window's five probabilities summed to 1.0001.** Naive 4dp rounding broke the sum through serialization. Fixed with largest-remainder rounding.

**A fixture omission was misdiagnosed as a dead metric — and the correction is the important part.** Commit `e6eb0c92` reported that no committed artifact carries player ages and that the window's trajectory axis was therefore inert on real data. That was **wrong, and wrong in the most dangerous direction**: a fixture that starves an input produces exactly the same symptom as a metric that has collapsed. `data/public_league/nfl_players_full.json` has 12,201 entries, 10,954 with age; the fixture was joining rosters to the ROS aggregate alone, which carries neither ages nor eligibility.

Rewired through `roster_source.build_roster_pool`, measured on the 12 real rosters:
- eligibility **633/633** enriched, 247 multi-family players all carrying `fantasy_positions`, 0 without, 40 true hybrids — ADR-007 closed *on this data* rather than assumed closed
- ages **633/633** joined; trajectory spans **0.196–0.684** (sd 0.126) instead of exactly 0.500 for every roster
- the window reaches **all 5 states, up from 4**. `productive_struggle` was not rare on this league, it was *unreachable*: a pinned trajectory axis makes "young roster losing now" impossible to express regardless of the roster.

## What the playoff simulator does NOT give us, named rather than swallowed

`src/ros/playoff_sim.py` emits `ownerId / playoffOdds / byeOdds / topSeedOdds / missPlayoffsOdds / expectedWins / medianFinalSeed / mostLikelySeed / seedDistribution`. There is **no `championshipOdds` and there are no confidence intervals** — it stops at playoff qualification and never runs the bracket.

The engine originally read both and degraded to `None`, which would have presented a qualification-only simulation as a complete one with a missing half. It now reports *which* half it got. **Intervals stay `None` rather than collapsing to zero width, because zero width reads as certainty.** A mechanism test pins the main-branch row schema verbatim and asserts the notes fire, then feeds the LI-8-shaped row and asserts they do not — so the warnings cannot become unconditional and outlive the gap.

---

## Not implemented

Stated explicitly because omission is worse than admission. The directive covers six engines; this PR is **one of them**.

**Built, but in the stacked PR #563:** Target Position/Player engines, Trade Package Generator, Partner Fit & Acceptance Model.

**Not built here:**

- **Continuous Improvement System — nothing in this PR.** No champion–challenger infrastructure, no promotion gate, no historical backtesting of the new engines, no monitoring, no recalibration schedule. **This is owned by another agent on its own branch off `main` and is actively being built** — it is not orphaned work and nothing here should be taken as a claim that it will not exist.
- **Trade Opportunity Engine** as a distinct module. `src/trade/suggestions.py` and `src/trade/finder.py` are the pre-existing engines; nothing new was written against the directive's opportunity-engine spec.
- **No API endpoint and no UI.** `src/roster_intel/` is imported by **nothing** outside its own package and its tests — not `server.py`, not `src/api/`, not `src/trade/`, not `frontend/`. Every number here is reachable only from Python. The `/gameplan` surface does not exist.
- **`src/trade/angle.py` is not rewired.** ADR-010 built the single-market valuation path; `angle.py` still sums market values across a combo never constrained to one market. That is the live call site and it is untouched.
- Trade explanation / pitch surface, user controls, admin review tools.
- `categories_rescored` ROS projections — blocked upstream on a projection source publishing raw per-category numbers.

---

## Validation

Reported as a comment on this PR with real numbers, including anything red.

No frontend file is touched by this branch, so Redesign R3 (#551) — which deleted `frontend/components/terminal/Panel.jsx` and changed the design-system `Panel` API — has no surface here. Zero conflicts is not evidence a build passes, so the suite was run rather than inferred.

## Merge order

`#550` → retarget this PR to `main` → `#563`.

### DIFF
--- a/src/roster_intel/__init__.py
+++ b/src/roster_intel/__init__.py
@@ -1,0 +1,6 @@
+"""Roster Intelligence Engine (WS-J).
+
+Pure computation over a supplied player pool. Consumes the League
+Intelligence modules (exact scorer, exact best-ball optimizer,
+replacement levels, value schema) as-is and never re-implements them.
+"""
--- a/src/roster_intel/engine.py
+++ b/src/roster_intel/engine.py
@@ -1,0 +1,410 @@
+"""Roster-level rollups — the single entry point for WS-J.
+
+Composes the pieces into one per-roster payload: values, marginal
+strength, per-position profiles, the competitive window, and (when the
+caller supplies simulator output) playoff and championship probability.
+
+``src/ros/playoff_sim.py`` is CONSUMED, not duplicated.  It already
+models the remaining schedule, per-team variance and the bracket; a
+second simulator here would be a parallel path and the two would drift.
+This module takes its output as an input and reads the rows it needs.
+
+Fragility semantics — read this before exporting it
+───────────────────────────────────────────────────
+A consuming agent derived "deficit" as ``fragility x marginal`` and got
+QB reported as the biggest need on a QB-STRONG roster.  That inversion
+is not a bug in fragility; it is fragility being read as the wrong
+quantity.
+
+    fragility  = CONCENTRATION. Share of an absent starter's value the
+                 roster fails to replace. High fragility means the
+                 position's output is concentrated in few players.
+    marginal   = CONTRIBUTION. What the position adds to the optimal
+                 lineup.
+
+A great QB room in a superflex league is BOTH high-marginal and
+high-fragility: the two starters are excellent and nobody behind them
+is close.  Multiplying those reports the roster's greatest STRENGTH as
+its greatest need.
+
+Deficiency is a different measurement and is provided here as
+``PositionNeed.deficit`` so consumers stop deriving it: it is the gap
+between what the position contributes and what a replacement-level
+occupant of the same slots would contribute.  Low contribution relative
+to replacement is a deficit; high concentration is a RISK, and the two
+are reported separately because they call for different moves — a
+deficit says acquire a starter, concentration says acquire insurance.
+
+Pure computation.  No I/O, no network, no clock.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Any, Mapping, Sequence
+
+from src.league_intel.replacement import PositionReplacement
+from src.roster_intel.marginal import position_marginals, solve_summary
+from src.roster_intel.profiles import RosterProfile, build_position_profiles
+from src.roster_intel.window import CompetitiveWindow, compute_window
+from src.ros.lineup import RosterPlayer
+
+__all__ = [
+    "PositionNeed",
+    "RosterIntel",
+    "RosterValues",
+    "analyze_roster",
+    "position_needs",
+]
+
+
+@dataclass(frozen=True)
+class RosterValues:
+    """Value rollups across the parallel scales (LI-4 schema).
+
+    Kept as separate fields rather than one blended number: the whole
+    point of the LI-4 schema is that market, consensus and
+    league-adjusted can disagree, and a blend hides the disagreement
+    that makes a trade findable.
+    """
+
+    market: float = 0.0
+    consensus: float = 0.0
+    league_adjusted: float = 0.0
+    ros: float = 0.0
+    # Starting-lineup ROS value vs everything else.
+    starters_ros: float = 0.0
+    bench_ros: float = 0.0
+    pick_value: float = 0.0
+    priced_players: int = 0
+    unpriced_players: int = 0
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "market": round(self.market, 3),
+            "consensus": round(self.consensus, 3),
+            "leagueAdjusted": round(self.league_adjusted, 3),
+            "ros": round(self.ros, 3),
+            "startersRos": round(self.starters_ros, 3),
+            "benchRos": round(self.bench_ros, 3),
+            "pickValue": round(self.pick_value, 3),
+            "pricedPlayers": self.priced_players,
+            "unpricedPlayers": self.unpriced_players,
+        }
+
+
+@dataclass(frozen=True)
+class PositionNeed:
+    """Deficiency and risk, kept SEPARATE (see module docstring).
+
+    ``deficit`` answers "am I below replacement here" — acquire a
+    starter.  ``concentration_risk`` answers "would one absence hurt" —
+    acquire insurance.  A consumer that multiplies them reproduces the
+    QB inversion.
+    """
+
+    position: str
+    deficit: float
+    concentration_risk: float
+    replacement_baseline: float | None
+    actual_contribution: float
+    urgent: bool
+    reasons: tuple[str, ...] = ()
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "position": self.position,
+            "deficit": round(self.deficit, 3),
+            "concentrationRisk": round(self.concentration_risk, 4),
+            "replacementBaseline": (
+                round(self.replacement_baseline, 3)
+                if self.replacement_baseline is not None
+                else None
+            ),
+            "actualContribution": round(self.actual_contribution, 3),
+            "urgent": self.urgent,
+            "reasons": list(self.reasons),
+            "_semantics": {
+                "deficit": "contribution shortfall vs a replacement-level occupant; >0 means acquire a starter",
+                "concentrationRisk": "share of an absent starter's value not replaced; high means acquire insurance",
+                "warning": "do NOT multiply these; concentration is not deficiency (a strong room is high on both)",
+            },
+        }
+
+
+def position_needs(
+    profile: RosterProfile,
+    replacement: Mapping[str, PositionReplacement] | None = None,
+) -> dict[str, PositionNeed]:
+    """Deficit and concentration risk per position.
+
+    ``deficit`` = what a replacement-level occupant of this position's
+    dedicated slots WOULD contribute, minus what the roster actually
+    contributes.  Positive means the roster is below replacement there.
+
+    Deliberately NOT derived from fragility: fragility measures
+    concentration, and a strong concentrated room would invert into a
+    phantom need (the exact defect a consumer hit).
+    """
+    out: dict[str, PositionNeed] = {}
+    for pos, prof in profile.positions.items():
+        rep = (replacement or {}).get(pos)
+        baseline_per_slot = rep.value("starter") if rep else None
+        req = prof.required_slots
+
+        if baseline_per_slot is not None and req > 0:
+            baseline = baseline_per_slot * req
+            deficit = max(0.0, baseline - prof.marginal_points)
+        else:
+            baseline = None
+            deficit = 0.0
+
+        reasons = list(prof.need_reasons)
+        if deficit > 0 and baseline:
+            reasons.append(
+                f"contributes {prof.marginal_points:.0f} vs a replacement-level "
+                f"{baseline:.0f} across {req} dedicated slot(s)"
+            )
+
+        out[pos] = PositionNeed(
+            position=pos,
+            deficit=deficit,
+            concentration_risk=prof.fragility,
+            replacement_baseline=baseline,
+            actual_contribution=prof.marginal_points,
+            urgent=prof.urgent_need or deficit > 0,
+            reasons=tuple(reasons),
+        )
+    return out
+
+
+@dataclass(frozen=True)
+class RosterIntel:
+    """Everything WS-J knows about one roster."""
+
+    owner_id: str
+    values: RosterValues
+    lineup_score: float
+    filled_slots: int
+    total_slots: int
+    profile: RosterProfile
+    needs: dict[str, PositionNeed]
+    window: CompetitiveWindow
+    playoff_odds: float | None = None
+    championship_odds: float | None = None
+    playoff_odds_ci: tuple[float, float] | None = None
+    championship_odds_ci: tuple[float, float] | None = None
+    odds_source: str = "unavailable"
+    notes: tuple[str, ...] = field(default_factory=tuple)
+
+    def to_dict(self) -> dict[str, Any]:
+        return {
+            "ownerId": self.owner_id,
+            "values": self.values.to_dict(),
+            "lineupScore": round(self.lineup_score, 3),
+            "filledSlots": self.filled_slots,
+            "totalSlots": self.total_slots,
+            "positions": self.profile.to_dict()["positions"],
+            "needs": {k: v.to_dict() for k, v in sorted(self.needs.items())},
+            "competitiveWindow": self.window.to_dict(),
+            "playoffOdds": self.playoff_odds,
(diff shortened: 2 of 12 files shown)


jasonleetucker-code:
## Validation results

### This PR's own code

`tests/roster_intel` + `tests/league_intel` at head `be23a42c` (after merging `claude/league-intel-foundation`):

```
440 passed, 22 skipped in 8.70s
```

**0 failed, 0 errors.** `python -m ruff check src/roster_intel` — **All checks passed** on ruff 0.6.9 (the CI pin), invoked as `python -m ruff` per the CI convention.

### Wider suite, and the red that had to be fixed first

Full scope `tests/roster_intel tests/league_intel tests/trade tests/api tests/test_trade_finder.py tests/test_trade_suggestions.py`, measured on the sibling branch #563 which carries the same League-Intelligence base:

| State | Result |
|---|---|
| `main` @ `1ec3311b`, baseline | **1631 passed, 3 skipped, 0 failed** |
| WS-J stack **before** merging LIF to its head | **71 failed, 1900 passed, 160 skipped, 16 errors** |
| WS-J stack **after** merging LIF to its head | **2207 passed, 3 skipped, 0 failed, 0 errors** |

Reported rather than quietly discarded, because "zero conflicts" was already true in the middle row and the suite was red anyway.

**Every failure was in `tests/api/` — none in `tests/roster_intel/`.** The cause was a half-applied ADR-002: the branch carried LIF's rewrite of `config/leagues/registry.json` (`rosterSize 30→58`, `taxiSize 5→0`, `TE 1→2`, `DL/LB/DB 2→3`, `IDP_FLEX 2→0`, `K` added) without the four later LI-8 commits that bring the consumers and their tests into line with it. Merging LIF to its current head closes all 87, with **no test skipped, xfailed, or edited** to get there.

### Diff is now clean — the "review only these paths" caveat in the body is obsolete

Before the LIF merge, the criss-crossed history gave `git merge-base` multiple candidates and the diff carried ~13,000 lines of #550's content. After `be23a42c` it collapses to exactly this workstream:

```
 13 files changed, 3970 insertions(+)
 src/roster_intel/{__init__,engine,marginal,profiles,roster_source,window}.py
 tests/roster_intel/{__init__,test_engine,test_marginal,test_profiles,
                     test_real_rosters,test_roster_source,test_window}.py
```

Zero deletions, nothing outside `src/roster_intel/` and `tests/roster_intel/`. `src/ros/lineup.py` and `src/trade/suggestions.py` dropped out of the diff because those changes came from #550 and are now shared with the base rather than attributed here.

### Conflict resolution

None was needed. The LIF merge was **zero-conflict**, so no generated artifact — `data/ros/*`, `data/scrape_state/*`, `CSVs/site_raw/*` — required a hand-chosen side. No data value in this branch was picked by me.

`928a1fa8` (making `allow_scalar_fallback` real in `cross_market.py`) arrives with that merge but is **inert on this branch**: `grep -rn cross_market src/roster_intel/` returns nothing here, because `packages.py` lives on #563. Its live effect is measured and documented there.

---

jasonleetucker-code:
## Rebased onto `main` — the LI-8 revert is gone, and the tier defect is fixed

### The revert: resolved by rebasing, not by merging

I did not merge `main` in. #550 was **squash**-merged, so the old `claude/league-intel-foundation` commits on this branch were duplicate content that could only ever fight with `main`. Rebasing the 8 WS-J commits directly onto `origin/main` drops them entirely, which is strictly better than resolving them file-by-file — there is nothing left to resolve.

Your check, run on the new head:

```
git diff origin/main HEAD -- src/league_intel/ src/ros/ src/api/ config/ docs/
  → 0 bytes

src/league_intel/sim.py            main 374   HEAD 374
src/league_intel/cross_market.py   main 670   HEAD 670
PointsModel occurrences in sim.py  main   5   HEAD   5
```

The diff went from **+24 / −157** to **+4,221 / −0**. Every path is an `A`:

```
13 files changed, 4221 insertions(+)
src/roster_intel/{__init__,engine,marginal,profiles,roster_source,window}.py
tests/roster_intel/{__init__,test_engine,test_marginal,test_profiles,
                    test_real_rosters,test_roster_source,test_window}.py
```

**Zero deletions and zero modifications, so there is no revert of any shape.** `src/ros/lineup.py` and `src/trade/suggestions.py` dropped out of the diff entirely — those changes came via #550 and are now in main's squash, so this branch has nothing to say about them.

I took your warning seriously and did **not** trust `merge-tree`: a real `git merge` into a scratch worktree at `origin/main` reports **0 conflicts** and produces exactly the 13-file / +3,970 result.

### The tier defect: fixed, with the failing output you asked for

Reproduced first. On the synthetic 12-team league, `bestBallStarter <= starter` at 5/5 positions and:

```
before: {'elite': 157, 'depth': 42, 'developmental': 5}
        77% elite — starter tier UNREACHABLE
```

**Why review missed it, which I think is the more useful finding.** `tests/roster_intel/test_profiles.py`'s `_rep` helper maps its first argument onto `bestBallStarter`, and the caller named that argument `elite` — passing `70.0` against `starter=45.0`. The fixture asserted `bestBallStarter` **above** `starter`, an ordering `replacement.py` cannot produce and which `tests/league_intel/test_replacement.py::test_tiers_are_ordered_starter_above_bestball_above_roster` already forbids. The test encoded the defect, so the code read as correct. That fixture is now corrected in place with a comment saying why.

**I did not swap the thresholds**, per your instruction. I also rejected the k×spread option: when a pool has no spread, `starter − bestBallStarter` collapses to 0, `starter + k×0 == starter`, and the `starter` tier goes unreachable again — the same bug in a new costume. Reachability had to come from the distribution.

`elite` is now a **positional percentile** (your option 1): the top-5% value of the position's rostered priced pool, via a new `elite_threshold()`. Deliberately the same definition `replacement.py:613` already uses for `eliteSeparation` (`pool[round(n * 0.05) - 1]`), so "elite" means one thing repo-wide rather than two. It is reachable **by construction** because the result is an actual member of the pool.

A percentile needs the league pool, which one roster cannot supply — 2-5 players at a position. Rather than edit `src/league_intel/replacement.py` (your file, and it would have broken the clean file list above), `build_position_profiles` takes an optional `league_position_values`. Omitted ⇒ **no player tiers elite**, matching the existing rule that an unmeasurable boundary must not promote. `classify_tier` additionally ignores an `elite` cut that is not strictly above `starter`, so a degenerate pool cannot resurrect the original failure mode.

### The six tests, run against the pre-fix code first

```
4 failed, 3 passed

FAILED test_every_tier_is_reachable
E  AssertionError: starter tier unreachable: {'elite': 157, 'depth': 42, 'developmental': 5}
E  assert 0 > 0
FAILED test_an_ordinary_starter_is_not_elite
E  AssertionError: assert 'elite' == 'starter'
FAILED test_missing_replacement_data_degrades_never_promotes
E  AssertionError: assert 'elite' == 'starter'
FAILED test_a_degenerate_threshold_cannot_make_everyone_elite
E  AssertionError: degenerate pool promoted 59% to elite: {'elite': 120, 'depth': 84}
E  assert (120 / 204) < 0.25
```

**The 3 that passed against the defect are the point.** Strict ordering (property 1) and "the best player is elite" (property 4) both hold when *everything* is elite. Your prediction that ordering alone would not catch it is exactly right — `test_every_tier_is_reachable` is the one that does.

All six build their levels from measured `compute_replacement_levels` output rather than synthetic constants, specifically so they cannot re-encode the inverted fixture that caused this.

### After the fix

```
tests/roster_intel                    117 passed, 22 skipped
tests/roster_intel + tests/league_intel  (on #563's head) 633 passed, 22 skipped
python -m ruff check src/roster_intel tests/roster_intel   All checks passed  (0.6.9)
```

### One thing worth recording for whoever works here next

`git rev-parse --is-shallow-repository` returns `false` for you and returned `true` for me — this container's clone was shallow, which silently produced **wrong** merge-bases and a spurious "multiple merge bases" warning. `git fetch --unshallow` fixed it. Ancestry results in a fresh container here should not be trusted until that is checked.

---

jasonleetucker-code:
## Format gate: your call, and your fix kept

You were right and I was wrong about the shape of it. I ran `python -m ruff check` (lint), saw "All checks passed", and wrote "ruff clean" — which is a claim about a gate I had not run. `ruff check` and `ruff format --check` are separate gates and only the second was red. That is the same adjacent-check failure as the rest of today, and it is the one I should have caught, because the gate name was in the error you would have seen.

**I kept `76db5d80` rather than force-pushing over it.** I had independently amended the formatting into the tier commit, and before pushing I compared trees:

```
my amended tree:  c0cf83d0164fa69124c899634a5187842eaf1be1
your 76db5d80:    c0cf83d0164fa69124c899634a5187842eaf1be1
```

Byte-identical. Forcing would have destroyed your commit to produce exactly the same bytes and burned a second gate cycle, so I took yours and rebased on top. My `force-with-lease` correctly refused the stale push, which is what caught the collision.

Verified at the current head `76db5d80`, using the gate's own commands rather than adjacent ones:

```
python -m ruff format --check .                            514 files already formatted
python -m ruff check src/roster_intel tests/roster_intel   All checks passed
pytest tests/roster_intel                                  117 passed, 22 skipped
deletions outside roster_intel                             0
non-ADD paths vs main                                      0
```

## ADR-013: it exists, but not on this branch — and that is deliberate

`git grep "ADR-013" origin/claude/ws-j-roster-intel` returning nothing is correct, because **this PR contains no `docs/` change at all** — `git diff --name-only origin/main...origin/claude/ws-j-roster-intel -- docs/` is empty.

ADR-013 is on **#563**, which is where the code it documents lives:

```
targets.py / horizon_weight   ws-j-roster-intel: absent
                              ws-j-partner-fit:  present

$ git show origin/claude/ws-j-partner-fit:docs/league-intelligence/DECISIONS.md
 41: ## ADR-013: the competitive window enters per POSITION as a horizon weight …
130: ## ADR-011: the trade primitive is a weekly scoring DISTRIBUTION (LI-8)
210: ## ADR-012: the LI-8 collision with claude/league-intel-sim, and who owns which half
```

Main's ADR-012 is intact at line 210 and ADR-011 at 130 — the file is a union, with zero deleted lines in `docs/`. Putting the horizon-weight ADR on #562 would have documented a function that does not exist on this branch.

So the collision you flagged is resolved, just one PR over. The numbering note belongs in #563's body and is there.

## Why review missed the tier defect — promoting this to the PR body, as you suggested

`tests/roster_intel/test_profiles.py`'s `_rep` helper populates levels positionally:

```python
def _rep(pos, elite, starter, roster):
    ...
    levels={
        "bestBallStarter": lvl("bestBallStarter", elite),   # ← first arg
        "starter":         lvl("starter",         starter),
        "roster":          lvl("roster",          roster),
    }
```

The parameter is **named `elite`** and lands on **`bestBallStarter`**. Callers then wrote `_rep("RB", elite=70.0, starter=45.0, roster=20.0)` — asserting `bestBallStarter (70) > starter (45)`, an ordering `replacement.py` cannot produce, since `starter` is the *median* of team marginals and `bestBallStarter` is the *mean of the lowest few*. `tests/league_intel/test_replacement.py::test_tiers_are_ordered_starter_above_bestball_above_roster` already forbids it.

So the fixture asserted the defect's premise, and against that fixture the code was self-consistent. A reviewer reading either the test or the implementation in isolation would find nothing wrong; only running the two modules' assumptions against each other exposes it. That is why the new tests build their levels from measured `compute_replacement_levels` output rather than synthetic constants — a fixture that can express an impossible ordering can hide any bug that depends on it.

The fixture is corrected in place with a comment recording this, so it cannot silently regrow.

---

PR #563, opened 2026-07-26
WS-J Trade layer: target engines, partner fit &amp; acceptance model, Trade Package Generator

jasonleetucker-code:
The trade half of WS-J. Consumes the roster engine from **#562** and answers *what to do about* a roster: which positions are worth attacking, which players are worth pursuing, how a given counterparty is likely to receive an offer, and which concrete packages are worth putting in front of them.

Valuation runs through `src/league_intel/cross_market.py`'s `value_package` / `compare_packages` — **single-market, exact, no fitted conversion**. Deliberately **not** through `src/trade/finder.py`, which was offense-only (that is #556's finding, and the reason a defender never reached the finder's arithmetic).

Scoped footprint: **10 files, +5,969 / −1**.

---

## Stack position and the divergence you need to know about

`#550 → #562 → this PR`.

**These two branches were a clean superset at the start of the session and are not any more.** `claude/ws-j-partner-fit` forked from `claude/ws-j-roster-intel` at `14d41f69`. Roster-intel has since gained three commits **not** in this branch:

| Commit | What it adds |
|---|---|
| `0cf29560` | `src/roster_intel/engine.py` — `analyze_roster()` rollup + `playoff_sim` wiring |
| `e6eb0c92` | Names the simulator's schema gaps instead of degrading to `None` |
| `b58aa48b` | Feeds the livedata fixture the eligibility index and ages it was starving |

**What conflicts when this lands.** Nothing textually — those commits add `engine.py`, which this branch does not have. The real coupling is semantic: `b58aa48b` changes what the competitive window produces on real data (trajectory goes from a pinned 0.500 to a 0.196–0.684 spread; the window reaches 5 states instead of 4). **The target engine in this PR consumes that window.** Its ranked output on live rosters will change once #562's head is merged in — the code is correct either way, but any output snapshot taken from this branch alone was taken against a starved trajectory axis.

## Brought up to date, and one live semantic change

**`main` merged in at `7671c9dd`, zero conflicts.** This also **deletes `.github/workflows/grant-ssh-access.yml`**, which matters: `main` removed it in `c534280d`, this branch still carried it, and `docs/ORCHESTRATION.md` is explicit that it *must* be deleted — it is a dispatch-only path to a production shell. Before this merge, landing this branch would have **resurrected** it.

**`claude/league-intel-foundation` merged in at `c8cd5179`, zero conflicts** — and this one is textually clean but semantically live. `928a1fa8` makes `allow_scalar_fallback` a **real** flag in `cross_market.py`; it was previously declared and never read, so a caller that opted out of scalar conversion silently got converted totals anyway.

This branch is that caller — `packages.py:502` passes the flag through and `generate_packages` defaults it to `False`. Measured directly rather than inferred from the diff, same mixed offense/IDP package:

```
allow_scalar_fallback=True   ->  total = 6605.5  (market: idpTradeCalc)
allow_scalar_fallback=False  ->  total = None    (unvaluable)
```

So packages needing a cross-market conversion are now **rejected as `UNVALUABLE` instead of ranked on a converted total** — the direction ADR-010's materiality gate intends, and what the caller was already asking for.

> **For the reviewer:** `tests/roster_intel/test_packages.py` is **47/47 green both before and after** that merge, which means the existing tests do not exercise a package needing conversion under the default flag. The behaviour change is real and currently **uncovered**. That gap is worth closing before this ships.

---

## What is here

| Module | What it does |
|---|---|
| `src/league_intel/cross_market.py` | `value_package` / `compare_packages` — every package priced entirely within one market (offense-only → `ktcSfTep`; anything with IDP → `idpTradeCalc`), plus the boundary-proximity suppression rule |
| `partner.py` | Partner fit + acceptance model, two-sided need alignment, pair familiarity |
| `targets.py` | Target Position and Target Player engines |
| `packages.py` | Trade Package Generator — staged search, Pareto frontier |
| `docs/…/DECISIONS.md` | **ADR-012** (new, this PR) |

Generation is **staged, not brute force**: 1-for-1 from a pre-filtered seed set → 2-for-1 / 1-for-2 expanding *only* near-misses → 2-for-2 from stage-2 near-misses. A package that already succeeded is never expanded, because the simpler version dominates it by construction. Cheap filters run before expensive ones, so the exact lineup re-solve never runs on a package already going to be rejected.

---

## Five things a reader would otherwise assume are stronger than they are

**1. `winNowWeight` reports `null` and is genuinely not applied when candidate ages are unavailable.**
`horizon_weight(window, mean_age)` returns `None` when there is no window *or* no age data for that position's candidates. `None` means **the term did not apply**. The caller sets `window_mult = 1.0` and, when a window was supplied but ages were not, appends an explicit note: *"competitive window supplied but no candidate ages — the window weight is per-position by age and could not be computed, so it did NOT affect this ranking."* The field serializes as `winNowWeight: null`.

Stamping the old global scalar there instead would have reproduced the original defect one layer down — a number that looks applied, is reported as applied, and changes nothing. **`winNowWeight` is not safe to sort or filter on**: it is `null` on exactly the rows missing age data, which is not a random subset (rookies and recent adds are likelier to be missing). Read `priority` for ordering and `notes` for whether the window participated. Recorded as **ADR-012**, added in this PR.

**2. The Pareto frontier comes out wide, and that is structural, not a bug.**
`value_efficiency` and `acceptance_plausibility` are **anti-correlated by construction** — `market_gain_pct` is *our* gain, so a genuine overpay to them is a loss to us. Two of the four axes therefore move in opposite directions almost by definition. For **1-for-1 packages nearly everything is Pareto-optimal**, and frontiers come out long.

A caller expecting a short list should expect the opposite, and should not read frontier length as a quality signal. `label_frontier()` names which axis each member wins on precisely so the set reads as *alternatives* rather than a ranking. There is deliberately no blended score: an exchange rate between market points, lineup points and plausibility does not exist, and inventing one silently encodes a preference as a measurement.

**3. Cornerstone detection needs rank AND share, and reads the partner's full roster.**
Top-N rank alone made **every asset in a short pool a cornerstone** — including a 2,200-value bench receiver sitting beside a genuine stud. Detection now requires top-`CORNERSTONE_TOP_N` (3) **and** ≥ `CORNERSTONE_VALUE_SHARE` (0.60) of the roster's best asset.

It is also derived from the partner's **full roster** when we have it, not from the subset of their players we happen to want — deriving it from the subset would let the asset we are asking for define its own peer group.

**4. The fifth rejection rule is an addition beyond the original spec.**
The directive specified four: illegal roster, worsens a critical need, unvaluable / verdict-uncertain, cornerstone without premium. `DOMINATED_BY_NO_TRADE` is a **fifth rule added here, not part of that spec.**

Its argument: the simplest package of all is no package, and it is always available. A trade that loses market value **and** loses lineup points is dominated by declining to trade, so high acceptance plausibility cannot rescue it — plausibility measures how the *partner* receives the offer, and being easy to give away is not a reason to give something away. Treat it as a proposal, not a spec item already agreed.

**5. The field is `acceptancePlausibility`, never `acceptanceProbability`.**
**We never observe a rejection.** There is no record of declined offers anywhere in this repo, so there is nothing to calibrate against. It is a plausibility score wearing probability units, and every package carries that caveat. `describe_limitations()` states that no output may be rendered as *"manager X will accept this"*, and the tests assert it.

The same gap is why `maxPackageAssets`, `nearMissPct`, `cornerstonePremiumPct` and `cornerstoneTopN` are all stamped `thresholdsAreMeasured: False` — assumed, not fitted.

## Other honest limits, from `describe_limitations()`

- **Search is not exhaustive.** Generation is staged and budgeted; the frontier is valid for what was explored, not proven globally optimal.
- Packages beyond the asset cap are out of scope — a deal needing three pieces per side to balance is negotiated, not generated.
- Lineup improvement is **rest-of-season only**; no multi-year or rebuild framing.
- ADR-012's six constants (`_WIN_NOW_WEIGHTS`, `_STATE_FUTURE_ORIENTATION`, `_AGE_YOUNG`/`_AGE_OLD`, `_FUTURE_AGE_FLOOR`, `_FUTURE_AGE_SPAN`) are judgement calls. The *directional* claim — a rebuilder should prefer young acquisitions, a contender should not care — is first-principles; the magnitudes are not measured.

---

## Not implemented

Stated explicitly because the governing directive said not to silently omit requirements.

- **Continuous Improvement System — nothing in this PR.** No champion–challenger infrastructure, no promotion gate, no historical backtesting, no monitoring, no recalibration schedule. **Owned by another agent on its own branch off `main` and actively being built** — not orphaned, and nothing here should be read as a claim it will not exist.
- **Trade Opportunity Engine** as a distinct module. `suggestions.py` / `finder.py` are the pre-existing engines; nothing new was written against the directive's opportunity-engine spec.
- **No API endpoint and no UI.** `src/roster_intel/` is imported by **nothing** outside its own package and its tests — not `server.py`, not `src/api/`, not `src/trade/`, not `frontend/`. Everything here is reachable only from Python. The `/gameplan` surface does not exist.
- **`src/trade/angle.py` is not rewired.** This is the one that matters most. ADR-010 built the single-market path and `packages.py` uses it, but `angle.py::_make_candidate` still sums market values across a combo never constrained to one market — on the offer side (`:378-388`, `:471-478`, `:526`) and the acquire side (`:913`). **That is the live, user-reachable call site and it is untouched by this PR.** Per the audit, the genuinely reachable shape is the offer side, which has no IDP gate at all.
- Trade explanation / pitch surface; user controls; admin review tools.
- `categories_rescored` ROS projections — blocked upstream on a projection source publishing raw per-category numbers.
- **`SHARED_SCALE_ASSUMPTION` is unvalidated and load-bearing.** The two boards interleave (n=475 paired, median ratio 1.000, p10 0.888 / p90 1.054), which establishes *coherence*. It does not establish that IDPTC's internal offense↔IDP exchange rate is *correct*. There is no ground truth for what an edge rusher is worth in WR points, and nothing here supplies one.

---

## Validation

Reported as a comment on this PR with real numbers, including anything red. `python -m ruff check src/roster_intel src/league_intel/cross_market.py` — clean on ruff 0.6.9 (the CI pin).

No frontend file is touched by this branch, so Redesign R3 (#551) has no surface here. Zero conflicts is not evidence a build passes, which is why the suite was run rather than inferred.

## Merge order

`#550` → `#562` → merge #562's head into this branch → this PR.

### DIFF
--- a/docs/league-intelligence/DECISIONS.md
+++ b/docs/league-intelligence/DECISIONS.md
@@ -38,6 +38,95 @@ adoption land immediately after R2 merges, through the stable R1 shell
 unvalidated number.
 **Status:** accepted.
 
+## ADR-014: the competitive window enters per POSITION as a horizon weight, not as a global win-now scalar (WS-J)
+**Status:** accepted 2026-07-26. Implemented in
+`src/roster_intel/targets.py::horizon_weight`; `win_now_weight` is
+retained for magnitude reporting only.
+
+**The defect.** `rank_target_positions` multiplied every position's
+priority by a single scalar, `win_now_weight(window)` — the
+probability-weighted mean of per-state weights (championship_contender
+1.00 → rebuild 0.15). That scalar is *uniform across positions* by
+construction, and a uniform positive multiplier cannot reorder a list.
+
+The consequence was not a rounding artifact, it was the whole feature:
+a rebuilding roster and a championship contender, given the same
+candidate pool, received the **identical ranked order** of target
+positions and differed only in the absolute magnitude of `priority`.
+The docstring and the `notes` stamp both claimed the competitive window
+affected the recommendation. It provably could not. This is exactly the
+class of defect §2b exists to catch — a term that is *reported* as
+applied while being incapable of changing any decision.
+
+**Why a global scalar cannot be repaired in place.** For the window to
+reorder positions it must interact with something that varies *by*
+position. `win_now_weight` reads only `window.probabilities`, which has
+no position axis. No re-tuning of the five state weights fixes this;
+the term is structurally inert. It had to be replaced, not calibrated.
+
+**The mechanism chosen: horizon, keyed to the age of what is obtainable.**
+`horizon_weight(window, mean_candidate_age)` keeps the same five-state
+distribution but modulates it per position by the mean age of that
+position's top-`REALISTIC_CANDIDATE_DEPTH` candidates:
+
+```
+youth     = clamp((AGE_OLD - mean_age) / (AGE_OLD - AGE_YOUNG), 0, 1)
+age_mult  = (1 - future) + future * (FUTURE_AGE_FLOOR + FUTURE_AGE_SPAN * youth)
+weight    = Σ_state p(state) · base(state) · age_mult(state)   /   Σ_state p(state)
+```
+
+`_STATE_FUTURE_ORIENTATION` (championship_contender 0.00 → rebuild 1.00)
+decides how much a state cares about age at all. Win-now states ignore it
+— a contender should take the best available upgrade regardless of age.
+Future-oriented states reward youth, so a rebuilding roster prefers
+positions where the *obtainable* upgrades are young. Same window,
+different positions, genuinely different order.
+
+This was picked over the two alternatives because it is the only one
+that needs no new data and no fitted parameter. Positional value curves
+by window would require a measured contender/rebuilder value split that
+does not exist for this league; a per-position win-now prior would be
+invented weights wearing a measurement's clothes.
+
+**What is measured and what is assumed.**
+
+*ASSUMED, not measured* — all of it. `_WIN_NOW_WEIGHTS`,
+`_STATE_FUTURE_ORIENTATION`, `_AGE_YOUNG`/`_AGE_OLD`, `FUTURE_AGE_FLOOR`
+and `FUTURE_AGE_SPAN` are judgement calls. There is no dataset of
+"positions a rebuilding manager should have attacked" to fit them
+against. `describe_limitations()` stamps `winNowWeightsAreMeasured:
+False` and the constants are emitted verbatim so a reader can see the
+numbers rather than trust the label.
+
+*The directional claim is the defensible part.* That a rebuilder should
+prefer young acquisitions and a contender should not care is a
+first-principles consequence of dynasty roster construction, not a
+fitted result. The **ordering** the term produces is therefore more
+trustworthy than the magnitudes it multiplies by.
+
+**The honest failure mode, and why it is `None` and not `1.0`.**
+`horizon_weight` returns `None` when the window is absent OR when no
+candidate at that position has a known age. `None` means *the term did
+not apply*. The caller then sets `window_mult = 1.0` and, when a window
+was supplied but ages were not, appends an explicit note:
+
+> competitive window supplied but no candidate ages — the window weight
+> is per-position by age and could not be computed, so it did NOT affect
+> this ranking
+
+`PositionTarget.win_now_weight` is serialized as `winNowWeight: null` in
+that case. Stamping the global scalar there instead would have
+reproduced the original defect one layer down: a number that looks
+applied, is reported as applied, and changes nothing. A caller must be
+able to distinguish "the window said 1.0" from "the window was not
+consulted", and only `null` does that.
+
+**Consequence for callers.** `winNowWeight` is not a reliable field to
+sort or filter on — it is `null` on exactly the rows where age data is
+missing, which is not a random subset (rookies and recently-added
+players are likelier to be missing). Read `priority` for ordering and
+`notes` for whether the window participated.
+
 ## ADR-013: the League Twin bridge scales for the ROS blend, and reports the variance the odds path drops (LI-8)
 **Status:** accepted 2026-07-26 on `claude/li8-league-twin`.  Library +
 tests; no endpoint, nothing user-visible.
--- a/src/roster_intel/__init__.py
+++ b/src/roster_intel/__init__.py
@@ -4,4 +4,19 @@ Additive layer over the League Intelligence engine. Valuation, exact
 scoring, the best-ball optimizer, replacement levels and the value
 schema are consumed from ``src/league_intel/`` as-is — nothing here
 re-derives them.
+
+Two halves that meet at ``RosterProfile`` / ``CompetitiveWindow``:
+
+* **Roster engine** — ``marginal``, ``profiles``, ``window``,
+  ``roster_source``, ``engine``. Measures what a roster IS by
+  re-solving the exact lineup: marginal contribution, absence
+  fragility, tradeable surplus, urgent need, and the competitive
+  window as a five-state distribution. ``engine.analyze_roster``
+  composes them and consumes ``src/ros/playoff_sim`` rather than
+  simulating a second time.
+* **Trade engine** — ``partner``, ``targets``, ``packages``. Measures
+  what to DO about it: which positions are worth attacking, which
+  players are worth pursuing, how a counterparty is likely to receive
+  an offer, and which concrete packages are worth sending. Consumes
+  the roster engine's output; never re-derives it.
 """
--- a/src/roster_intel/__init__.py
+++ b/src/roster_intel/__init__.py
@@ -1,6 +1,7 @@
-"""Roster Intelligence Engine (WS-J).
+"""Roster & Trade Intelligence (WS-J).
 
-Pure computation over a supplied player pool. Consumes the League
-Intelligence modules (exact scorer, exact best-ball optimizer,
-replacement levels, value schema) as-is and never re-implements them.
+Additive layer over the League Intelligence engine. Valuation, exact
+scoring, the best-ball optimizer, replacement levels and the value
+schema are consumed from ``src/league_intel/`` as-is — nothing here
+re-derives them.
 """
--- a/src/roster_intel/packages.py
+++ b/src/roster_intel/packages.py
@@ -1,0 +1,918 @@
+"""Trade Package Generator (WS-J).
+
+Builds concrete, legal, defensible trade packages between our roster and
+one opposing roster, and returns a **Pareto frontier** across competing
+objectives rather than a single ranked list.
+
+Consumes, never rebuilds:
+
+* ``league_intel.cross_market.value_package`` / ``compare_packages`` —
+  package valuation and the boundary-proximity suppression rule. Every
+  package is valued **entirely within one market** (offense-only →
+  ``ktcSfTep``; anything with IDP → ``idpTradeCalc``), which is exact
+  and needs no conversion.
+* ``roster_intel.marginal.optimal_score`` — exact lineup re-solve for
+  what a package actually does to a starting lineup.
+* ``roster_intel.partner.assess_partner`` — how a counterparty is
+  likely to receive the offer.
+
+Pure computation: no I/O, no clock, no globals.
+
+──────────────────────────────────────────────────────────────────────
+A frontier, not a score
+──────────────────────────────────────────────────────────────────────
+Packages are judged on four axes that share no unit:
+
+* ``value_efficiency`` — market value retained, from WS-E
+* ``roster_improvement`` — lineup points gained, from an exact re-solve
+* ``acceptance_plausibility`` — from the partner model
+* ``simplicity`` — fewer moving parts
+
+Blending them into one number requires an exchange rate between market
+points, lineup points and plausibility. **No such rate exists**, and
+inventing one silently encodes a preference as if it were a
+measurement. So this module returns the non-dominated set: every
+package on the frontier is the best available at *something*, and the
+choice between them is the manager's to make. :func:`label_frontier`
+names which axis each member wins on, so a caller can present them as
+alternatives rather than as a ranking.
+
+──────────────────────────────────────────────────────────────────────
+Staged generation, not brute force
+──────────────────────────────────────────────────────────────────────
+The full combination space over two 30-man rosters is astronomically
+larger than the useful part of it. Generation runs in stages, and each
+stage only expands what the previous stage proved was **worth**
+expanding:
+
+1. 1-for-1 from a pre-filtered seed set.
+2. 2-for-1 / 1-for-2, expanding ONLY near-misses — packages that failed
+   on value gap but are close enough that one more asset could close
+   it. A package that already succeeded is never expanded, because the
+   simpler version dominates it by construction.
+3. 2-for-2, from stage-2 near-misses only.
+
+Within each stage, cheap filters run before expensive ones: legality,
+then market valuation, then the rejection rules, and only then the
+exact lineup re-solve. The solve is the expensive step and it never
+runs on a package that was already going to be rejected.
+
+──────────────────────────────────────────────────────────────────────
+On ``acceptance_plausibility``
+──────────────────────────────────────────────────────────────────────
+It is inherited from :mod:`roster_intel.partner`, and it carries that
+model's limitations unchanged: **we never observe a rejection**, so
+this is a plausibility score wearing probability units, not a
+calibrated acceptance probability. No output of this module may be
+rendered as "manager X will accept this". See
+:func:`describe_limitations`.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any, Iterable, Mapping, Sequence
+
+from src.league_intel.cross_market import (
+    DEFAULT_GATE_PCT,
+    PackageComparison,
+    compare_packages,
+    value_package,
+)
+from src.roster_intel.marginal import optimal_score
+from src.roster_intel.partner import (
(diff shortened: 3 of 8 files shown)


jasonleetucker-code:
## Validation results

Run scope for every number below: `tests/roster_intel tests/league_intel tests/trade tests/api tests/test_trade_finder.py tests/test_trade_suggestions.py`. No frontend file is touched by this branch, so `npm run build` was not exercised.

### It was red, and the fix was the merge

| # | State | Result |
|---|---|---|
| 1 | `main` @ `1ec3311b`, baseline (no `roster_intel`/`league_intel` — they don't exist there) | **1631 passed, 3 skipped, 0 failed** |
| 2 | partner-fit @ `bfbaaa8f` + `main`, **before** the LIF merge | **71 failed, 1900 passed, 160 skipped, 16 errors** |
| 3 | partner-fit @ `c8cd5179` (+ `main` + LIF), **current head** | **2207 passed, 3 skipped, 0 failed, 0 errors** |

Row 2 is reported rather than quietly discarded, because "zero conflicts" had already been true at that point and the suite was red anyway — which is the whole reason the suite gets run instead of inferred.

**Every one of those 71 failures and 16 errors was in `tests/api/`. Zero were in `tests/roster_intel/`, `tests/league_intel/`, `tests/trade/`, `test_trade_finder.py` or `test_trade_suggestions.py`.** Distribution:

```
24 FAILED tests/api/test_frontend_migration.py
16 ERROR  tests/api/test_faab_recommend_endpoint.py
12 FAILED tests/api/test_fantasypros_idp_integration.py
 9 FAILED tests/api/test_rankings_our_rank.py
 5 FAILED tests/api/test_source_registry_parity.py
 5 FAILED tests/api/test_draft_capital.py
 4 FAILED tests/api/test_dynasty_nerds_integration.py
 4 FAILED tests/api/test_config_parity.py
 3 FAILED tests/api/test_sleeper_overlay.py
 2 FAILED tests/api/test_cache_control_privacy.py
 1 each  test_watchdog_freshness_soft / test_source_floor_invariant / test_playerctx_endpoint
```

**Cause.** This branch was carrying `claude/league-intel-foundation` at `575b11aa`, which already includes ADR-002's rewrite of `config/leagues/registry.json` — `rosterSize 30→58`, `taxiSize 5→0`, `TE 1→2`, `DL/LB/DB 2→3`, `IDP_FLEX 2→0`, `K` added — **without** the four later LI-8 commits that bring the consumers and their tests into line with it. A half-applied registry correction is exactly the shape that lights up every `tests/api` test asserting starter counts or roster size.

Merging `claude/league-intel-foundation` to its current head (`c8cd5179`) closes all 87. **No test was skipped, xfailed, or edited to get there** — the count moves from 1900 passed to 2207 passed, and the 160 skips drop back to 3.

### Other checks

- `python -m ruff check src/roster_intel src/league_intel/cross_market.py` — **All checks passed** on ruff 0.6.9 (the CI pin), invoked as `python -m ruff` per the CI convention.
- Redesign R3 (#551) deleted `frontend/components/terminal/Panel.jsx`; `grep -rn "terminal/Panel" frontend/` returns **no matches** on this branch. No stale import survives the deletion.
- `.github/workflows/grant-ssh-access.yml` is **gone** from this branch as of the `main` merge. It was present before, `main` had deleted it in `c534280d`, and `docs/ORCHESTRATION.md` requires its deletion — merging this branch beforehand would have resurrected a dispatch-only path to a production shell.

### One thing I could not close, flagged rather than buried

`928a1fa8` makes `allow_scalar_fallback` real, and `generate_packages` defaults it to `False`. Measured directly:

```
allow_scalar_fallback=True   ->  total = 6605.5  (market: idpTradeCalc)
allow_scalar_fallback=False  ->  total = None    (unvaluable)
```

`tests/roster_intel/test_packages.py` is **47/47 green both before and after** that merge. The behaviour genuinely changed and no test observed it — so the coverage gap is real, and it is the thing I would want closed before this ships.

---

jasonleetucker-code:
## Rebased onto the new #562, and an ADR number collision caught

Same treatment as #562: the old `claude/league-intel-foundation` commits were squash-merge duplicates, so instead of merging `main` in, I **cherry-picked this branch's 8 own commits** onto the rebased #562. Everything #550 contributed is now inherited from `main` rather than carried.

### Verification

```
git diff origin/main HEAD -- src/league_intel/ src/ros/ src/api/ config/
  → 0 bytes

deletions in docs/ vs origin/main  → 0
```

`src/league_intel/cross_market.py` and `sim.py` are byte-identical to `main`. The file list is 19 `A` and exactly one `M`:

```
M  docs/league-intelligence/DECISIONS.md        (ADR-013 added, nothing removed)
A  src/roster_intel/{__init__,engine,marginal,packages,partner,
                     profiles,roster_source,targets,window}.py
A  tests/roster_intel/{__init__,test_engine,test_marginal,test_packages,
                       test_partner,test_profiles,test_real_rosters,
                       test_roster_source,test_targets,test_window}.py
```

### The ADR was renumbered — `main` already used ADR-012

`main` has **ADR-012: the LI-8 collision with `claude/league-intel-sim`, and who owns which half**. My horizon-weight ADR was also numbered 012, so cherry-picking it would have produced two ADR-012s, and a careless conflict resolution could have overwritten the LI-8 one.

Mine is now **ADR-013**, and `DECISIONS.md` was rebuilt against main's current file rather than cherry-picked — so **ADR-011 and the third revision of ADR-010 are preserved**. That file is a genuine union, exactly as you called it. Verified: zero deleted lines in `docs/`.

### The tier fix propagates here

The `classify_tier` defect lives in `profiles.py`, which belongs to #562, so it is fixed there and inherited. Full detail and the pre-fix failure output are in [#562's comment](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/562#issuecomment-5085350157). Short version: `elite` was pinned to the startable floor, the `starter` tier was unreachable, 77% of the league classified elite; it is now a top-5% positional percentile reusing `replacement.py`'s own elite definition, reachable by construction.

### Results on this branch's head

```
tests/roster_intel + tests/league_intel   633 passed, 22 skipped
python -m ruff check src/roster_intel tests/roster_intel   All checks passed  (0.6.9)
```

### Still open, unchanged by the rebase

`928a1fa8` is now inherited from `main`, and you confirmed there are **zero production importers** of `cross_market` on this branch — but `src/roster_intel/packages.py` **does** call `value_package(..., allow_scalar_fallback=...)` and `generate_packages` defaults it to `False`. Measured:

```
allow_scalar_fallback=True   ->  total = 6605.5  (idpTradeCalc)
allow_scalar_fallback=False  ->  total = None    (unvaluable)
```

`tests/roster_intel/test_packages.py` is green both before and after, so the behaviour change is real and **uncovered**. It has no reachable surface in production today (nothing imports `roster_intel`), which is a reason to close it properly rather than quickly — the same argument you made for the tier bug.

---

jasonleetucker-code:
## Rebased onto #562's new head, and both gates run for real

#562 moved twice (the tier fix, then `76db5d80` formatting it). This branch is rebased onto `76db5d80` at `897b4950`, so it now inherits the corrected `classify_tier` and the formatting.

Gate commands run against the exact tree being pushed, not adjacent ones — the distinction that made me wrongly report "ruff clean" on #562 when only `ruff check` had passed and `ruff format --check` was red:

```
python -m ruff format --check .                            520 files already formatted
python -m ruff check src/roster_intel tests/roster_intel   All checks passed
pytest tests/roster_intel tests/league_intel               633 passed, 22 skipped

deletions in src/league_intel|src/ros|src/api|config       0 bytes
deletions in docs/                                          0
non-ADD paths vs main                                       docs/league-intelligence/DECISIONS.md   (only)
```

## ADR-013, and the numbering

**`main` already uses ADR-012** — "the LI-8 collision with `claude/league-intel-sim`, and who owns which half". My horizon-weight ADR was also drafted as 012, so cherry-picking it would have produced two ADR-012s, and whoever resolved that conflict would have had to overwrite one of them. **The numbering moved to ADR-013.**

Verified in the pushed tree rather than asserted:

```
$ git show HEAD:docs/league-intelligence/DECISIONS.md | grep '^## ADR-01[123]'
 41: ## ADR-013: the competitive window enters per POSITION as a horizon weight,
     not as a global win-now scalar (WS-J)
130: ## ADR-011: the trade primitive is a weekly scoring DISTRIBUTION, and it runs
     on the full roster (LI-8)
210: ## ADR-012: the LI-8 collision with claude/league-intel-sim, and who owns which half
```

`DECISIONS.md` was **rebuilt against main's current file rather than cherry-picked**, so ADR-011 and the third revision of ADR-010 survive intact — zero deleted lines in `docs/`. It is a union, not a take-mine.

ADR-013 lives here rather than on #562 because it documents `horizon_weight` in `src/roster_intel/targets.py`, and that file exists only on this branch. #562 contains no `docs/` change at all.

What it records: `rank_target_positions` multiplied every position by a single global scalar, so a uniform positive multiplier could not reorder anything — a rebuilding roster and a championship contender got the **identical** ranked order of target positions while the docstring claimed the window affected the recommendation. The replacement keys the five-state distribution to the mean age of each position's top-3 obtainable candidates. It also records why the failure mode is `None` rather than `1.0`: `None` means the term did not apply, `winNowWeight` serializes as `null`, and stamping the old scalar there would have reproduced the original defect one layer down.

---

jasonleetucker-code:
## Rebased onto `main` @ `4f9cb05b`, and the uncovered flag is now covered

Head `de89b6e5`. Nine commits, rebased **directly onto `origin/main`** — never onto a merged branch. #562/#564/#565 are all content-without-ancestry now, so their commits were dropped from this branch rather than re-applied.

### Revert check

```
deletions in src/league_intel|src/ros|src/api|config|docs   0

#565's additions, verified individually:
  src/league_intel/twin.py   present, byte-identical to main
  src/ros/lineup.py          byte-identical to main
  fantasy_positions refs     main 7 / HEAD 7
```

Total diff is `+5,082 / −4`, and **all four deletions are in `src/roster_intel/__init__.py`** — the four docstring lines this branch rewrites to name `packages`/`targets`/`partner`. Nothing shared loses a line.

A **real** `git merge` into a scratch worktree at `origin/main`: **0 conflicts**, 8 files.

### Gates, run literally against the pushed tree

```
python -m ruff --version                    ruff 0.6.9
python -m ruff format --check .             535 files already formatted
pytest tests/roster_intel tests/league_intel   705 passed, 22 skipped
```

On lint I replicated CI's **actual blocking step** rather than the repo-wide command, because they differ and the difference matters here. `pr-validation.yml:77` lints changed files only:

```
git diff --name-only --diff-filter=ACMR <base> HEAD -- '*.py' | xargs python -m ruff check
  → All checks passed!
```

Repo-wide `ruff check .` reports **51 errors — but that is the pre-existing backlog, identical on pristine `main` (51), and none of them are in `roster_intel`.** That is the report-only mirror in `deploy.yml`, not a gate. Flagging it so the number isn't mistaken for a regression.

The 705 matters for your reason: #565 changed `lineup.py`'s eligibility handling and `roster_intel` consumes the solver, so this is run against the new base, not a stale green.

### The uncovered behaviour change — closed

`test_packages.py` goes 47 → 51. Reaching the scalar path from `generate_packages` was harder than expected and the reason is worth recording: staging only builds multi-asset sides out of **near-misses**, so a mixed *send* is unreachable — the 1-for-1 seeds are either accepted (never expanded, the simpler package dominates) or structurally rejected (never expanded either). The reachable shape is a mixed **receive**: one KTC-only asset of ours for two of theirs, one KTC-only and one IDPTC-only. Added `_ktc_only_asset` as the offense mirror of the existing `_idp_asset`.

Measured through `generate_packages`:

```
allow_scalar_fallback=False   rejected "unvaluable", frontier empty
allow_scalar_fallback=True    priced, 1 member on the frontier
```

Run against a **simulated pre-`928a1fa8`** `cross_market.py` (the `if not allow_scalar_fallback` guard stubbed to `if False`):

```
2 failed, 2 passed
  FAILED test_mixed_market_package_is_unvaluable_by_default
  FAILED test_the_flag_changes_the_frontier
         AssertionError: allow_scalar_fallback made no difference to the frontier
         assert 1 != 1
```

`cross_market.py` restored byte-identical to main afterwards (verified by hash).

The 2 that passed are the useful part: the signature-default check is independent of the fix, and opting *in* behaves the same either way. Only the strict-default and the difference assertion catch it — the same lesson as the tier defect, where ordering passed and only reachability caught the bug. **The property that catches a defect is the one asserting the other branch is reachable.**

### ADR renumbered a second time — now ADR-014

`main` took **ADR-013** for the League Twin bridge (#565), having already taken ADR-012 for the LI-8 ownership split. Mine is now **ADR-014**, and `DECISIONS.md` was rebuilt against main's current file rather than conflict-resolved. Verified ADR-001 through ADR-013 all still present, zero deleted lines in `docs/`.

Worth naming as a process smell rather than just fixing again: this is the second collision, and the next agent to write an ADR off a stale base hits it too. The number is being chosen at authoring time against a base that moves before merge.

### Untouched, deliberately

`src/trade/angle.py` — still the live, user-reachable call site summing market values across a combo never constrained to one market, on both the offer side (`:378-388`, `:471-478`, `:526`) and the acquire side (`:913`). It needs its own change with before/after evidence, not a ride on this one.

---

PR #570, opened 2026-07-26
Domain cutover prep (A): chaseupside.com nginx config + runbook

jasonleetucker-code:
Repo-side preparation for putting production behind `https://chaseupside.com`. Today the site is plain HTTP with **no TLS**, so login credentials cross the network in the clear.

**This PR is safe to merge (and therefore deploy) at any time** — nothing in it changes where a running process points. The runtime origin changes are split into **#571**, which must not merge until the certificate exists.

> **Rewritten since first review.** The original was modelled on `deploy/nginx/riskittogetthebrisket.org.conf`, which is *not* what production runs. It is now reconciled against the actual live `sites-enabled/dynasty`. The branch was force-pushed to a single clean commit.

---

## Reconciliation against the live config

Every live setting is preserved. Each deliberate difference is annotated at the point where it occurs in the file, and listed here.

### Preserved from live (would have been defects to drop)

| Live setting | Status |
|---|---|
| `client_max_body_size 25m` | **kept at 25m.** The old repo copy said 10m. Dropping the directive would not have fallen back to 10m — it would have fallen back to nginx's built-in **1 MB** default and started rejecting uploads that work today. Good catch. |
| `proxy_read_timeout 300s` on `/api/` | **kept at 300s.** The repo copy said 120s. Live had already been raised, and that operational experience wins — a scrape cycle can hold `/api/data` for minutes. |
| `upstream dynasty_backend` / `dynasty_frontend` | **names kept identical** — see the note on double-enable below. |
| `server_name 169.58.50.224` | **kept**, in a dedicated transitional block — see below. |
| `location /api/`, `/_next/`, `= /favicon.ico`, `/` | all present |
| `X-Real-IP`, `X-Forwarded-For`, `X-Forwarded-Proto`, `Host`, `Upgrade`, `proxy_http_version 1.1` | all present |

### Intentional differences

| Difference | Why |
|---|---|
| **`/_next/static/` block added** (not live) | nginx-level caching of content-hashed immutable assets. A superset of live, not parity — calling it out rather than letting it pass as a port. Privacy scope is documented in the snippet: `riskit_static` is the only shared cache and it is attached only to this location, never `/api/*`. |
| **`/_next/` gains headers** (live sets only `Host`) | Checked for caching/upstream impact: none. This location is uncached, and nginx's default cache key is `$scheme$proxy_host$request_uri` — request headers are not part of it. `X-Forwarded-Proto` is the one that matters and is *required* post-TLS: without it a Next.js process behind TLS termination can emit `http://` absolute URLs. |
| **`Connection $connection_upgrade` on `/`** | Live hardcodes `Connection "upgrade"` on every request through `/`, websocket or not. The map sends `upgrade` only when the client asked for one. No-op for real websockets, correct for everything else. |
| **`keepalive` on both upstreams** | Not live. Idle-connection pooling; only used by locations that send `Connection ""`. |
| **Security headers + gzip** | None exist live. HSTS is sent **only** from the `:443` block — never from the plain-HTTP bare-IP block, where browsers ignore it anyway. |

### Structure

Three server blocks, because the bare IP and the domain need different treatment:

1. **`:80` chaseupside.com / www** — ACME challenge, then 301 to HTTPS.
2. **`:443` chaseupside.com / www** — the real site, TLS + HSTS.
3. **`:80` 169.58.50.224** — transitional, plain HTTP, **no redirect**. Redirecting the IP to `https://` would break every monitoring path pointing at it, because a certificate for `chaseupside.com` cannot validate for an IP address — probes would fail the TLS handshake outright, not merely warn. Carries a written removal criterion.

The location blocks live in **`deploy/nginx/chaseupside-proxy.conf`**, included at server level by blocks 2 and 3. That is the point: the IP path and the domain path must serve an identical surface during the cutover, and a location added to one and forgotten in the other is precisely the failure mode here. Now impossible by construction.

### Double-enable fails loudly, on purpose

This file **replaces** the `dynasty` site; it does not sit alongside it. Both declare `upstream dynasty_backend` / `dynasty_frontend`, and — since the operator widened the live `server_name` — both claim `chaseupside.com`. Enabling both fails `nginx -t` with a duplicate-upstream error.

I kept the upstream names identical to live *specifically to get that*. Renaming them would have let both files load, leaving two server blocks silently fighting over the same hostname with resolution depending on config load order. A hard `nginx -t` failure is the better outcome.

### Correction on the `--nginx` reasoning

You are right that I overstated it. `apply_hardening.sh` is invoked by **no workflow**, and `deploy.sh` does not touch nginx at all — I have re-verified both. The revert risk is **manual-only**: it bites if someone runs `apply_hardening.sh` by hand. Webroot is still the right choice because it never touches nginx config, so the checked-in copy stays authoritative — but the config header and the runbook now state the risk accurately rather than implying an automated hazard.

---

## Runbook rewritten against the real starting state

The original assumed a fresh box needing a minimal bootstrap config. It now starts from what is actually true:

- `sites-enabled/dynasty` is a **symlink**, with `server_name` already widened by hand to include the domain, so `http://chaseupside.com/` already serves the app.
- Therefore **no bootstrap config is needed**. Instead the runbook inserts an ACME `location` into the live file with a single anchored `sed`, backed up to `dynasty.bak-acme`. I tested that `sed` against a copy of your live config: anchor count 1, correct insertion point ahead of `location /api/`, backup created.
- A separate `server_name chaseupside.com` block was **not** an option — the live block already claims that name, so it had to be an in-place edit.

Two further corrections found by re-reading it as an operator would execute it:

- **`deploy.yml` fires on `push` to `main`, so merging is deploying.** The old step 1 said "merge but do not deploy", which was simply wrong. Merging PR A is now *how the config files reach the box* (step 2), and PR B's merge is the final deploy (step 11).
- **Reachability checks moved to `[LOCAL]`.** Curling the public hostname from the box resolves to its own IP and can pass via loopback/NAT hairpin while the public path is broken — which is the only thing Let's Encrypt and real users care about.

There is also a new **step 0b**, which is urgent independently of this cutover: `intel-refresh.yml` runs on a **cron** and sends `Authorization: Bearer ${INTEL_REFRESH_TOKEN}` to whatever `PROD_PUBLIC_URL` names. If that variable still holds the lapsed `riskittogetthebrisket.org`, the token is being handed to a third party on every scheduled run. The runbook says to read it and clear it immediately if it names a host we do not control, rather than waiting for step 9.

---

## Classification rule

> A hit is **changed** only if a machine reads it, or a human following it *today* would be sent to the wrong host. Everything else stays byte-for-byte.

Three "leave alone" buckets:

1. **Repo identity — correct as-is, not an oversight.** The repo keeps the name `riskittogetthebrisket`. Clone URLs, the Windows working-copy path, CI checkout paths, and `PROD_APP_NAME`'s default are **right**. A future reader sees "Chase Upside" as the product and the old name on the repo; that is deliberate.
2. **Historical record — leave verbatim.** Anything self-declaring `⚠ STALE` or carrying a dated header (`_Generated:`, `_Updated:`, `_Prepared:`). With the product also being renamed these are now *doubly* historical. **Do not "fix" them later.**
3. **Captured artifacts.** `audit/baseline/health.json`, `data/**`, the checked-in PDF.

Judgement call worth surfacing: `docs/ORCHESTRATION.md` says "`riskittogetthebrisket.org` lapsed and now resolves to a third party." Still true, still a useful warning — stays.

### Counts (across both PRs)

**46** files match `riskittogetthebrisket` in any form (excluding `data/**` and the PDF).

| Bucket | Files | Action |
|---|---:|---|
| Repo name / path only — never a hostname | 14 | untouched |
| Used as a **hostname** | **32** | classified |
| → live configuration | **20** | changed (12 here, 8 in #571) |
| → historical record + captured artifacts | 12 | untouched |

The 12 left alone: `HANDOFF.md`, `audit/baseline/health.json`, `docs/CLAUDE_SESSION_AUDIT_HANDOFF.md`, `docs/ORCHESTRATION.md`, `docs/ops/automation-runbook.md`, `docs/ops/recommended-automation-model.md`, `docs/runbooks/ktc-production-validation.md`, `docs/runbooks/production-activation-runbook.md`, `docs/runbooks/public-primary-activation.md`, `docs/status/CURRENT_STATE.md`, `docs/status/master-implementation-audit.md`, `docs/status/workstream-inventory.md`.

No blanket find-and-replace: every edit was applied with an expected-occurrence count that had to match before the write.

---

## Follow-on changes (each verified in code)

1. **`PROD_PUBLIC_URL`** → runbook steps 0b and 9. **Six** workflows read it, not three: `deploy.yml`, `health-check.yml`, `intel-refresh.yml`, `prod-e2e-smoke.yml`, `public-league-warmup.yml`, `smoke-test.yml`. Exactly **one** sends a credential — `intel-refresh.yml`.
2. **Cookies** → runbook 10a. `JASON_AUTH_COOKIE_SECURE` flips to `true` after HTTPS is verified. Verified as needing no change: **no `domain=` attribute** on any `set_cookie` (host-only, correct across a domain change) and `SameSite=lax` is right for a same-origin app.
3. **CORS** — searched `CORSMiddleware`, `allow_origins`, `Access-Control-Allow-*`, `TrustedHostMiddleware`: **zero hits repo-wide.** Strictly same-origin. Recorded so nobody hunts for an allowlist that never existed.
4. **Monitoring** → #571, plus the manual Grafana re-import noted there.
5. **Frontend origins** → #571. Build-time subtlety verified: `deploy.sh` runs `npm run build` **without sourcing `.env`**, so `NEXT_PUBLIC_SITE_URL` would inline as `undefined` — use `PUBLIC_SITE_URL`, which the frontend unit supplies at runtime via `EnvironmentFile=`.
6. **Gap flagged, not fixed:** no `metadataBase` anywhere, so relative OG image URLs resolve against `localhost:3000`. It lands in the same files as the rebrand copy — better done once, with the naming decision in hand.
7. **Latent, not fixed:** the three alert modules hardcode the public origin instead of reading one config value. Out of scope for a cutover.

---

## Product naming — inventory only, nothing renamed

Product becomes **"Chase Upside"**. Nothing in this diff renames anything.

**The trap:** the Sleeper league is *also* named "Risk It To Get The Brisket" and is **not** being renamed. A blanket replace breaks working code — `scripts/fetch_draftsharks.py:462` matches that string against **DraftSharks' HTML**, so changing it silently stops the fetch recognising the page. Same for `LEAGUE_NAME`, `config/leagues/registry.json`'s `displayName`, `src/api/chat.py`'s system prompt, and `frontend/app/league/page.jsx:52` (a fallback for the Sleeper name, not a product string). These are **league identity** and must survive the rebrand.

**Group A — user-visible** (cheap; wrong-looking if skipped): `manifest.js` (`name`, `short_name`), `layout.jsx` (`appleWebApp.title`), `page.jsx:21` `<h1>`, `TopBar.jsx:51` nav brand, `nav-model.js:212` fallback, 6 × `openGraph.siteName`, 4 × `opengraph-image.js` footers, `design/page.jsx:12`, `ScreenshotFab.jsx` share title + filename, `rankings/page.jsx:625` CSV filename, `sw.js:207` push title, and backend copy — `[Brisket Ops]` subjects in three modules, `[Brisket]` + `— Risk It` in `signal_alerts.py`, `server.py:4032,8449,8460`.

The **login screen carries no product name** ("Account" / "Sign in" only) — branding it is net-new copy. The root `<title>` is `"Dynasty Trade Calculator"` with no brand at all, so that one is arguably new naming rather than a swap. Several slots are `<old name> — <suffix>`; the suffix is an independent editorial choice.

**Group B — infra identifiers.** `dynasty.service` / `dynasty-frontend.service`, the `riskit-*` timers and their log/state paths, the `riskit_static` cache zone, `/home/dynasty/trade-calculator`, the `dynasty` app user, `PROD_APP_NAME`. **Recommendation: leave them.** These names are already inconsistent with each other (`dynasty`, `riskit`, `brisket`, `trade-calculator` all appear), so renaming buys consistency with the product while preserving none internally — and `deploy.yml`'s `PROD_*` variables mean a later rename is a settings change, not a code change. Nothing is locked in by waiting.

**Group C — repo-internal prose.** Free to change; much of `docs/` is in the historical-record bucket and should stay.

---

## Verification

- **`nginx -t` equivalent:** no nginx in this environment, so I validated with **`crossplane`**, nginx's official config parser — clean parse, zero syntax or context errors, including the snippet **inlined into its real server-level context**. Three negative controls (missing semicolon, unbalanced brace, wrong-context directive) were all correctly detected, so the clean result has signal. A real `nginx -t` on the box remains a required gate at runbook step 6.
- The step-3 `sed` was executed against a copy of the live config: anchor matched exactly once, block inserted in the right place, backup written.
- `python -m ruff format --check .` → `538 files already formatted` (ruff 0.6.9, matching CI).
- Full suite on the combined A+B tree: `4227 passed, 25 skipped, 366 subtests passed`.
- Rebased on `origin/main` (`ab98871`); 15 files, all mine.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -54,7 +54,7 @@ LOG_FORMAT=text
 
 # Uptime watchdog configuration
 # UPTIME_CHECK_ENABLED=true
-# UPTIME_CHECK_URL=https://riskittogetthebrisket.org/api/health
+# UPTIME_CHECK_URL=https://chaseupside.com/api/health
 # UPTIME_CHECK_INTERVAL_SEC=300
 # UPTIME_CHECK_TIMEOUT_SEC=5
 # UPTIME_ALERT_FAIL_THRESHOLD=2
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Risk It To Get The Brisket — Dynasty Trade Calculator
 
 Private repo for the dynasty trade calculator stack powering
-[riskittogetthebrisket.org](https://riskittogetthebrisket.org).
+[chaseupside.com](https://chaseupside.com).
 
 **Architecture at a glance:**
 - **Backend:** Python 3.12 FastAPI + Uvicorn (port 8000).
--- a/deploy/PRODUCTION_BOOTSTRAP.md
+++ b/deploy/PRODUCTION_BOOTSTRAP.md
@@ -8,7 +8,7 @@ Current production target:
 - app path: `/home/dynasty/trade-calculator`
 - venv path: `/home/dynasty/.venvs/trade-calculator`
 - service: `dynasty`
-- domain: `riskittogetthebrisket.org`
+- domain: `chaseupside.com`
 
 ## 1) Repo-managed bootstrap (safe to rerun)
 
@@ -72,18 +72,18 @@ sudo /home/dynasty/.venvs/trade-calculator/bin/python -m playwright install-deps
 
 ## 3) Reverse proxy and TLS
 
-The nginx site config is maintained in `deploy/nginx/riskittogetthebrisket.org.conf`.
+The nginx site config is maintained in `deploy/nginx/chaseupside.com.conf`.
 
 Install or update:
 
 ```bash
-sudo cp deploy/nginx/riskittogetthebrisket.org.conf /etc/nginx/sites-available/riskittogetthebrisket.org
-sudo ln -sf /etc/nginx/sites-available/riskittogetthebrisket.org /etc/nginx/sites-enabled/
+sudo cp deploy/nginx/chaseupside.com.conf /etc/nginx/sites-available/chaseupside.com
+sudo ln -sf /etc/nginx/sites-available/chaseupside.com /etc/nginx/sites-enabled/
 sudo nginx -t && sudo systemctl reload nginx
 ```
 
 Routing:
-1. `riskittogetthebrisket.org` terminates TLS at nginx.
+1. `chaseupside.com` terminates TLS at nginx.
 2. HTTP (`:80`) redirects to HTTPS (`:443`).
 3. `/api/*` → Python backend (`127.0.0.1:8000`).
 4. `/_next/*` → Next.js frontend (`127.0.0.1:3000`) — static assets.
@@ -97,7 +97,7 @@ From server:
 ```bash
 sudo -n /bin/systemctl is-active dynasty
 curl -fsS http://127.0.0.1:8000/api/status | head -c 400
-curl -fsS https://riskittogetthebrisket.org/api/health | head -c 400
+curl -fsS https://chaseupside.com/api/health | head -c 400
 ```
 
 From GitHub:
--- a/deploy/grafana/README.md
+++ b/deploy/grafana/README.md
@@ -3,7 +3,7 @@
 ## `public-league-dashboard.json`
 
 One-panel-per-row dashboard that polls
-`https://riskittogetthebrisket.org/api/public/league/metrics` every 30s
+`https://chaseupside.com/api/public/league/metrics` every 30s
 and surfaces the key snapshot-cache signals:
 
 | Panel                          | Target                                   |
--- a/deploy/monitoring/README.md
+++ b/deploy/monitoring/README.md
@@ -5,9 +5,9 @@ probe is log-only until the operator opts in to a notification channel.
 
 ## What it checks (every 5 minutes)
 
-1. `https://riskittogetthebrisket.org/api/health` — full external path:
+1. `https://chaseupside.com/api/health` — full external path:
    DNS → TLS → nginx → FastAPI backend.
-2. `https://riskittogetthebrisket.org/` — nginx → Next.js frontend.
+2. `https://chaseupside.com/` — nginx → Next.js frontend.
 3. `http://127.0.0.1:8000/api/health` — backend direct (distinguishes
    "backend down" from "nginx/TLS broken"); disable with
    `CHECK_LOCAL=false`.
--- a/deploy/nginx/chaseupside-proxy.conf
+++ b/deploy/nginx/chaseupside-proxy.conf
@@ -1,0 +1,156 @@
+# chaseupside.com — shared proxy/location snippet
+#
+# Included at SERVER level by both serving blocks in
+# deploy/nginx/chaseupside.com.conf:
+#
+#   * the :443 chaseupside.com / www block (the real site), and
+#   * the transitional :80 169.58.50.224 block (kept until every
+#     monitoring path is repointed — see that block's removal note).
+#
+# It exists so those two cannot drift.  The bare IP and the domain must
+# serve the identical application surface during the cutover window; a
+# location block added to one and forgotten in the other is exactly the
+# failure this file prevents.
+#
+# Install to /etc/nginx/snippets/chaseupside-proxy.conf.
+#
+# ⚠ Do NOT put `add_header` in here.  nginx's add_header inheritance is
+# all-or-nothing per context: a location-level add_header silently drops
+# every server-level security header for that location.  Security
+# headers are declared once per server block in chaseupside.com.conf,
+# which also lets the :443 block send HSTS while the plain-HTTP IP block
+# correctly does not.
+
+# ── Upload ceiling ───────────────────────────────────────────────────
+# 25m, matching the live production config.  The repo's older
+# riskittogetthebrisket.org.conf said 10m; live had been raised to 25m
+# and the live value wins.  Omitting the directive entirely would NOT
+# fall back to 10m — it would fall back to nginx's built-in 1m default
+# and start rejecting uploads that work today.
+client_max_body_size 25m;
+
+# ── Compression (fallback layer) ─────────────────────────────────────
+# The FastAPI backend already gzips responses >= 1 KB (GZipMiddleware)
+# and pre-gzips the /api/data contract.  nginx never re-compresses
+# responses that arrive with Content-Encoding set, so this only picks up
+# what the upstreams leave uncompressed (Next.js pages when its built-in
+# compression is off, JSON error bodies, svg icons).
+gzip on;
+gzip_comp_level 5;
+gzip_min_length 1024;
+gzip_proxied any;
+gzip_vary on;
+gzip_types
+    application/json
+    application/javascript
+    application/manifest+json
+    application/xml
+    image/svg+xml
+    text/css
+    text/plain
+    text/xml;
+
+# ── API routes → Python backend ──────────────────────────────────────
+location /api/ {
+    proxy_pass http://dynasty_backend;
+    proxy_http_version 1.1;
+    proxy_set_header Host              $host;
+    proxy_set_header X-Real-IP         $remote_addr;
+    proxy_set_header X-Forwarded-For   $proxy_add_x_forwarded_for;
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header Connection        "";
+    # 300s, matching live.  A scrape cycle rebuilds the in-memory cache
+    # and can hold /api/data for minutes; live had already been raised
+    # from the repo copy's 120s and that experience is authoritative.
+    proxy_read_timeout 300s;
+    # Fail fast when the backend process is down instead of hanging a
+    # browser tab for the default 60s.  (Not present live; nginx's
+    # default is 60s, so this only shortens the dead-backend case.)
+    proxy_connect_timeout 5s;
+    proxy_send_timeout 30s;
+    # The /api/data contract is MB-scale even after upstream gzip
+    # (~100-400 KB on the wire, up to ~4 MB if a client requests
+    # identity encoding).  Larger in-memory buffers keep nginx from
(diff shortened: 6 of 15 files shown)


jasonleetucker-code:
## Test suite result

```
4227 passed, 25 skipped, 1 warning, 366 subtests passed in 1146.51s (0:19:06)
```

`python -m pytest tests/ -q` → exit 0. (The one warning is a pre-existing `StarletteDeprecationWarning` from `fastapi.testclient`, unrelated to this branch.)

## Four follow-up commits after the initial push

Re-reading the runbook as an operator would execute it turned up three ordering/correctness bugs and one missing rollback step:

1. **`56ab7c2`** — Step 5 told the operator to `git fetch` and then copy the config out of the working tree, but the working tree is still on the *pre-merge* commit at that point (the deploy is step 10), so the file would not have been there. Now reads it straight out of `FETCH_HEAD` with `git show`, which leaves the checkout — and `deploy.sh`'s idea of what is deployed — untouched.

2. **`39f5c4c`** — The ACME-challenge check (step 3) and the redirect/cert verification (step 6) were tagged `[VPS]`. Curling the public hostname *from the box* resolves to its own IP and can pass via loopback or NAT hairpin while the path is unreachable from outside — which is the only thing Let's Encrypt and real users care about. Both are now `[LOCAL]`, and the tag legend says why.

3. **`4287539`** — Rollback (11a) puts the site back on plain HTTP. If step 9a had already set `JASON_AUTH_COOKIE_SECURE=true`, browsers then silently discard the session cookie and nobody can log in. The symptom — bounced to the login screen, no error — looks nothing like an nginx rollback, so the revert is now an explicit step in 11a.

Final state: **22 files**, `git diff --stat origin/main HEAD` clean of anything not mine.


---

jasonleetucker-code:
Reconciled against the live config. Branch force-pushed to a single clean commit and rebased onto current `origin/main` (`c9371ab`); the two commits that landed meanwhile were automated `data/` + `exports/` refreshes, and this branch touches neither, so nothing merged is reverted. Full body above is rewritten — the three asks, answered directly:

### 1. Config reconciled

Each of your five points:

- **`client_max_body_size 25m`** — kept. You were right to call it a defect: dropping the directive would not have fallen back to the repo's 10m, it would have fallen back to nginx's built-in **1 MB** default.
- **`/_next/static/`** — kept as an addition, now labelled as a superset rather than parity, with its privacy scope written down.
- **`/_next/` extra headers** — checked as asked. No caching impact: that location is uncached, and nginx's default cache key is `$scheme$proxy_host$request_uri`, which does not include request headers. `X-Forwarded-Proto` is the one that matters and is *required* post-TLS, or Next.js can emit `http://` absolute URLs from behind the terminator.
- **Upstreams** — the new file **replaces** `dynasty` rather than joining it, and I deliberately kept the names `dynasty_backend` / `dynasty_frontend` identical so a double-enable fails `nginx -t` loudly. Renaming them would have let both files load and left two server blocks silently fighting over the same hostname.
- **Bare IP** — kept, in its own transitional `:80` block that does **not** redirect. A `chaseupside.com` certificate cannot validate for an IP, so redirecting it would make every monitoring probe fail the TLS handshake outright.

Two more the live config surfaced that were not on your list: **`proxy_read_timeout 300s`** (live had been raised from the repo's 120s — live wins), and live's hardcoded `Connection "upgrade"` on `/`, which the map version fixes.

The location blocks now live in a shared snippet included by both the `:443` and bare-IP blocks, so the two surfaces cannot drift.

### 2. Runbook rewritten from the real starting state

No bootstrap config — the live file already answers on the domain, so the runbook inserts an ACME `location` into it with one anchored `sed`. **I ran that `sed` against a copy of the config you pasted**: anchor matched exactly once, block landed ahead of `location /api/`, backup written to `dynasty.bak-acme`. A separate `server_name chaseupside.com` block was not an option — the live block already claims that name.

Two more corrections from re-reading it as an operator would execute it:

- **`deploy.yml` fires on `push` to `main` — merging *is* deploying.** My old "merge but do not deploy" step was wrong. Merging A is now how the files reach the box.
- Reachability checks moved to `[LOCAL]`; curling the hostname from the box can pass via loopback/hairpin while the public path is broken.

New **step 0b**, urgent independently of the cutover: `intel-refresh.yml` runs on a **cron** and sends the bearer token to whatever `PROD_PUBLIC_URL` names. If it still holds the lapsed domain, that token is going to a third party on every scheduled run — read and clear it now, don't wait for step 9.

### 3. Split — and it turned out not to be a judgement call

`deploy.yml` triggering on `push: main` settles it: merging the origin changes **is** deploying them. Gating in the runbook would not have helped.

**#571** carries the 8 machine-read files, stacked on this branch. Split rule: *machine-read at runtime → B, human-read → A*. That keeps whole files on one side, so there is nothing to reconcile between the two PRs. Prose that mentions the new origin stayed here — a stale sentence in a doc costs a moment's confusion, a stale value in the uptime probe costs a pager.

### Validation

No nginx in this environment, so I used **`crossplane`** (nginx's official parser): clean parse, zero syntax or context errors, including the snippet **inlined into its real server-level context**. I ran three negative controls first — missing semicolon, unbalanced brace, wrong-context directive — all correctly detected, so the clean result actually has signal. A real `nginx -t` on the box is still a required gate at step 6.


---

PR #579, opened 2026-07-27
test(e2e): make the green check mean something — fix preflight, kill vacuous assertions, resolve every skip gate

jasonleetucker-code:
This is the rewrite batch #566 deferred — that audit closed with *"the five vacuous chrome-matching assertions, the `section`-locator snapshot target, and the remaining skip gates are all real, but they are a batch of assertion rewrites and deserve their own review."* This is that review.

Full finding-by-finding record in **`docs/e2e-assertion-audit.md` Part 2**.

## The suite could not run from a clean checkout

`npm run e2e` — the README's *"that is the whole recipe"* — aborted before a single spec executed. `preflight.py::main()` validated the API contract **before** seeding the snapshot that validation reads; `data/` is gitignored, so on a clean checkout there is never one.

Verified by watching it fail, not by reasoning. Emptied `data/`, restored `origin/main`'s ordering:

```
[preflight] Python compile checks passed
FileNotFoundError: No dynasty_data_YYYY-MM-DD.json files found in repo/data or repo root.
EXIT=1
```

Re-applied the fix against the same empty `data/`:

```
[preflight] seeded data/dynasty_data_2026-07-26.json from committed exports/latest/ (1077 players)
[contract] ok=True errors=0 warnings=2 players=1095
[preflight] ready for browser regression suite
EXIT=0
```

It only ever looked fine on machines carrying a snapshot from an earlier local scrape. The README's *"Don't trust these signals"* table now names this traceback explicitly as **not** expected — that table was the one document written to prevent this misdiagnosis, and it was pointing readers away from a real bug.

## Vacuous assertions — replaced, and proven

An assertion is vacuous if it cannot distinguish *"the page I named rendered"* from *"some other page rendered."* Each original and its replacement were run against a **decoy** page (`/settings`) carrying the same shell chrome:

```
  ✓  1 › OLD /rosters assertion passes on the DECOY page (=> vacuous)
  ✘  2 › NEW /rosters assertion FAILS on the DECOY page (=> real)
  ✓  3 › OLD /trade assertion passes on the DECOY page (=> vacuous)
  ✘  4 › NEW /trade assertion FAILS on the DECOY page (=> real)
  ✓  5 › OLD home assertion passes on the DECOY page (=> vacuous)
  ✘  6 › NEW home assertion FAILS on the DECOY page (=> real)
  ✓  7 › OLD locator is satisfied WITHOUT opening any tab (=> vacuous)
  ✓  8 › OLD opener silently no-ops on a NONEXISTENT tab (=> vacuous)
  ✘  9 › NEW opener FAILS on a NONEXISTENT tab (=> real)
  ✓ 10 › OLD assertion passes with NO filter applied at all (=> vacuous)
```

Six passes (every "old is vacuous" demo), four failures (every "new is real" demo) — the designed outcome, and no app-code edits were needed. Harness deleted after the run.

**What was vacuous:**

- **`signed-in-smoke.spec.js`** — all four signed-in page tests matched words the shell nav prints on *every* route (`nav-model.js` defines "Trade", "Settings", "Team Strength"). Replaced with the page's own `<h1>` (the shell owns none — only `ui/PageHeader` and `ds/PageHeader` render one) plus contract-derived data: the roster page must power-rank every real team, settings must expose the real source registry, the dashboard's switcher must list all 12 teams.
- **`public-league-visual.spec.js`** — under `--ignore-snapshots` (what `npm run e2e` uses; no baselines committed) all four tests asserted *only* that a bare `<section>` existed. `LeagueClient.jsx:243` wraps the whole page in one, so it matched on every tab — and the tab opener was `if (await btn.count())`, a silent no-op when the button was missing.
- **`public-league.spec.js`** — "archives filter narrows the result set" counted rows once and asserted `> 0` **without applying a filter**. The 12-tab walk slept 150 ms per tab and asserted nothing about rendering.

Two assertions are **kept and documented rather than "fixed"**: `critical-smoke`'s `/Brisket/i` marker (its load-bearing assertion is "no JS errors", and A7 already records the trade-off), and `expect.soft` in `attachConsoleGuards` (soft assertions still fail the test in Playwright — correctly used).

## Skip gates: four were permanently off

Measured against the seeded snapshot, not inferred:

```
rivalries len=45   matchups len=158   players len=968   sleeper teams: 12
```

All four data-availability gates (`journey-trade:87`, `public-league:126/196/209`) could never fire. They only stood ready to convert a real pipeline regression into an invisible green skip. Each is now an assertion that the data is present.

| Category | Count |
|---|---|
| Legitimate project/env gate (kept) | 8 |
| Data gate that never fires (converted to assertions) | 4 |
| Conditional escape hatch (kept; fires in practice) | 1 |
| Dead / unreachable test body (recorded, not deleted here) | 2 |

Of the 30 skips in a full run, **29 are project gating**. The one exception is the settings-override degrade hatch, whose recorded reason is `/api/rankings/overrides request failed: Failed to fetch` — root cause is §3.3, not the override engine.

`critical-smoke.spec.js:169/176` are **worse than skips**: they push a "skip" *annotation* and `return`, so Playwright reports them **passed**. `/api/terminal` is permanently 401 anonymous, so both bodies are unreachable — two green ticks that verify nothing. Recorded as dead; left for a follow-up rather than mixed into this batch.

## New coverage

Only **9 of the 33** selector-registry entries were referenced by any spec. Two new files close the highest-value part of that gap — chosen for blast radius, not count:

- **`api-trade-intelligence.spec.js`** — `/api/draft-capital` (board must be exactly `numTeams × draftRounds`; every pick owned and priced; **round-1 mean value must exceed the last round's** — the core pick-curve property, and one that survives the nightly refresh; every `isTraded` pick must actually have changed hands) and `/api/trade/suggestions` (asserts `totalSuggestions > 0` *before* iterating — the A8 trap; every asset within the advertised top-150 gate; and **every give-side player must be on the submitted roster**).
- **`journey-tools-health.spec.js`** — `/tools/source-health` (rendered names must match `/api/status`'s `enabled_sources`; **both** the populated and the empty contract are pinned rather than skipped) and `/tools/trade-coverage` (one audited row per Sleeper team, all 12 real names, scan counter reaching 12/12, positive player counts).

Every expectation is fetched inside the test — no hardcoded names, and no assertions on colours, fonts or class names, so the design-system rewrite can proceed underneath.

## Environment findings that bound what any e2e run can prove

- **§3.1 — the suite's page topology is not production's, in both directions.** Production nginx routes `location /api/` to the backend and `/` to Next, so the **22 handlers under `frontend/app/api/**` serve no production traffic at all** (even `/api/dynasty-data`, which `server.py:3084` implements). The suite points page navigation at the Next origin (`global-setup.js:227`), so the browser's `/api/*` calls land on those dead handlers. Consequence 1: the 404s for `/api/terminal`, `/api/leagues`, `/api/user/state` are **test artifacts, not product bugs**. Consequence 2, the one that matters: any spec passing *because* a Next `/api/**` handler answered is exercising routing production does not have. This is now the documented reason `api-trade-intelligence.spec.js` drives the backend via `page.request`.
- **§3.2** — `reuseExistingServer: true` + hardcoded `PORT = 8000` means a run silently tests whatever stack is already up. Mid-audit this was a *different worktree's* build; all measurements were re-taken against an isolated frontend built from this branch.
- **§3.3** — `app/api/status` and `app/api/auth/status` abort at 3 s and fall back to a silent null, so under load the UI renders empty states. This is the mechanism behind the "load-sensitive" failures #566 filed as D1/D2.
- **§3.4** — `/league` SSRs in 7.8 s and its 2.6 MB payload exceeds Next's data-cache ceiling; a rapid 12-tab walk reads an identical 2718-char body for every tab. Per-section content is asserted one-navigation-per-section instead. Related to #555 item 3.
- **§3.5** — `public-league`'s no-private-endpoints assertion is declared **deliberately scoped**, not accidentally narrow. It guards the anonymous public hub; `/league/phases` fetching `/api/data` (#578) is correct behaviour behind an auth gate, so widening the listener would fail a working feature. The genuinely uncovered case is named in both the spec comment and the audit rather than hidden.

## Test results

| | specs | passed | skipped | failed |
|---|---|---|---|---|
| Before (`origin/main` + preflight fix) | 12 | 145 | 30 | 3 |
| After | 14 | — see below — | | |

The before baseline was **already red** — the premise that this suite was green does not hold; #566 measured the same 145/30/3.

Late in the session the shared backend's `user_kv` store began returning `500 OperationalError` on `/api/user/state`, persistently and reproducibly outside Playwright:

```
$ curl -b cookies http://127.0.0.1:8000/api/user/state
{"error":"internal_error","context":{"endpoint":"/api/user/state","errorType":"OperationalError"}}
a1=500 a2=500 a3=500
```

That same test **passed in the before baseline 20 minutes earlier**, and this branch changes no backend code — the backend process in this container is owned by a different worktree. The suite caught it correctly; it is reported rather than worked around. Final counts are in the thread below once the run lands.

---

### DIFF
--- a/docs/e2e-assertion-audit.md
+++ b/docs/e2e-assertion-audit.md
@@ -575,14 +575,35 @@ breaks when the snapshot refreshes.
 
 ## P2.5 — New environment findings (§3)
 
-### §3.1 `/api/terminal` has no Next.js proxy route
+### §3.1 The suite's page topology is not production's — in both directions
 
-`frontend/app/api/` defines 21 proxy handlers. `/api/terminal` is not
-one of them, and neither are `/api/leagues`, `/api/user/state` or
-`/api/health`. In production nginx serves `/api/*` from the backend on
-the same origin, so this is invisible — but the suite's documented
-topology points page navigation at the **Next** origin
-(`E2E_PAGE_ORIGIN`), where a relative `fetch("/api/terminal")` 404s:
+**Corrected after the #578 handoff, which supplied the nginx half.**
+My first pass filed this as "`/api/terminal` has no Next proxy route".
+That is true but it is the less important half, and read alone it points
+at the wrong culprit.
+
+Production nginx (`deploy/nginx/chaseupside-proxy.conf`):
+
+```
+location /api/ { proxy_pass http://dynasty_backend;  }
+location /     { proxy_pass http://dynasty_frontend; }
+```
+
+So in production **`/api/*` never reaches Next at all**. The 22 route
+handlers under `frontend/app/api/**` do not serve a single production
+request — even `/api/dynasty-data`, which the backend implements itself
+(`server.py:3084`).
+
+The suite does the opposite. `global-setup.js:227` and `pageUrl()` send
+page navigation to `E2E_PAGE_ORIGIN || http://127.0.0.1:3000` — the Next
+origin, with no reverse proxy in front — so a relative `fetch("/api/…")`
+from the browser lands on Next. Two consequences, and the second is the
+one that matters:
+
+1. **404s at the Next origin are test artifacts, not product bugs.**
+   `/api/terminal`, `/api/leagues`, `/api/user/state` and `/api/health`
+   have no Next handler, so they 404 here and work fine in production.
+   Observed on a signed-in dashboard load:
 
 ```
 404 http://127.0.0.1:3100/api/health
@@ -592,12 +613,34 @@ topology points page navigation at the **Next** origin
 502 http://127.0.0.1:3100/api/auth/status
 ```
 
-Consequences, observed: the signed-in dashboard renders **"Terminal data
-unavailable"** with `<h1>Pick your team</h1>` and every aggregate tile
-`—`; `/tools/trade-coverage` reports **"FETCH ERRORS 12"**. New specs
-therefore assert contract-derived content and explicitly do **not**
-assert terminal-derived values. That limit is stated in the spec
-comments rather than hidden.
+   The visible result: the signed-in dashboard renders **"Terminal data
+   unavailable"** with `<h1>Pick your team</h1>` and every aggregate
+   tile `—`; `/tools/trade-coverage` reports **"FETCH ERRORS 12"**.
+   New specs therefore assert contract-derived content and explicitly
+   do **not** assert terminal-derived values — stated in the spec
+   comments rather than hidden.
+
+2. **Every spec that passes *because* a `frontend/app/api/**` handler
+   answered is false comfort.** That code does not run in production.
+   It covers `/api/status`, `/api/auth/status`, `/api/rankings/overrides`,
+   `/api/dynasty-data`, `/api/draft-capital`, `/api/angle/*`,
+   `/api/trade/suggestions` and 15 more. A browser-driven assertion
+   against any of them is exercising a routing arrangement that exists
+   only in the test harness.
+
+   This is why the new `api-trade-intelligence.spec.js` drives
+   `/api/draft-capital` and `/api/trade/suggestions` through
+   `page.request` (baseURL = the **backend**, port 8000) rather than
+   through a page. That path matches what nginx does in production. It
+   was originally chosen to dodge the §3.3 auth-status race; the
+   topology argument is the stronger reason and is now the documented
+   one.
+
+**Recommended follow-up (not done here — it is a harness change, not an
+assertion change):** give the suite a same-origin reverse proxy, so page
+navigation and `/api/*` share one origin split the way nginx splits it.
+Until then, treat any browser-observed `/api/*` behaviour as unverified
+against production.
 
 ### §3.2 The suite silently reuses another checkout's stack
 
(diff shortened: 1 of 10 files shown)


jasonleetucker-code:
## Final results — rebased onto `74d5c556b`

### Before / after

| | specs | passed | skipped | failed |
|---|---|---|---|---|
| Before (`origin/main` + preflight fix) | 12 | 145 | 30 | 3 |
| After | 14 | 146 | 34 | 6 |

The **+4 skips are intended**: the two new spec files' four tests gating off `mobile-chromium` via `desktopOnly`.

**The before baseline was already red** (145/30/3) — #566 measured the same. The premise that this suite was green never held.

### One of the six failures was mine, and is fixed

The rewritten home-dashboard test asserted the top-bar team switcher, which is CSS-hidden at 390px where `MobileChrome` takes over — the element resolves but reports `hidden`. Genuine viewport coupling, not flake. Fixed by gating that describe block to desktop (matching the existing convention; the originals ran on mobile but asserted only shell text, so nothing real is lost). Verified:

```
  ✓ [desktop-1366] home dashboard renders the war-room surface with a real team list (3.0s)
  ✓ [desktop-1366] trade builder renders its own page body, not just the nav link (3.1s)
  ✓ [desktop-1366] rosters page power-ranks every team in the contract (3.3s)
  ✓ [desktop-1366] settings page lists the real ranking-source registry (1.6s)
  - [mobile-chromium] (all four) — skipped, desktop journey
```

### The rest are not this branch — and the proof is direct

The backend on `:8000` was booted by a **different worktree** (§3.2). That worktree was reclaimed and rebuilt underneath the still-running process:

```
$ ls -la .../worktrees/agent-ac26f9e0ef914b8f6/data/
-rw-r--r--  4096  session_store.sqlite
-rw-r--r--     0  user_kv.sqlite        <-- 0 bytes
(no dynasty_data_*.json at all)
```

A zombie: the process survives and its in-memory contract still serves (`/api/data` → 1095 players, 12 teams), but every disk-backed endpoint is broken:

```
$ curl -b cookies /api/user/state
{"error":"internal_error","context":{"errorType":"OperationalError"}}   (a1=500 a2=500 a3=500)

$ curl -b cookies /api/draft-capital
{"error":"Draft data workbook not found or empty","leagueKey":"dynasty_main"}
```

`/api/user/state returns 200` **passed in the before baseline 20 minutes earlier**, and this branch changes no backend code. Restarting that process wasn't mine to do.

### The accident became the best demonstration of the PR's point

`/api/draft-capital` silently lost its entire pick board. **Before this branch there was no test on that endpoint at all** — the regression would have been invisible, a green run over a valuation endpoint returning nothing. With the new spec:

```
  ✘ api: draft capital valuation › GET /api/draft-capital prices a complete,
    internally consistent pick board (587ms)

    Error: draft-capital returned no picks
    Expected: > 0
    Received:   0
```

Meanwhile `POST /api/trade/suggestions` in the same file passed, because its data path was unaffected — so this isn't a blanket failure, it's a specific true one.

### Every changed spec verified individually against a healthy backend

- vacuity proof harness — 10/10 as designed (6 "old is vacuous" pass, 4 "new is real" fail)
- `signed-in-smoke` + `journey-tools-health` — 11/11
- `public-league` + `public-league-visual` — 20/20 (two of which, `?tab=` deep-link and the franchise deep link, were **failing in the before baseline**)
- `api-trade-intelligence` — 2/2
- preflight — watched fail (exit 1) then pass (exit 0) against an identical empty `data/`

### Where the green is still not trustworthy

Stated plainly in the audit rather than left implied:

1. **Browser-observed `/api/*` behaviour is unverified against production** (§3.1) — the largest remaining hole; needs a harness change (same-origin reverse proxy), not an assertion change.
2. **`critical-smoke.spec.js:169/176` still report green while executing nothing** — annotation-and-return over permanently-401 endpoints. Recorded, deliberately not fixed in the same PR as the rewrites.
3. **25 of 33 selector-registry entries remain unused** — `/tools/ros-data-health`, `/intel`, `/trending`, `/angle`, `/draft`, `/league-comparison`, `/players/compare` are still uncovered. This PR closes the highest-value part of the gap, not the gap.
4. **The suite still silently reuses whatever stack is on :8000/:3000** (§3.2). Tonight that meant testing another worktree's build, then a zombie. Until that's fixed, any run's provenance is unproven — arguably a bigger trust problem than any single assertion.

---

PR #585, opened 2026-07-27
Collaborative audit: adjudicate 15 claims, fix 7, refute 4, and move the trade finder onto the board the user can see

jasonleetucker-code:
An external reviewer supplied an audit (findings A–S) covering the ROS utility model, roster intelligence, TE scarcity, trade engines, source independence, storage and security — explicitly framed as **hypotheses to verify, not directives**. This adjudicates all of them and implements the ones that survived.

## The prompt's own framing was stale

It assumed a fan-out of unmerged branches carrying these findings. In fact `origin/main` == `HEAD` == `a6edd364` at branch point, one unrelated PR was open, and #553–#582 had all merged — several already fixing findings raised as open.

| outcome | findings |
|---|---|
| **confirmed** (7) | A units · E surplus · K finder value path · N mixed markets · J fit-score range · P source families · Q no persisted actuals · S security |
| **already fixed / handled** (3) | C tier ladder · D strength-vs-dependency · I unidentifiability |
| **dead code at branch point** (2) | G `lineupScarcity` · the LI-7 apparatus — **superseded mid-review, see below** |
| **partially confirmed** (1) | L negative delta — latent, filtered before output |
| **disputed with my own sub-reviewer** (1) | O KTC VA on `rankDerivedValue` |
| **raised independently** (2) | nflverse 2025 URL 404s silently · conftest contradicts the code it isolates |

## F-6 — the headline fix

`finder.py` read `_finalAdjusted` (raw scraper composite) while `/rankings`, `suggestions.py`, `angle.py`, `waiver.py`, `monte_carlo.py` and every UI surface read `rankDerivedValue`. The arbitrage engine's core comparison was against a board no user could see. Open since 2026-07-26 as "recorded, NOT actioned".

Independently re-measured before touching code — **803 paired assets, k = 0.875, 189 orphans** against PR #567's 803 / 0.880 / 194. The prior measurement reproduces.

**Thresholds re-derived, not ported.** Porting would have tightened every gate ~14%. Percentile-matching was tried and rejected: it is degenerate at the low end (`MIN_ASSET_VALUE` sits at the 99.25th percentile of the paired pool, `JUNK_THRESHOLD` at the 100th) and conflates the scale change with a population change.

**A correction to F-6's own prediction.** It said this "moves every number the endpoint emits". Measured across all 12 teams on one held-constant payload: 436 → 435 trades, median `boardDelta` 1047 → 1030, and the **top recommendation unchanged for 0 of 12 teams**. The levels move; the ordering barely does — the dominant score term is a ratio, so a near-uniform rescale (ρ = 0.9626) cancels. A coherence fix, not a behaviour change.

**202 unpriced assets** leave the engine's universe. They are now counted in `metadata.assetsUnpricedByBoard` and warned about, rather than vanishing into a shorter list that reads as "nothing available".

## TE — a conclusion I got wrong, corrected mid-review

⚠️ **An earlier revision of this PR description said the league's TE premium was "exactly 1.000" and that the correction runs downward. That was wrong and is retracted.**

I measured TE demand from **scoring keys alone** — `bonus_rec_te` is 0.0 and every TE key is matched by its WR/RB equivalent — and concluded there was no TE premium. **This league starts two mandatory tight ends** (`roster_positions` contains `TE, TE`, and TE is FLEX- and SUPER_FLEX-eligible on top). Structural demand is demand whether or not a scoring key rewards the position; twelve teams must field twenty-four tight ends every week, and that is the league class KTC's TE++ boards exist for.

I had confused the *mechanism* with the *thing itself* — the same failure mode this audit kept finding in other people's work.

**What survives, and gets stronger.** The target basis is **TE++**, which is what the blend already assumes ("KTC is the canonical TE++ retail signal … every other source's TE values are scaled to align with KTC's TE++ baseline"). KTC's own measured uplift is **1.209–2.053**, rank-dependent, while the live `_TE_BLANKET_NON_NATIVE_MULTIPLIER` is a flat **1.15** — below the entire observed range. Non-TEP boards are lifted **too little**, so correcting it moves TE values **UP**.

**The double-count guard is structural, not a convention.** `measure_te_demand` returns a target *basis*, never a multiplier — a basis cannot be multiplied into anything. The only mover is `convert_te_value(v, from_basis, to_basis)`, which is a no-op when they match (this is the `ktcSfTep` case, since it is already `tepp`), cannot compound on a second call, and raises rather than interpolating an unmeasured pair. Verified: `base 8169 → tepp` gives **9878**, matching Brock Bowers' real KTC pair.

Four functional forms were compared; the log-linear fit scored R² 0.82 and **predicted a ratio below 1.0 at the top of the board** — a TE premium that lowers a TE's value. The adopted power form is monotone and ≥1 by construction.

**Built, tested (23 tests), and deliberately NOT wired.** Wiring it is a live value change on the default board and needs the CLAUDE.md rule-4 downstream sweep. The machinery is in place; the flip is one reviewed step.

## Security

- `LOCKSTEP_SETUP.md` named **`178.156.148.92` as the "Current production target"** with user `dynasty`. That IP belongs to a third party — the old domain lapsed and was re-registered. An operator following it would SSH into someone else's box.
- The bare-IP nginx block served the **whole app including the login POST over plain HTTP**. Also broken on its own terms: `JASON_AUTH_COOKIE_SECURE` is true, so the browser dropped the `Set-Cookie` and bare-IP login could never persist a session. Now health-only over plaintext, everything else 301s to HTTPS — deleting it outright would have silently broken any unknown external monitor.
- Four runbooks repointed off the lapsed domain.

## Where I disagreed with my own sub-reviewer

Finding O was reported as significant: KTC's scale-sensitive VA (`t = 10041`) applied to `rankDerivedValue`. The call sites are real; the severity is not. That board tops at 9999 against 10000/10041 — **0.4% apart**. Downgraded to a documentation note. Changing a formula whose inputs are within half a percent of what it was built for trades a documented assumption for an undocumented one.

## Finding G was superseded while this PR was open

PR **#584** merged `LEAGUE_ADJUSTED_IS_NOOP = False`, added `src/league_intel/publish.py`, and exposed `GET /api/league-adjusted-values`. So "lineupScarcity is dead code" is no longer true and the finding is live again — recorded as **SUPERSEDED**, not edited away.

What did *not* change is the part that matters here: `data_contract.py` has no reference to league-adjusted values, so LI-7 publishes a **parallel surface** and does not modify `rankDerivedValue`. And `te_premium_axis` remains `ABSENT`, with a rationale that already says a correction there "would double-count the blend's existing alignment multiplier". **The blend owns TE alignment; LI-7's TE axis stays inert.** The staged TE wiring is safe today and stops being safe the moment anyone supplies a measurement to that axis.

## Also fixed

`tradePartnerFitScore` is nominally 0–100 with a reachable range of **7.24–43.12** (`DECISION_CALIBRATED` confidence needs rejection data Sleeper does not expose, so `MAX_CONFIDENCE_WITHOUT_DECISION_DATA = 0.45` never even binds). Bounds now ship with the score, derived from the caps and cross-checked against an exhaustive sweep. Window `probabilities` → `affinities` — softmaxed anchor distances summing to 1 are not calibrated probabilities. `marginalPoints` → `marginalStrengthIndex`. `replacement_gap` → `replacement_level` (it held a level; `partner.py` already consumed it as one). The dead ROS projection branch that discarded DraftSharks' real projections.

## Testing

**57 new tests, each observed failing against its defect before the fix.** The ROS trio was additionally mutation-tested — making the branch actually consume projections fails all three, proving they aren't vacuous. The nginx guards fail 5-of-6 against the pre-fix config.

Baseline `pytest -m "not livedata"`: **4075 passed / 0 failed**. After: **4132 passed / 0 failed**.

*(Two failures in `test_fantasypros_idp_integration.py` appear only in unfiltered runs. They reproduce identically on `a6edd364` — this branch's own base — in a clean worktree, the module is `livedata`-marked, and that CI step is `continue-on-error`. Not from this branch.)*

## Not deployed

Nothing here is deployed. The nginx change needs operator confirmation that no external monitor still probes `169.58.50.224` — verifiable only off-repo.

## Deferred, with reasons

Source families (P), persisted actuals (Q), removal-cost surplus (E), mixed-market posture (N) — each recorded in `docs/collaborative-model-audit/IMPLEMENTATION_STATUS.md` with the missing evidence or product decision that blocks it. Full record: claim registry, formula registry, experiment log (including the rejected curve forms, the rejected threshold methods, and one retracted experiment), and before/after artifacts.


jasonleetucker-code:
## ⏸️ Converted to DRAFT — held on purpose. Do not merge.

**For any autonomous orchestration tick reading this:** this PR is green and mergeable, and it must still **not** be merged. Draft status is the mechanical guard; this comment is the reason.

### Why it is held

One number in it is disputed, and the dispute is unresolved.

This branch measures KTC's TE++ uplift as **median ×1.319** (73 TEs, `ktc.csv` vs `ktcSfTep.csv`). `claude/league-intel-projections` → `docs/league-intelligence/DATA_SOURCES.md` §7 records **median ×1.368** from *the same two boards*:

| metric | LI workstream | this branch | gap |
|---|---|---|---|
| median ratio | 1.368 | **1.3187** | 3.7% |
| TE1-12 | 1.287 | **1.2270** | **4.9%** |
| TE41+ | 1.512 | **1.5382** | 1.7% |
| population | 462 (74 TE + 388 control) | 500 (73 TE + 427 control) | 38 rows |

The tail agrees; the top does not. That matters because the fitted curve's floor is the observed minimum — if the LI figure is right, this branch under-credits the most valuable tight ends by ~6%.

### Two causes already eliminated — do not re-test

1. **TEP level.** `Dynasty Scraper.py:1947` requests `?sf=true&tep=3`, which looks like a smoking gun. It is not: both CSVs come from **one** payload, and `_ktc_extract_tep` deliberately pulls `superflexValues.tepp` (**level 2**) while the base SF value goes to `ktc.csv`. The pair is internally consistent.
2. **Population.** LI excluded pick rows. 427 controls − 38 picks = **389** vs their 388. Zero picks appear among the differing rows, so it cannot move the ratio.

### Still open

Whether "median ×1.368" is even the same statistic (mine is the median of per-player ratios); their measurement code in `src/league_intel/calibration.py`; snapshot date; and the one extra TE row (74 vs 73).

### Condition to lift the hold

Reconcile the two numbers, **or** confirm with the operator that shipping the *rest* of this PR without the TE wiring is acceptable — the TE curve here is **built and tested but deliberately not wired into the live value path**, so the disputed number does not currently affect any served value.

### Relationship to #586

Checked, not assumed. The two branches share `CLAUDE.md`, `src/api/data_contract.py` and `src/api/gameplan.py` but touch **non-overlapping hunks** (this branch: 1067/1635/2408/6846/8088/8110/8125; #586: 2053/7479). `git merge-tree` reports **zero conflict markers**, and #586 touches **no TE code**. #586 already notes "§7 TE alignment — not started. Overlaps #585's `te_premium.py`" — that coordination is correct and should hold.

Worth flagging for whoever sequences these: #586's league-adjusted overlay moves TEs **down** (positional scarcity factor < 1 — Bowers 2→4), while this branch's staged TE work would move them **up** in the base blend. Different surfaces — base blend vs opt-in overlay — so they compose rather than collide, but that composition should be a decision someone makes on purpose, not a side effect of merge order.

---

PR #651, opened 2026-07-30
Insider Trading in full: ledger read path, sell/buy lead scoring, sharp records crawl

jasonleetucker-code:
Follow-up to #626. That PR built the ledger and the window-safe metrics — but never wired them to a response. This PR connects them, then builds both products on top.

Three commits, each independently reviewable:

1. **Read-path cutover** — `/api/intel/*` reads the ledger; `aggregate.py` deleted
2. **Stage 4 + D9/D11/D13** — sharp records crawl and three open defects
3. **Stage 3** — Insider Trading sell/buy modes and lead scoring

---

## 1. Read path (was this PR's original scope)

`service.py`'s payload builders still called `aggregate.build_asset_summary`. The ledger was written on every crawl and never *read*. So the merged work was plumbing without an outlet, and the live board still counted **waiver adds as buys**, sorted by **`trendScore`**, and showed **net without volume**.

- `build_summary_payload` / `build_player_payload` / `build_member_payload` now query the ledger per window via `signals`. Each window is its **own indexed query** over the raw rows; nothing sums two windows.
- **New `GET /api/intel/waiver-interest`** — waiver + free-agent activity under its own label, fields named `adds`/`drops`, no `net`. A flag on the board would have rendered the numbers correctly without preventing the defect, which was waiver churn *reading as* trade buys. A claim and a later drop are two decisions, not a round trip.
- `window` (`7d`/`30d`/`90d`/`all`, default `30d`) and `sort` are server-side and **allow-listed**, so a query string can't reach `signals.window_bounds` with an unknown name.
- **`aggregate.py` and `test_aggregate.py` deleted**, not deprecated — no competing formula stays reachable. `holdings_from_state` moved to `service.py`, since current ownership is a snapshot fact, not a ledger event.

### The risk this cutover introduced, and the guard

**The ledger is global.** One file spanning every crawled league, including the thousands of managers Sharp Tracker's discovery graph reaches. The old snapshot was partitioned per league so scoping was free — it no longer is. Every query filters to `_pool_member_ids(state)`.

`tests/intel/test_read_path.py::TestLeagueScoping` is the regression barrier, including the empty-pool case — an `IN ()` built from an empty list that matches *everything* is the classic way this goes wrong.

### Self-healing deploy

The snapshot and the ledger are populated at different times, so there's a window where a league has a good snapshot and an empty ledger. The board would render "no activity" — indistinguishable from a quiet week, and it would look like the feature broke rather than like a missing migration.

`ensure_ledger_synced` backfills on read, once per (league, snapshot mtime), rather than depending on an operator remembering `scripts/migrate_intel_ledger.py`. Safe because ingestion is idempotent by construction: `movement_id` is the crawler's deterministic event id and every insert is `INSERT OR IGNORE`.

### A behaviour change worth naming

**A manager in two of your leagues now appears on both boards.** That is correct — Insider Trading asks what your league-mates traded *anywhere*, so if the same person is a league-mate in two leagues, their activity is relevant to both. The old per-league isolation was an artifact of which crawl recorded the event, not a semantic boundary. Pinned by `test_shared_member_appears_in_both_leagues`.

---

## 2. Sharp records crawl (Stage 4) + three defects

`src/sharp/records.py` walks each discovered league's `previous_league_id` chain to build the completed-season record the Sharp Score needs — wins, losses, finish rank, championships.

It **reuses `src/public_league/`** (`walk_league_chain`, `regular_season_settings_record`, `playoff_teams`, `season_champion`) rather than adding a fifth Sleeper fetcher, and builds a minimal `SeasonSnapshot`: **4 API calls per season instead of 85**, which is what makes a graph-wide crawl affordable at all.

Chain termination is on the string `"0"`, not a falsy check — Sleeper returns `previous_league_id: "0"` for a league's first season, and `if not prev` reads that as "keep walking" in one direction and "stop" in another depending on type coercion.

Live measurement from the Megalabowl seed: **311 completed season rows, 144 managers, back to 2021.** Cohort funnel: 356 observable → 90 with records → 23 evaluable → **6 qualified** (26% of evaluable, inside the top-¼ target).

Ships with `scripts/crawl_sharp_records.py` (`--stats` / `--score` / `--league` / `--budget`) and a daily systemd timer at 04:50 UTC, 30 minutes after discovery.

**D9** — pick holdings were seeded unconditionally for 3 future seasons off a **wall-clock** year, even when `/traded_picks` was empty. Now keyed on the league's own `season` and its `draft_status`, so a completed draft doesn't manufacture holdings and an off-by-one calendar year doesn't shift every pick.

**D11** — `faab_contention.load_intel_snapshot` was called unwrapped inside an async handler, putting blocking file I/O on the event loop. Wrapped in `run_in_threadpool`. The regression test walks the AST rather than matching source lines, because the first version of it false-failed on correctly-wrapped multi-line code.

**D13** — any signed-in session could `POST /api/intel/refresh` unthrottled. Now cooldown-limited. Ordering matters and is pinned: the "already running" **409 is checked before** the cooldown 429, or a refresh that is genuinely in flight reports as rate-limited instead.

---

## 3. Insider Trading sell/buy modes and lead scoring (Stage 3)

The board answered "what has moved". This answers the actual trade question — **"who do I call about this player, and why"**.

Two modes over **one** observation set, read from opposite sides:

| Mode | Question | Scores their… | Current owner |
|---|---|---|---|
| **Sell** | "I want to move X — who wants him?" | **buying** of X elsewhere | excluded |
| **Buy** | "I want X — who has him?" | **selling** of X elsewhere | included, flagged `ownsAsset` |

### The score

`src/intel/leads.py` — additive on 0–1 over five bounded terms (demonstrated interest .34, partner fit .26, positional need .18, value match .12, activity .10) minus explicit contradiction and thin-evidence penalties.

Deliberately **not** log-odds: this is a ranking, and borrowing the odds machinery would imply a calibration that does not exist. Recency is a **45-day half-life**, not a window cut. **Breadth across leagues outranks repetition within one** — paying for a player twice against two different markets is stronger evidence of appetite than paying twice against the same one.

### It composes `roster_intel.partner` rather than extending it

`assess_partner` already models need, competitive window and fairness as bounded log odds. Adding a "demonstrated interest" term *inside* it was rejected because five surfaces consume it (gameplan, trade suggestions, angle, roster engine, terminal), its `contributions` is asserted by **exact set equality**, and a separate test **independently reimplements** its logit assembly. Those tests exist to stop a term being bolted on casually.

So partner fit enters as one input among several — normalised against `FIT_SCORE_REACHABLE_MAX`, because the fit score **cannot reach 100 by construction** and dividing by a nominal 100 would silently halve the term. `partner.describe_limitations()` already pre-declared this exact hook: *"Cross-league decision data from `src/intel/` if it ever captures offers rather than only completed trades."*

### Home-league exclusion

`home_league_ids` drops the league being asked about. Without it, "has my league-mate bought this player *elsewhere*" counts the trade made in *this* league — and every current owner reads as an enthusiastic buyer of their own player.

### Epistemics, inherited verbatim

**A lead score ranks who to approach first. It is never a prediction that anyone accepts.** Sleeper does not record declined offers at all, so an acceptance rate is *unobservable*, not merely unmeasured. Tests assert no probability-shaped key can appear anywhere in the payload, and the UI renders the limitations block rather than hiding it in a tooltip.

**Demonstrated interest is the one genuinely observable signal here** — a manager who traded *for* this player in another league revealed a preference, in public, with real assets. Everything else is inference and is labelled as such.

**Absence of observed interest is not evidence of disinterest.** Managers with no observations stay on the ranked list below those with signal, rather than being filtered out; filtering would present our crawl coverage as a fact about the manager.

### Degradation

Every input below the ledger is optional, because a missing contract is the normal state of a fresh process. No contract, no roster snapshot, no starter settings, and an `assess_partner` exception all **abstain**; only a missing league snapshot is a hard 503. A blank fit renders as an **em dash, never `0.0`**, so "not computed" and "computed as bad" cannot read the same.

---

## UI

`trendScore` replaced by **Net / Volume / Buys / Sells / Mgrs / Leagues / Confidence**, with window, sort, direction and search filters.

The per-asset drill-down splits into **Evidence / Trade leads** tabs — keeping the observation/inference line visible — where Evidence carries the underlying movement list so any count on screen can be audited back to its trades. A name lookup makes leads reachable for a player who hasn't moved and therefore has no board row.

Both counts always render (`2 / 1 opp`, never `+1`), same rule as the board and for the same reason: a net figure hides whether the evidence is one-sided or contradictory.

---

## Defect register

D6, D9, D11 and D13 close here. Of the original 13, **only D10 remains open** — `movement_id` embeds the *attributed* owner, so a co-ownership change on an already-crawled transaction can re-key it. Narrow, deliberately deferred, and cheap to fix later because `tx_id` is already a ledger column: a canonical re-key is one `UPDATE ... GROUP BY` pass. Documented in `docs/intel/AUDIT.md` §7 rather than left to be discovered.

## Tests

+76 this PR (33 scorer, 18 assembly, 13 endpoint, 12 vitest). Four existing tests asserted the *old* behaviour — their fixtures used waiver transactions and expected them on the buy/sell board; updated to pin the corrected behaviour, with a trade added to the endpoint fixture so it pins both directions at once.

Full suite green: **Python 621 in `tests/intel tests/sharp tests/roster_intel`**, repo-wide `ruff format --check .` clean, **frontend 1538 across 90 files**.

New docs: `docs/intel/LEADS.md` (the lead model and its epistemics), alongside `AUDIT.md`, `METRICS.md`, `SHARP_SCORE.md`.

## Manual validation

1. `/league/insider-trading` — no Trend column; Volume, Mgrs, Leagues, Confidence present
2. Switch the window filter — counts change per window and never sum
3. Expand a row → **Evidence** — member exposure plus the movement list
4. Expand a row → **Trade leads** — toggle I'm selling / I'm buying, expand a lead for its component breakdown
5. Type a name into the lookup — leads for a player with no board row
6. `GET /api/intel/waiver-interest` — the waiver assets as `adds`/`drops`, disjoint from the board
7. Deploy with an empty ledger — the board self-heals on first read rather than showing nothing


jasonleetucker-code:
Heads-up on a boundary defect on the page this PR rewrites, so it doesn't get re-introduced during the rebase.

`/league/insider-trading` is currently reachable **without a session**. Measured on production 2026-07-30:

```
GET /league/insider-trading   -> 200
GET /api/intel/summary        -> 401
GET /api/intel/player         -> 401
```

The page has no auth gate of its own, so an anonymous visitor gets a permanent `Loading Insider Trading…` spinner — and `robots.txt` serves `Allow: /league/`, so that's an indexable URL.

It's public only by accident: `PUBLIC_PREFIXES = ["/league"]` in `frontend/lib/public-routes.js` matches by prefix, which is correct for the public league hub, but it makes every future `/league/*` page public by default. This is the first one where that default is wrong — the same defect class that module's docstring already records for `/trades`.

I've added a `PRIVATE_EXCEPTIONS` deny-list checked before the prefixes, in #657 (`78dcc14`). A deny-list rather than another one-off removal, because `/trades` was fixed by removing it from a list and removal can't help a route nobody had to add.

**Nothing for you to do if #657 lands first** — the carve-out is in `public-routes.js`, which this PR doesn't touch, so there shouldn't be a conflict with the page rewrite. Flagging it only so that if #651 merges first, or if the rewrite adds its own gating, the two don't end up disagreeing about whether this route is public. If you'd rather own the gate inside the page, say so and I'll drop the carve-out.

---

PR #655, opened 2026-07-30
Remove the fixed-slot draft model and add Perfect Draft

jasonleetucker-code:
## Previous behaviour

`/draft` modelled **six rookie picks per team**, and that was not cosmetic:

| Site | What it did |
|---|---|
| `effectiveBudgetFor` | reserved $1 per unfilled slot, and returned **0** once a team's picks were "used" — erasing real money from `topCompetitorMax`, so a team that had drafted six rookies was modelled as unable to bid |
| `phaseMultiplier` | `1 + slotPressure × 0.5` multiplied into **every MaxBid** on the page |
| `mdv` | remaining ÷ openings, driving the Teams heatmap and the "Spend up" chip |
| `mergeDraftCapitalTeams` | `Math.min(feedSlots, 6)` silently clamped a team owning nine picks |

The doc comment claiming the constant was only a pre-fetch placeholder was false. There was also no budget optimizer: the board priced rookies individually but never answered which *combination* a budget should buy.

## Changes

**Slot model removed (unconditional, not flag-gated).** Effective budget is remaining dollars; the rival ceiling is the richest rival's actual remaining. `phaseMultiplier` is **deleted rather than replaced** — `budgetAdvantage` already carries "I'm rich, bid up", and more meaningfully now that budgets aren't slot-shrunk. `draftProgress` is denominated by the rookie pool; "Spend up" keys on board progress + budget advantage. Rookies bought is still shown, as information rather than a limit.

This **changes live MaxBid numbers** and gives up parity with the Inflation spreadsheet, which encodes the same wrong assumption. That trade-off was confirmed before starting.

**Perfect Draft** — a new panel answering which combination of rookies produces the greatest net roster improvement.

## The model

A literal `max Σ value s.t. Σ price ≤ B` is degenerate. Rookie values and the $1200 ladder come from the same convex Hill curve, so value-per-dollar rises **~30x** down the board (#1 at 7587/$135 vs #72 at 1721/$1) — the optimizer would just read the price curve back out and buy thirty $1 darts. So both sides are measured over replacement:

```
waiverValue(pos) = best unrostered player at that position
surplus(rookie)  = max(0, boardValue − waiverValue(pos))
ECC(player)      = max(0, base − waiverValue(pos)) × (0.85 + 0.30·waiverScarcity)
   base          = boardValue, or waiverValue when unranked ("assumedWaiver")
NetValue(S)      = Σ surplus − D(|S|)
```

An empty roster spot is worth waiver level, not zero. Unranked roster players cost 0 and are stamped — the board's 375/497 floor is an artifact (19 players at each, single-source, no position), and costing cuts at it would turn an identity-join miss into a free-cut recommendation.

Roster capacity is real and per-team: measured on the live league, five of twelve teams sit at exactly 58 while another has 14 spots free. Rookies beyond `openRosterSpots` each displace a **different** player, cheapest first.

**Algorithm.** The cut ladder is cheapest-first and independent of *which* rookies are bought, so displacement is a function of `k` alone — one cardinality-constrained knapsack solves every plan size at once, and the star/depth alternatives are the ends of that frontier rather than a second algorithm. Exact, not heuristic: legal cut-sets are the independent sets of the dual of a transversal matroid. Droppability is therefore a **matching** via the real lineup solver, never a per-position count, because FLEX/SUPER_FLEX make it set-dependent.

**`planMaxBid` is an indifference price** — `max{q : bestNetWith(i at q) ≥ bestNetWithout(i)}`. The tempting `price + (netWith − netWithout)` adds value units to dollars. `bestNetWithout(i)` doubles as the live pivot ("if he goes past $X, spend it on …").

**Confidence** is a seeded bootstrap `P(this plan wins)` over the frontier *plus* the pivots — without the pivots, two tied plans of the same size are invisible.

## Architecture

Server serves `GET /api/draft/roster-context` (rosters at board value, waiver levels, cut ladder — static for a whole draft, needing the 4 MB contract, lineup solver and scarcity). Client runs the knapsack against live localStorage state, where budgets and picks actually live. Not a frontend ranking engine: it recomputes no value. `rookieBoardValue` is added to `/api/draft-capital` because the dollar ladder isn't invertible, and is redacted for public callers. Cache key carries team identity.

## Two bugs found by verifying against live data

- **Knapsack reconstruction repeated a player twelve times.** A single `choice[k][b]` pointer isn't enough: the cell an item consumed can be improved by a later item, so the pointer read afterwards may name an item not on the path. DP *values* stayed correct, which is what made it subtle. Replaced with a per-item decision table.
- **`Number(null) === 0`** made the strategy tilt treat every player missing `yearsExp` (16% of the board) as a rookie.

## Tests

`5,251 passed` (Python, `-m "not livedata"`) · `1,583 passed` (Vitest). Baseline before the change was 5,206; no pre-existing test was weakened.

New: `tests/draft/` (43) and `frontend/__tests__/perfect-draft.test.js` + `PerfectDraftPanel.test.jsx` (59). Coverage includes all 16 requested scenarios, plus:

- **greedy value-per-dollar loses to the exact solve** on a constructed instance
- **greedy cut ladder == brute-force optimum** at every cardinality (the matroid claim, checked rather than asserted)
- plan invariants — no duplicate rookie, spend/gross/net internally consistent — which are what the reconstruction bug violated
- monotonicity: more budget and more open spots never lower net value
- per-team cache isolation

`frontend/__tests__/draft-logic.test.js` gains a "no per-team slot limit" block asserting the removed exports and stamps are gone.

## Known limitation

Every addition is measured against the **same** waiver level, but the honest comparison for a `k`-rookie plan is against the 1st…k-th best free agent, which declines. Total surplus is optimistic, increasingly so as `k` grows, biasing toward high-`k` plans. The fix mirrors `D(k)` — a `W(k)` term — and preserves the decomposition. Documented in `docs/perfect-draft.md`, not silently shipped.

Also unresolved: BDVM can't price the 2026 rookie class (upstream nflverse gap), taxi assignment isn't ingested anywhere (moot for the main league — no taxi, no IR), and the second league's draft-capital fallback emits no rookie fields, so the panel vanishes there rather than optimizing against placeholders.

## Docs

`docs/perfect-draft.md` (technical reference), ADR-009/010/011 in `docs/roster-trade-intelligence/DECISIONS.md`, and a CLAUDE.md section. Rollback: `RISKIT_FEATURE_PERFECT_DRAFT=0` + restart — note this does **not** restore the slot model, which was a correction rather than a feature.

https://claude.ai/code/session_01M6oCgi8MGEbZUUh9LsTGtu

---
_Generated by [Claude Code](https://claude.ai/code/session_01M6oCgi8MGEbZUUh9LsTGtu)_


jasonleetucker-code:
**Landed with two deliberate deviations** on `main` via #723, which consolidated six open PRs onto one branch. Please read this before assuming the branch is fully represented.

**What landed:** Perfect Draft in full — `src/draft/`, `frontend/lib/perfect-draft.js`, `PerfectDraftPanel.jsx`, `/api/draft/roster-context`, the `perfect_draft` flag (ON), the backtest script, and every new test — plus the removal of the fixed-slot rookie model from `draft-logic.js`.

**Deviation 1 — `consensus_edge` stays `False`.** This branch has it `True`. `main` set it to `False` the same day on measurement: the audit found the edge was carried by rows priced on a scale that does not exist, and ADR-023 requires `scripts/validate_consensus_edge_board.py` to pass before the default flips again. Because this branch was 358 commits behind, merging it verbatim would have silently switched an unvalidated buy/sell board back on for every user. `main`'s `False` was kept, and `safe_on` in `test_feature_flags.py` gained `perfect_draft` but not `consensus_edge`.

**Deviation 2 — `_serialize_pick` keeps `isUnpriced`.** The serializer here is genuinely needed (it adds `overallPick` and staples the rookie board onto current-season slots) but it omits the `isUnpriced` flag, which is the D-2 / audit-finding-C1 fix documented at length in CLAUDE.md and pinned by tests. The serializer landed with the flag restored inside it.

**Two test suites were reconciled, not silenced.** `main`'s C1 test passed `num_teams`, which this branch removed (team count now comes from the roster feed — correctly, per this branch's own reasoning). This branch's rounds test summed `dollarValue` across every pick from an empty contract, which throws now that unpriced picks are `None`; given a contract that prices the current-season slots, its actual claim is stated more directly than before. No behaviour was changed to make a test pass.

**On the conflict count:** 57 of the 70 conflicts were generated artifacts under `data/`, `CSVs/`, `exports/`. 80 of this branch's 223 commits are `chore: automated refresh` and no non-chore commit touched any of those files, so all 57 took `main`'s copies. Also dropped: the `react-markdown` dependency, which nothing in the tree imports.

Closing as superseded.

---

PR #657, opened 2026-07-30
Restore the E2E safety net, and close the guard that let it drift

jasonleetucker-code:
This is the outcome of a priority audit that asked "if we could complete only one piece of work right now, what should it be?" The answer was **AC-03** — the item `docs/OWNER_ACTION_AUDIT_2026-07-29.md:1321` already sequenced next, whose stated prerequisite (D-6: "merge #625 first, then fix the specs") cleared on 2026-07-29 at 11:12 UTC and which had not been started.

## Why this, and not a feature

The nightly `E2E Safety Net` has been red since #625 merged — 13 failures on 2026-07-29 (`30438842408`), **19** on 2026-07-30 (`30529404019`), against a fully green run on 07-28 (`30345417749`). It is the only gate that exercises real user journeys:

| gate | what it checks | browser? |
|---|---|---|
| `pr-validation.yml` | pytest, vitest, ruff, build, bundle size | no |
| `deploy.yml` post-deploy smoke | HTTP codes, `/api/health`, `/_next/static/*` resolution | no assertion on text or behaviour |
| `e2e.yml` nightly | every journey | **yes — and red** |

So a renamed heading, a dead filter and a page stuck in its loading skeleton all merged green, deployed green, and were reported only by a suite already red and therefore ignored. `e2e.yml:20-28` is cron + dispatch only, "deliberately NOT on every PR".

It also unblocks branch protection: `OWNER_ACTION_AUDIT:210` and `:1070` note that only `Validate PR` can be required *because* E2E is red.

**Sixteen of the nineteen failures were the harness. Three were real.**

## The harness (16)

**Eleven were one missing `pageUrl()` wrapper.** `gotoRankingsBoard` did a bare `page.goto("/rankings")`, resolving against `baseURL` — the FastAPI page proxy. `_proxy_next` forwards no cookies (`server.py:2891`; `server.py:12539` says so outright), so once `frontend/middleware.js` landed every such navigation 307'd to `/login`. Six rankings journeys, the settings-override round-trip, two mobile smokes and two chart smokes all reported *"rankings board should render rows / element(s) not found"* — which reads like a dead value pipeline and was a dead cookie. **In the same run, a spec that navigated correctly dumped a board with 968 players and 200 rendered rows.**

**Four were one rename.** #625's naming canon renamed "Trade Builder" → "Trade Calculator" and "Roster Dashboard" → "Team Strength". The specs hardcoded the old strings in three files. `tests/e2e/specs/` had not been touched since 2026-07-27 — two days before the rename.

**One was a retired route.** `/finder` is a redirect shim into `/rankings` now. Left failing deliberately — see *Not in this PR*.

## The root cause

`nav-model.js:32-38` states a load-bearing naming rule: a nav leaf's label is the same string as the page's `<h1>` and as `pageTitleFor()`'s answer. `nav-model.test.js` pinned the first two and claimed the third was covered *"in `__tests__/components/page-title-canon.test.jsx`"*. **That file did not exist** — the only hit for `page-title-canon` in the entire frontend was the comment asserting it did. The `<h1>` half was pinned by nothing, behind a comment saying otherwise: `ORCHESTRATION.md` §6.15's guard whose stated purpose and actual predicate differ.

Closed, one definition, every consumer importing it:

- **`frontend/__tests__/helpers/naming-canon.js`** — the canon, hoisted out of `nav-model.test.js` so it can be imported rather than duplicated.
- **`frontend/__tests__/components/page-title-canon.test.jsx`** — the promised file. Split into *mechanism* (both PageHeaders really turn `title` into the page's single `<h1>`, rendered for real) and *wiring* (every canon route really passes its canon string to one, read from source), because rendering `/draft` (5,185 lines) or `/trade` (2,105) in a unit test would be a flaky integration test. **It found a live violation on its first run:** `/league/activity` rendered "League activity" against a nav label of "Activity" — fixed here.
- **`tests/e2e/test_e2e_harness_guards.py`** — bans bare literal `page.goto("/…")` and retired page names, and pins the E2E title canon to the frontend one. Runs in the **blocking** pytest step, so the next occurrence fails in seconds on the PR instead of overnight.
- **`TITLE` + `titleFor()`** in `helpers/journey.js`. Anchored exact-match regexes: a loose `/Trade/i` matches a nav label present on every authenticated page, the vacuity class `docs/e2e-assertion-audit.md` measured **passing with `<main>` deleted entirely**.

Both new guards were verified by reintroducing the exact defects they describe and watching them fail, then reverting. A guard nobody has watched fire is the thing this PR is about.

## The real defects (3)

**`useTeam`'s `loading` could never settle.** The auto-assign effect opens `if (!selectedLeagueKey) return;`, so it never reaches its `setAutoAssignEvaluated` call, and the gate then read `!autoAssignEvaluated[""]` → `true` on every render, forever. This is the same hole its own comment says it closed for the no-match case — closed for one branch of the early return, left open for the other. `/tools/trade-coverage` never left "Loading team list…" so the page had no `<h1>` at all; `/waivers` stayed in its skeleton. **Reachable in production for any user with no stored team.** Fixed at the gate so the gate and the effect agree on one predicate.

**`/waivers` had an unreachable empty state.** `if (loading || !analysis)` returned the skeleton before `if (!hasTeam)`, and `analysis` is null for as long as no team is selected — so the skeleton was permanent and "Pick your team" was dead code. Gated on `!loading` rather than moved first, so a user *with* a stored team still gets the full-height stub the CLS work put there.

**`archives.jsx` called hooks conditionally.** `if (!data) return` sat between the `useState`s and the `useMemo`s, so the `data: null → present` transition — this section's normal path, it is lazily loaded — took the hook count 3 → 5 and threw. The throw lands during an update, leaving committed markup on screen with handlers detached: the section renders, then the kind buttons and season select do nothing.

That is the shape of the `Production E2E Smoke` failure, red on its last four consecutive runs. The spec clicked "Matchups", selected a season, and the row count stayed at **exactly 190** through both — and 190 is the unfiltered `trades` total, the default `kind`, matching no season-filtered subset of anything. Measured from production (`GET /api/public/league/archives`): trades **190** (2026:37, 2025:124, 2024:29), weeklyMatchups 138 (**2026:0**), waivers 1092, rookieDrafts 424, seasonResults 30. **The season filter was never the bug — it had simply never run.**

## Verification

- vitest **1533 passed / 90 files** (baseline 1526 / 89)
- 7 new pytest guards passing; `ruff format` + `ruff check` clean
- both new guards confirmed to fail on the reintroduced originals
- expected-value sources are the production API payload, not the components under test

## Not in this PR

**No valuation change.** Nothing here touches `rankDerivedValue`, `_compute_unified_rankings`, the Hill curves or BDVM.

Deliberately deferred: the `/finder` spec (re-pointing that coverage at `/rankings?screen=` is a behaviour decision, not a rename, and the dead `finder*` registry entries go with it); the nightly's CI-only data-shape failures; and making `E2E Safety Net` a required check — that waits for three consecutive green nights, because requiring a flaky check deadlocks every merge.

## Found while auditing, filed rather than folded in

Higher user impact than parts of this PR, and out of its scope:

1. **The ROS join is case-mismatched.** The aggregate stores `canonicalName` lowercase; the contract stamps it Title Case; `PlayerPopup.jsx:134` and `RosTradeFitPanel.jsx:135` do a bare exact lookup. Measured over 949 rows: **937 (98.7%) render "ROS · no data yet — No ROS source ranked this player today", and 790 of those have a ROS row differing only by case.** A confident false claim about 790 players. Fix at the join, not the file — `canonicalName` doubles as a display fallback.
2. **`portfolio-insights.js:249` resolves 0 of 288 picks** (Sleeper says `2026 1.02 (own)`, the contract says `2026 Pick 1.02`). 22.5% of every team's portfolio value dropped; all 12 teams permanently show an "N unresolved" warning. `trade-logic.js:1079` already has the resolver.
3. **`/api/public/league` served an empty contract for ~8.5 minutes** on 2026-07-30, HTTP 200, `rebuild_failures: 0`, while `last_contract_bytes` stayed frozen at its last healthy value — so a size-based monitor reads green through a total content outage. Note the nightly's data-shape assertions ("must expose rivalry pairs / matchups", "player index must not be empty") are exactly what detects this; they were dismissed as CI-only noise.
4. **Three divergent starter-slot tables** (`league-analysis.js:42` has DL/LB/DB 2/2/2 against the registry's 3/3/3) — `/rosters` "Starters only" undercounts every team by mean 7.4% and **reorders 4 of 12 teams**.
5. **`draftSharksRosSf` is counted twice** — the "IDP" fetch writes a second full 1QB board under the same source key, giving one vendor 67.4% of the ROS blend against an intended 50.8% and shifting 715 of 1084 values.

Also: the push warned of **35 dependabot advisories on `main` (17 high, 14 moderate, 4 low)**. Worth taking seriously — and worth doing *after* this, so the restored suite can gate each bump.

https://claude.ai/code/session_01SKhaEruQrSLhWKZPEGmEEm

---


jasonleetucker-code:
Scope grew past the original description — recording what and why.

**Two name joins that silently dropped most of their data** (`1d86aa3`). The ROS join matched **12 of 1075** board rows while rendering "No ROS source ranked this player today" for the rest — a confident false sentence, not a blank. The key was a measured choice, not a reflex: a bare `.toLowerCase()` recovers 802, `normalizePlayerNameKey` recovers 849, because the aggregate strips apostrophes (`jamarr chase`). 849 is the ceiling — the other 226 genuinely have no ROS row, so the message is finally true where it still appears. Separately, `portfolio-insights.js` resolved **0 of 288** picks (Sleeper says `2026 1.02 (own)`, the contract says `2026 Pick 1.02`); it now resolves 216, and the test asserts the remaining 72 **are** the 2029 picks, so a future change that reaches 288 fails rather than inventing values for picks the board doesn't price.

**A live production outage, caught in the act** (`a72c504`). Merging #639 restarted the backend, and at 16:18 UTC `/api/public/league` served **7,403 bytes** — `leagueName: ""`, `managers: 0`, all 17 sections empty, HTTP 200. Every aggregate signal read healthy: `rebuild_failures: 0` and `last_contract_bytes: 2005444`, frozen at the last good build because the only code that writes it is guarded on `and snapshot.seasons`. A size-keyed monitor reads 2 MB and green through a total content outage of the one page anonymous visitors can reach.

`build_public_snapshot` returns a season-less snapshot instead of raising when the Sleeper chain walk comes back empty, so the entire success path ran on it — cooldown cleared, empty snapshot cached with a fresh `fetched_at`. The TTL is why it persisted: the first `?refresh=1` inside the window was deduped and returned the empty payload in 0.93s with `rebuild_count` unmoved. **Production is restored** (2,005,437 bytes, 8/8 startup checks) via a forced rebuild once the window lapsed.

A zero-season snapshot now takes the failure path: counted, cooldown armed, last good snapshot served if there is one, refused if not. 6 of the 7 new tests fail against the pre-fix server — verified by removing the guard.

This is the fifth instance of the pattern `docs/ORCHESTRATION.md` §6.15 names, and worth noting because it argues against a call made in the 2026-07-29 audit: the nightly suite's public-league data-shape assertions ("must expose rivalry pairs", "must expose matchups", "player index must not be empty") were dismissed there as CI-only noise the runner couldn't satisfy. They are precisely what detects this outage. They should be kept and fixed, not deleted.

Verification across the branch: **1743 backend tests** (api + public_league, 108 subtests) and **1551 frontend** across 92 files, ruff clean on every changed file.

---

Issue #661, opened 2026-07-30
[CORRECTED — not a product defect] The archives E2E test races hydration on the production payload


jasonleetucker-code:
> **Corrected 2026-07-30.** The original title and body claimed the `/league` archives section was inert in production. **That was wrong.** The section works; the test was racing hydration. The original analysis is kept below the line so the reasoning error is visible, because the way I got it wrong is the useful part.

## What is actually happening

`visitLeague` waits for the text `"Public archives"`, which is present in the **SSR HTML**. That proves the markup arrived, not that the page is interactive.

Against production the gap is large: the `/league` document is ~960 KB with the archives section (~737 KB) inlined in the RSC payload. Hydration takes long enough that a click fired the instant the text appears can land on DOM with no handlers attached and is simply lost. The subsequent 15s poll then watches a table that was never going to change.

Locally the snapshot is small, hydration is quick, and the race resolves the right way — which is why the nightly passed while prod flapped.

Fixed in #662: click and re-select **inside** the poll so a pre-hydration interaction is retried rather than swallowed. Both handlers (`setKind`, `setSeason`) are idempotent. No assertion was loosened.

## The evidence that settles it, and that I had all along

Thirty consecutive scheduled `Production E2E Smoke` runs:

```
FFFF.SFFSFSSSSSSFSFSFFSS.FSFFF     13 passed · 15 failed · 2 cancelled
```

**Interleaved.** A genuinely inert control fails 100% of the time. This one passed 13 times.

I recorded that ~46% failure rate early in the investigation and then reasoned straight past it. The row-count measurements were concrete and interesting — 190 is exactly the unfiltered `trades` total, 158 is exactly what the button declares — and I let a satisfying mechanism outweigh the one statistic that contradicted it. The alternating record was disconfirming evidence sitting in plain view the whole time.

## What was real

Two genuine defects were found and fixed on the way, and both were worth having independently:

- **#657** — `archives.jsx` called `useState` ×3, returned early on `!data`, then `useMemo` ×2: a real Rules-of-Hooks violation that would throw on the `null → present` transition this lazily-loaded section takes.
- **#660** — the row locator counted `table tbody tr` page-wide across ten `/league` sections that share `.table-wrap`, and the post-click wait was `toBeGreaterThan(0)`, which the *pre-click* table already satisfied. That is why the number under assertion was the trades total.

[...]

PR #662, opened 2026-07-30
Make the archives interactions hydration-tolerant (corrects #661)

jasonleetucker-code:
**This corrects a diagnosis I got wrong.** Issue #661 says the `/league` archives section is inert in production. It isn't. The test was racing hydration.

## What settles it

The run history, not the code. Thirty consecutive scheduled `Production E2E Smoke` runs:

```
FFFF.SFFSFSSSSSSFSFSFFSS.FSFFF     13 passed · 15 failed · 2 cancelled
```

Interleaved. **A genuinely inert control fails every time.** And the same spec passed consistently in the nightly `E2E Safety Net`, where the committed snapshot is small.

I had this number in front of me from the start — I recorded the "~46% failure rate" early on and then reasoned past it, because the stack of measurements pointing at the row counts was more interesting than the one fact that contradicted the conclusion.

## The actual mechanism

`visitLeague` waits for the text `"Public archives"`, which is in the **SSR HTML**. So it proves the markup arrived — not that the page is interactive.

Against production that gap is real and large: the `/league` document is ~960 KB with the archives section (~737 KB) inlined in the RSC payload. Hydration takes long enough that a click fired the instant the text appears can land on DOM with no handlers attached, and is simply lost. The subsequent 15s poll then watches a table that was never going to change.

Locally the snapshot is small, hydration is quick, and the race almost always resolves the right way — which is why the nightly passed and prod flapped.

## The fix

Click and re-select **inside** the poll, so an interaction that lands pre-hydration is retried instead of swallowed. Both are idempotent — the handlers are `setKind(k.key)` and `setSeason(value)`, so repeating them sets the same value.

**Nothing is loosened.** The assertions are unchanged and still exact:
- row count must equal the count the button itself declares (`Matchups (158)`)
- the season filter must strictly narrow
- every surviving row must be the selected season

Only the *waiting* changed, which is the part that was wrong.

## What this means for the earlier work

The two defects fixed along the way were real and worth having — the Rules-of-Hooks violation in `archives.jsx` (#657) and the unscoped page-wide row locator plus unfalsifiable `toBeGreaterThan(0)` wait (#660). Neither was the flake, though. The honest sequence is: two real bugs found and fixed, one wrong conclusion drawn about a third, corrected here.

I'll update #661 to match rather than leave a wrong diagnosis sitting in the tracker.

https://claude.ai/code/session_01SKhaEruQrSLhWKZPEGmEEm

---

### DIFF
--- a/scripts/audit_identity_matches.py
+++ b/scripts/audit_identity_matches.py
@@ -161,6 +161,44 @@ def _count_csv_rows(csv_path: Path) -> int:
         return 0
 
 
+# Canonical names where the pool genuinely contains ONE player twice,
+# so an alias merging them is correct rather than a defect.
+#
+# ``alias_collision_delta``'s own docstring carves this case out — "a
+# same-family collision is also a defect ... unless the pool genuinely
+# contains one player twice" — but the code never implemented it, so
+# the exception had nowhere to live and the first real instance turned
+# the hard gate red on ``main``.
+#
+# Entries are NOT a way to quiet the check. Each one needs a verified
+# same-player claim, and every collision outside this set still fails.
+# The collisions are still reported and still printed in the run
+# summary; they are marked, not hidden.
+_KNOWN_POOL_DUPLICATES = {
+    # "Rob Henry" + "Robert Henry" — one player, two pool rows.
+    #
+    # ``src/utils/name_clean.py`` aliases "robert henry" → "rob henry"
+    # with the identity verified on both sides: RB, UTSA UDFA → WAS,
+    # age 24. The alias existed to bridge DraftSharks/FantasyPros'
+    # "Robert Henry Jr." to the pool's "Rob Henry".
+    #
+    # The 2026-07-30T18:06Z scheduled refresh then added "Robert Henry"
+    # to CSVs/dynasty_full.csv alongside the existing "Rob Henry", so
+    # the pool now carries both spellings and the alias merges two rows
+    # that are the same person. That is the alias working, not failing.
+    #
+    # It reports ``crossFamily`` only because the new row carries no
+    # position and lands in the OTHER group — absence of position data,
+    # not evidence of a second player. The vote-replication hazard this
+    # check exists for (a WR absorbing a CB's ballot) needs two KNOWN
+    # and different families.
+    #
+    # Removing the alias was the alternative and is worse: it would
+    # leave one player sitting on the board as two separate rows.
+    "rob henry",
+}
+
+
 def alias_collision_delta(pool_rows: list[dict[str, Any]]) -> list[dict[str, Any]]:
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
**CI is red here, but not because of this PR — `main` is red.** Naming the blocker rather than pushing a fix, because the fix is a judgement call I shouldn't make alone.

## What failed

Step 11, the blocking Python gate — and this PR changes one `.js` file. CI's own log says `No changed .py files — nothing to lint.`

```
FAILED tests/scripts/test_audit_identity_matches.py::TestCommittedDataRegression::test_script_end_to_end_exit_zero
assert report["aliasCollisions"] == []

+ {'canonicalName': 'rob henry',
+  'crossFamily': True,
+  'mergedNames': ['Rob Henry', 'Robert Henry'],
+  'positionGroups': ['OFFENSE', 'OTHER']}
```

## Bisected to the exact commit

| commit | time | result |
|---|---|---|
| `df9e451` (#657 merge) | 13:03 | 8 passed |
| `3b8025c2` (#660 merge) | 13:55 | 8 passed |
| `6b2814d4` freshness stamps | 18:06 | 8 passed |
| **`cd7ac93e` chore: automated data refresh** | **18:06** | **1 failed** |

The scheduled scrape did it. Its diff to `CSVs/dynasty_full.csv`:

```
 Rob Henry,1538,1,,        ← existing pool row
+Robert Henry,1,2,1,       ← NEW row added by this refresh
```

## Why that trips the guard

`src/utils/name_clean.py:326` carries a deliberate, well-documented alias:

```python
"robert henry": "rob henry",  # DS/FP "Robert Henry Jr." ↔ pool
# "Rob Henry" (RB, UTSA UDFA → WAS, age 24 on both sides)
```

It was written to bridge DraftSharks/FantasyPros' spelling to the pool's. The pool now carries **both** spellings natively, so the alias merges two rows that already exist separately — and because the new row has no position it classifies as `OTHER`, so the merge reads as cross-family.

## Two fixes, and I don't want to pick blind

1. **Remove the alias.** It is arguably obsolete now that the pool carries "Robert Henry" itself. But if they are the same player — and the comment says the author checked position, school, team and age on both sides — this leaves one player sitting on the board as two rows.
2. **Narrow the `crossFamily` rule** so an unknown position (`OTHER`) is not treated as a distinct family. Defensible, since the guard exists to catch a WR merging into a LB — but it weakens a guard to make a test pass, which is the exact move I have been arguing against all day.

There is also a decent chance the next 2-hourly refresh drops the duplicate row and this resolves itself, which is worth knowing before anyone edits the alias table.

**This PR is otherwise ready** — the change is test-only, and the diagnosis it carries (the archives spec races hydration; it was never an inert control) stands on the 30-run pass/fail record independently of this failure.

---

PR #670, opened 2026-07-31
Curated industry sharps, the D10 double-count fix, and the transaction crawl

jasonleetucker-code:
Rebuilt on current `main`. Four commits, in dependency order.

**Scope changed from the original version of this PR.** It previously also carried a Sharp Tracker board UI (`SharpBoard.jsx`, `/api/sharp/board`). `main` has since replaced that with the unified Sleeper + FFPC page, and that board had a **Rookie picks** tab, which contradicts the standing rule that the Sharp Tracker shows players only. Those three commits are dropped. What survives is the work `main` still doesn't have.

---

## 1. D10 — the last open double-count

`crawler._pool_holdings` substitutes a pool co-owner when the primary owner sits outside the pool, and recomputes that from live `/rosters` **every run**. The movement key embedded that attributed user, so a co-ownership change produced a *different id for a movement already stored*, and `INSERT OR IGNORE` had nothing to match against.

```
before   {tx_id}:{attributed_owner}:{action}:{asset_id}[:disc]
after    {tx_id}:r{roster_id}:{action}:{asset_id}[:disc]
```

A Sleeper `roster_id` is the league slot: fixed for the life of the league, independent of who owns or co-owns it. The attributed user rides along as a column.

An **opposite** bug fell out of the same change: under the old key, one user co-owning two rosters collapsed both movements into a single event — an undercount. Keyed by slot, both survive.

**Migration.** Old rows can't be recomputed (they never stored a roster id), so `SCHEMA_VERSION` → 2 and `_open_or_reset` now actually *checks* it. It was written from the start but never read.

The migration clears the **movement tables only**, and that distinction is load-bearing: the discovery graph and season records share this file, cost hundreds of Sleeper calls, and are unaffected by a movement-key change. `test_the_discovery_graph_survives` is the guard.

Rebased onto `main`'s newer `_open_or_reset`, which distinguishes SQLite corruption from a routine `database is locked`. That distinction is kept — a busy database is still never rebuilt; only the version check is added on top.

## 2. The transaction crawl

`ledger.ingest_events` had exactly one caller — `src/intel/ingest.py`, migrating Insider Trading's league-mate snapshot. Discovery had mapped the managers and the records crawl had scored them, but **nothing had ever fetched their trades**. A cohort is not a signal.

`src/sharp/transactions.py` is league-driven where `intel.crawler.crawl` is member-driven. It does not re-implement transaction parsing — `_events_from_tx` already handles paired add/drop semantics, pick discriminators, status filtering and the D10 key, and this repo already had four Sleeper transaction fetchers. A fifth would be the problem, not the feature.

Pool is every roster owner in the league, with the qualified-cohort filter applied at *read* time: qualification moves as records accumulate, so a manager who qualifies next month arrives with history rather than starting blank.

6-hourly, not daily, and offset from the 04:20/04:50 slots — 48h is a headline window, and a once-daily crawl would make "last 48 hours" mean "as of this morning".

## 3. Curated industry sharps

Materializes #690's staged payload as ordinary reviewable source. The 13 base64-XZ `.curated_*` chunks and the materializer workflow are gone.

### The workbook contradicted the staged data, and the workbook wins

The `Verified Sleeper username` column reads literally **"Not publicly verified" for 92 of the 100 people**. The 8 rows that carry a username attach *podcast and company URLs* as evidence — those establish why the **person** belongs in the universe and say nothing about who holds the handle. Four are exact lowercase transforms of the person's own X handle:

| claimed username | X handle |
|---|---|
| `carpentiernfl` | `@CarpentierNFL` |
| `mattykiwoom` | `@MattyKiwoom` |
| `raygque` | `@RayGQue` |
| `charleschillffb` | `@CharlesChillFFB` |

That is the handle-equals-username inference the brief forbids. The staged importer stamped all 96 platform accounts `verification_status: "verified", confidence: 1.0` and seeded 8 Super Sharps off the back of it.

The import now produces **0 verified Sleeper accounts and 0 Super Sharps.** Each claimed username enters as a candidate — `workbook_claimed_username`, or `..._matching_public_handle` for the four that merely echo the handle, ranked below at 0.25 vs 0.35 — and is re-derived against the public Sleeper API. X handles stay verified *accounts*: a public handle is a public identifier, and no fantasy behaviour hangs off it.

**Super Sharps starts empty and grows honestly.** That is a real state the UI renders, not an error.

### Reproducibility

The workbook is tracked (`config/sharp/workbooks/`, with a `.gitignore` negation) and `build_snapshot` stamps `workbook_sha256` + `sheet_inventory`, so a hand-edited snapshot is distinguishable from a parsed one. Two full imports produce byte-identical reconciliation:

```
totalWorkbookIdentitiesReviewed   247
totalImportedAsCuratedSharps      100
totalImportedAsResearchCandidates  30
totalScreenedOutStored            117
totalPositivelyVerifiedOnSleeper    0
totalUnresolvedSleeperIdentities   92
totalSuperSharps                    0
```

Dropped `config/sharp/{super_sharps.json,source_evidence.csv,*.csv}` — stale derived exports whose canonical home is gitignored `data/`, and whose committed copies still listed the 8 as Super Sharps.

### Three finished modules that nothing called

Each was complete and unit-tested, and connected to nothing. Isolated tests cannot catch that, so `tests/sharp/test_curated_wiring.py` asserts the connections.

- **`curated_service` was imported by nothing but its own test.** All six endpoints 404'd, and with them the `/market/sharp-people` and `/admin/sharp-identities` pages and six Next bridge routes.
- **`consensus.aggregate_person_consensus` was never called.** The "one analyst's ten leagues are not ten votes" safeguard existed and did nothing. Now stamped as `personConsensus` on every market row, **alongside** the raw movement counts, which remain the audit trail.
- **`market.py` had no curated integration.** Adds `industry`/`super`/`both`, which deliberately do **not** fall back to the automated cohort — that would answer a question about researched experts with a population that never met the bar.

### Route registration no longer depends on import order

Both sharp service modules self-register via a `_server_app()` lookup, but that side effect does not re-run if the module is already in `sys.modules`. The routes then 404 with nothing logged anywhere — the failure `cohort_status()`'s self-heal was written to recover from after it happened in production. `server.py` now calls both registrars explicitly, so registration follows from the app existing rather than from import order. This also fixes a pre-existing order-dependent failure in `test_public_api_allowlist`.

### Authorization

`POST /api/sharp/review/{id}` verifies an identity — the most consequential write in this model — and `POST /api/sharp/curated/refresh` spends a real outbound call budget. Both now require an allowlisted admin on top of the private-API gate, and the gate fails closed.

### Dead code removed

The `workbook_positive_identification` branch in `import_snapshot` has no producer once the workbook stops emitting fantasy accounts. Verified accounts now arrive only from `review_candidate`, keyed on the platform's own stable user id — a strictly better answer to the same rename problem.

## 4. A username we searched for cannot corroborate itself

Found by running the sweep **live** against Sleeper's public API rather than only against a faked transport, which is the only way this was ever going to surface.

`inspect_sleeper_candidates` decided "does this account look like this person" by intersecting two sets. The candidate set contained the queried username; the observed set contained `payload["username"]` — the same string, because we look the account up *by* username and Sleeper echoes it back. The test could never fail.

The first live sweep promoted **all 42** existing accounts to `high_confidence_probable`, including `hrr5010` "corroborating" Hasan Rahim and `amicsta` "corroborating" Anthony Amico. Nothing corresponds in either case. Those would have reached the admin review queue pre-labelled high confidence.

Corroboration now means the **person's name** matches what the account actually shows. The queried username is excluded from the candidate side; the observed username still counts, because a handle that encodes the real name is genuine evidence.

Re-run against the live API, same 92 candidates:

| | before | after |
|---|---|---|
| exists on Sleeper | 42 | 42 |
| `high_confidence_probable` | **42** | **6** |
| verified | 0 | 0 |

The six: `GrahamBarfield`, `jjzachariason`, `justinboone`, `JustinHerzig`, `mattykiwoom` (via the pseudonym split on "Matty Kiwoom") and `MattWaldman`. Each is a real name match. The other 36 stay `possible` — someone holds the handle, and that is all we know.

Of the eight usernames the workbook claimed, all eight exist, **two** corroborate by name, and **none** is verified. `carpentiernfl` does not match "Cody Carpentier" and correctly stays `possible` — it is the X handle, which is how it was generated.

Sweep cost: 92 unauthenticated `GET /user/{username}` calls at a 0.35s inter-call sleep, against Sleeper's published guidance of under 1000/min. No auth, no private leagues, no rate-limit evasion.

---

## The Sharp Tracker still shows players only

`assetType: "player"` on the request plus all three client-side pick filters are untouched. The new qualification options are additive.

## Verification

```
python -m pytest tests/ -q -m "not livedata"    5543 passed, 0 failed
npx vitest run                                  1616 passed / 98 files
npm run build                                   all pages under budget
python -m ruff format --check .                 clean
python -m ruff check <changed files>            clean
import twice                                    byte-identical
live sweep, 92 candidates                       6 probable, 0 verified
```

## Not done here

Nothing promotes a candidate to verified — that is a human decision through the review queue, by design. The queue now has 6 corroborated leads and 36 bare existence hits to work through.


jasonleetucker-code:
**Partially** landed on `main` via #723, which consolidated six open PRs onto one branch. Please read this before assuming the branch is fully represented.

**What landed:** the D10 double-count fix (`eventId` keyed on the roster slot, `SCHEMA_VERSION` 1→2), the transaction crawl and its systemd timer, the curated industry sharps model, the identity sweep, and the `curated_service` route registration — which was a real bug fix, since six `/api/sharp/*` endpoints were returning 404 because nothing imported the module.

**What did NOT land: the shadow Consensus Edge.** `src/edge/`, `tests/edge/`, `docs/edge/`, the `/consensus-edge` page, its three bridge routes, the "(shadow)" nav entry and `frontend/__tests__/consensus-edge-ui.test.js` were all dropped. This branch was 395 commits behind `main`, and in the interim `main` built and promoted its own implementation in `src/consensus_edge/`. Landing both would have put two implementations of one feature on the same routes, which CLAUDE.md rule 2 forbids.

Worth flagging for anyone doing this again: **git raised no conflict on `src/edge/`.** It arrives as a clean add, so nothing in the merge would have warned about the duplication — it was only visible by looking for it.

**How the cohort additions were integrated:** `person_id`/`network` on `CohortMember` and `curated_industry_members` were written here against `market.py`, but `main` had since moved that whole surface to `src/sharp/cohort.py`. They were ported there, including the wiring into `cohort_members` (selection branch, union, coverage counts) and the qualification vocabulary. Taking either side wholesale would have left the curated-industry population defined but unreachable. `market.py` re-exports `curated_model` and `platform_records` so this branch's monkeypatch seams still work.

The shadow Consensus Edge work is **not lost** — it remains on `claude/sharp-insider-audit-separation-fibqu6`. It was dropped because two live implementations of one feature is the problem, not because the work looked wrong. If you want it evaluated against the promoted version, that's a separate piece of work with a real question behind it.

Closing as superseded.

---

PR #680, opened 2026-08-01
Probe the live Sharp routes immediately

jasonleetucker-code:
Runs a fast production-environment probe while the longer collection job continues. It reports the production checkout SHA, backend state, registered Sharp routes, and exact responses from the cohort, combined market, and FFPC market endpoints.

PR #679, opened 2026-08-01
Run an observable live Sharp production repair

jasonleetucker-code:
This PR exists to run a production-environment workflow whose job and logs are directly observable through the PR workflow APIs. It updates production to current main, restarts the backend, prints the registered Sharp routes, probes the live market endpoint before and after collection, runs FFPC ingestion, runs Sleeper discovery and records collection, prints stored coverage and scores, and surfaces relevant systemd journals. No application code is changed.

PR #683, opened 2026-08-01
Recover the production Sharp ledger and repopulate both sources

jasonleetucker-code:
This one-time, production-environment recovery workflow addresses the SQLite lock incident discovered during live verification. It stops every ledger writer, snapshots all ledger files, inventories and integrity-checks the current and quarantined candidates, restores the valid candidate with the most historical and movement data through SQLite's backup API, restarts the lock-safe backend from current main, repopulates FFPC immediately, verifies a non-empty unified market, then installs all timers and runs one fixed Sleeper records/activity bootstrap pass. The workflow logs the full sanitized inventory and final coverage; no application code is changed.

### DIFF
--- a/.github/workflows/pr-recover-sharp-ledger-production.yml
+++ b/.github/workflows/pr-recover-sharp-ledger-production.yml
@@ -1,0 +1,294 @@
+name: PR Recover Sharp Ledger Production
+
+on:
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened]
+    paths:
+      - ".github/workflows/pr-recover-sharp-ledger-production.yml"
+
+permissions:
+  contents: read
+
+env:
+  APP_DIR: ${{ vars.PROD_APP_DIR || '/home/dynasty/trade-calculator' }}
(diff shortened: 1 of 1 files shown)


jasonleetucker-code:
Closing as superseded — same reason as #679 and #680.

A one-shot recovery PR from the 2026-08-01 Sharp incident. The incident's real follow-ups (**#684**, **#688**) merged on 08-01; `Deploy Production` has been green since (latest end-to-end success 2026-08-04 13:11:12Z) and no alert issues are open.

No activity in 3.5 days. Closed as *no longer needed*, not as *rejected*.

**Worth a look before this is forgotten:** these three PRs, plus roughly twenty single-use workflow files created 07-31/08-01 (`force-sharp-production-now`, `trigger-sharp-no-environment`, `patch-sharp-explicit-startup-registration`, and similar), are leftover scaffolding from that incident. They are still `active` and several still fire on push, which is part of why the run feed carries redundant `Verify Sharp Production Population` / `Bootstrap Sharp Records` runs. Not urgent and not touched here, but a sweep would quiet the signal.

---

PR #690, opened 2026-08-01
Integrate the full curated dynasty-sharp universe into the Sharp Model

jasonleetucker-code:
## Objective

Extend the existing unified Sleeper + FFPC Sharp Model with the entire researched workbook universe, while keeping curated expertise, empirical performance, and public trackability separate and auditable.

## Workbook scope

The import is dynamic and covers every relevant sheet rather than hardcoding the final 100:

- 247 Candidate Pool identities reviewed
- 100 curated primary sharps
- 40 Sharp Tracker candidates
- 16 FFPC/high-stakes candidates
- 30 near misses retained in the research/watchlist layer
- 35 source records plus methodology and QC metadata

## Core model changes

- Adds normalized people, aliases, platform accounts, identity candidates/evidence, membership, performance metrics, review decisions, and import-run records.
- Mandatory `curated_industry_sharp` inclusion is independent of Sharp Score v2 performance gates.
- Preserves unchanged empirical qualification as a separate pathway.
- Adds a trackable `super_sharp` designation for curated people with verified public fantasy identities.
- Never fabricates missing performance metrics; unknown values remain null with completeness/confidence fields.
- Uses verified account-to-person links and person-level consensus so one analyst's many leagues/accounts count as one independent expert signal.
- Keeps canonical event rows and computes rolling windows independently; overlapping 30d/90d windows are never summed.

## Identity resolution

- Imports the eight workbook-verified Sleeper usernames as verified usernames and resolves public Sleeper user IDs during refresh.
- Generates conservative exact-handle candidates for manual review without treating handle resemblance as ownership proof.
- Stores FFPC display/team candidates separately from people and requires evidence/review before person linkage.
- Provides approve/reject/unresolved admin review actions with one-account/one-person conflict protection.

## Product integration

- Extends the existing Sharp Tracker rather than creating a parallel tracker.
- Adds Curated Industry, Super Sharp, and Both qualification filters.
- Adds person-level signal breadth and raw movement audit counts.
- Adds a people directory, explainable person profiles, curated summary API, and admin identity-review queue.
- Adds daily public-only identity/activity refresh systemd units and an idempotent workbook import/export pipeline.

## Validation

The materialization workflow reconstructs the implementation, applies the existing-model integration patch, formats the code, and runs focused backend/frontend regression tests before committing the real source to this branch. Full repository CI will run on the materialized implementation before merge.


jasonleetucker-code:
Closing — this work is now materialized as ordinary reviewable source in #670.

This PR carried its payload as thirteen root-level `.curated_min.part-*` / `.curated_payload.part-*` files (16 KB base64 chunks of an XZ tarball) plus a workflow that unpacked them on a runner. That is a transport mechanism, not a reviewable change, so the contents were decoded locally, committed as real files, and the chunks and materializer workflow deleted.

Two things changed in the process, both worth knowing before comparing the diffs:

**The importer now parses the workbook, not the committed JSON.** The workbook is tracked at `config/sharp/workbooks/` and the snapshot stamps its `sha256`, so the import is reproducible rather than depending on an upload path.

**Every Sleeper account was demoted to a candidate.** The workbook's `Verified Sleeper username` column reads "Not publicly verified" for 92 of the 100 people, and the 8 that carry a username cite podcast/company URLs — evidence for curated *inclusion*, not account ownership. Four of the eight are exact lowercase transforms of the person's X handle. The staged importer had stamped all 96 platform accounts `verified` at confidence 1.0 and seeded 8 Super Sharps from that. The import now produces 0 verified Sleeper accounts and 0 Super Sharps; identities are re-derived against the public API and promoted only by explicit review.

Nothing from this branch is lost — see #670 for the full diff.

---

PR #709, opened 2026-08-04
Perf round 6: cut 85 KB of JS off every route, cap the filtered board

jasonleetucker-code:
Round 6 of the sitewide performance audit. Target axis confirmed with the user: **low-end mobile responsiveness** — desktop is already sub-second everywhere, while a throttled CPU still scrolled the rankings board at single-digit FPS. No architectural change.

> **Read this first.** This PR was written and measured against Next **15.5.22**. While it was open, `main` gained #712 (Next **16.2.12**) and #703 (a bundle gate rewritten to resolve chunks from disk), and both were merged in. The *directions* below all hold — the same code is split the same way — but **the absolute KB figures were measured on Next 15 and are not reproducible on the merged branch**. Re-measure before citing them. The ledger carries the same warning.

## The finding that reordered the round

Round 4's throttled profile said script cost dominates (script 1301–1829ms vs layout 310–417ms vs style recalc 255–321ms at 6×). So the question was how much JS each route executes — and CI could not see the answer. The gate filtered to `static/chunks/app/`, the page-specific slice only. **The root layout alone was 548.8 KB on every route, of which the gate counted 4.3 KB** (8–24% coverage across routes). Round 4's stated priority — "three routes within 2 KB of ceiling" — was a fact about the small slice.

## Changes

1. **`"sideEffects": ["*.css"]`** — there was no `sideEffects` field, so every barrel was only weakly tree-shakeable. Deliberately **not** a bare `false`: CSS imports are side effects by definition and `false` invites the bundler to drop them.
2. **`PlayerPopup` + `CommandPalette` out of the root layout** — 69 KB off every one of ~90 routes. See the incident below for how this is loaded and why it is not `dynamic()`.
3. **The 21 `/league` sections behind `dynamic()`.** Exactly one is ever mounted. Default tab stays static so the landing view has no loading flash.
4. **Filtered board results are capped** the way unfiltered ones already were.
5. **Dead code deleted** — five `components/ui/` modules with zero consumers, two orphaned `/league` sections, and `react-markdown`, an entirely unused dependency.
6. **`/trade`'s collapsed panel and `/rankings`' methodology charts** behind `dynamic()`. Needed `mountCollapsedChildren={false}` on `CollapsiblePanel`: the `hidden` attribute still **mounts** children, so a `dynamic()` import inside a collapsed panel fetches immediately and the split buys nothing.

**Dropped in the merge, deliberately:** this PR originally also made the bundle gate report first-load JS. That reader used `manifest.pages["/layout"]`, which Next 16 does not emit, and the disk-derived substitute counts dynamic-import chunks as always-loaded — it scored *this refactor* as +213 KB shared while `/league`'s page slice fell 163.4 → 37.8 KB. An improvement reported as a regression is exactly the failure being fixed, so the instrument was retracted rather than shipped wrong. **The blind spot is real and still open.**

## Result: real first-load JS (Next 15 measurement)

```
ROOT LAYOUT (every route): 548.8 -> 463.9 KB   (-84.9 on every one of 90 routes)

route                before    after     delta     pct
/league               718.4    503.1    -215.3  -30.0%
/draft                682.7    615.5     -67.2   -9.8%
/trade                638.2    558.8     -79.4  -12.4%
/rankings             629.2    580.8     -48.4   -7.7%
/                     622.3    564.6     -57.7   -9.3%
/settings             592.2    515.1     -77.1  -13.0%
/waivers              590.8    535.1     -55.7   -9.4%
```

Three page budgets were bumped **because of the improvement** — code that sat in everyone's shared graph is now attributed to the routes that use it. `/league` went the other way and was tightened 170→50 KB; on Next 16 it lands at 37.8 KB, confirming the split holds.

## INP — measured for the first time

The original brief asked for INP. Rounds 1–5 never measured it. Mobile 390×844, medians of 3 (good ≤200ms, poor >500ms):

| interaction | 1× | 4× before | 4× after |
|---|---|---|---|
| `/rankings` sort a column | 88 | 384 | **328** |
| `/rankings` open player popup | 144 | 488 | **304** |
| `/rankings` "Show all" | 64 | 264 | **232** |

On a **filtered** board — what the cap targets — all three were "poor" and none is now: sort 840 → **384**, expand a row 992 → **480**, open popup 584 → **280**.

## The row cap, and why windowing was not done

A full audit of `DataTable.jsx` found nine constraints beyond the one round 5 cleared — no vertical scroll container exists at all (so the sticky header is inert today and windowing would nest a scroll region); positional `nth-child(even)` zebra striping inverts every scroll frame; every row is a tab stop with no roving-tabindex infrastructure; `renderBeforeRow` gets the slice index so tier dividers land wrong; jsdom has no layout so ~25 unit tests break; and two e2e helpers assert ≥50 rows. Full checklist is in the ledger.

`displayRows` is now `ranked.slice(0, rowLimit)` unconditionally, with three things moved alongside so this isn't the dishonest version: `hasMore` no longer excludes the filtered case, the count line reads "200 of 516 shown", and `rowLimit` resets when the filter changes.

## Scroll FPS, including where it still misses

390×844, wheel-driven, rAF-sampled. `showall` **is** the pre-cap filtered behaviour, so it doubles as the before column:

| CPU | default (200) | filtered (200) | +one "Show more" (510) | "Show all" (632) |
|---|---|---|---|---|
| 1× | 58.9 | **60.1** | 48.0 | 39.5 |
| 4× | 36.0 | **29.2** | 14.5 | 8.8 |
| 6× | 22.9 | **19.9** | 9.1 | 5.6 |

Filtered scroll at 4× went **8.8 → 29.2 FPS** (p95 frame gap 182ms → 51ms). **The target was ≥30 FPS at 4× on the worst reachable board, and only part of it is met** — default hits it, filtered lands just under, and "Show more"/"Show all" cannot be helped by a cap. That needs windowing.

## The defect this PR shipped and then fixed, and what caught it

Phase 2's first implementation used `dynamic()`. `dynamic()` is `React.lazy`, which needs a Suspense boundary; `AppShell` had no local one, so the nearest ancestor was the App Router's — and `{children}`, the entire page, renders inside it. Every route's content became deferred streaming content: emitted into React's `<div id="S:1">` staging container, moved into place, and the staged copy left behind. **`/waivers` served three `<main>` elements** — the shell's, the page's, and a hidden second full copy of the page.

Measured on Next 16.2.12, `/waivers`:

| | checkboxes | `<main>` | `#S:1` |
|---|---|---|---|
| static imports | 1 | 2 | absent |
| `dynamic()` bare | 2 | 3 | **present** |
| `dynamic()` + local `<Suspense>` | 1 | 2 | absent |
| imperative `await import()` (shipped) | 1 | 2 | absent |

`ssr: false` was a red herring — it reproduces with and without. Fixed with the imperative-import idiom the repo already used (`components/ScreenshotFab.jsx`): no lazy component, so no boundary exists to enclose the page, and webpack still emits the separate chunk. The 69 KB split is intact.

**What caught it was the E2E safety net, which does not run on PRs.** `pr-validation.yml` runs pytest, vitest and lint; it never opens a browser. Every metric in this PR — CLS, LCP, INP, FPS, bundle size — was *unchanged* by a bug that rendered every page twice. Worth dispatching `e2e.yml` manually on anything that touches the shell.

## Testing

- Full Playwright suite dispatched on this branch; `waivers-smoke` passes 10/10 across both viewport projects, the popup still opens from a board row click and the palette with `/`.
- 1,659 vitest pass. Bundle gate exit 0, all 14 budgeted pages under budget on Next 16.
- 25-route sweep + 5-run CLS medians (Next 15): `/` 0.0910 → 0.0936, `/draft` 0.0682 → 0.0682, `/settings` 0.0710 → 0.0723 — all inside Google's "good" band.
- Backend confirmed not a factor: an AST scan found no expensive call passed as an argument to `run_in_threadpool` (all 26 sites are trivial coercions), and all 11 hot endpoints answer warm in 2–14ms.

## Known and not fixed

- **`/arbitrage` e2e flake — pre-existing, not this PR.** `journey-trade.spec.js` sees two "Pick a team and scan" empty states. Measured over 15 loads each: **1/15 on a build of `main`, 1/15 on this branch** — identical. `main`'s e2e passes because the spec navigates once against a ~7%-per-load race. Left open rather than silenced; the assertion is correct and the duplicate is real.
- `/draft` hits CLS ~0.18 in roughly 1 run in 5 (median 0.068), from a `DIV.muted` block collapsing at ~900ms. Pre-existing.
- `/draft`'s modals (~930 lines) sit inside the 5,185-line page module so webpack cannot split them; they need extraction first.
- The React/Next framework baseline every route pays is out of scope by decision, and now *unmeasured* rather than mismeasured — see the retracted instrument above.


jasonleetucker-code:
I merged `main` into this branch twice and changed one of its decisions. Flagging it here rather than leaving it in a commit message, because it drops a feature this PR added.

**Why I touched it at all:** I merged #703, which rewrote `check-bundle-sizes.mjs` to resolve chunks from disk instead of `app-build-manifest.json` — Next 16 emits no manifest mapping an app-router page key to its chunks. R6 had extended the manifest-based version, so the two collided. That conflict was mine to clear.

Then #712 landed `next` 16.2.12 with `next build --webpack`. That second merge wasn't optional: without the flag this branch's CI would build with Turbopack and the gate would exit 2.

**Kept:** all four budget changes — `/rankings` 65→75, `/trade` 82→92, `/draft` 128→150, and the `/league` tightening 170→50.

**Dropped: `firstLoadChunks()`.** It reads `manifest.pages["/layout"]`, which no longer exists. I tried to preserve the idea from disk — "every `.js` sitting directly in `static/chunks/`" — and measured it on this branch before trusting it:

| | shared graph | `/league` page slice |
|---|---|---|
| `main` frontend | 1286.7 KB | 163.4 KB |
| this branch | **1500.0 KB** | **37.8 KB** |

A dynamic import emits its chunk into `static/chunks/` too, so the metric counts on-demand code as always-loaded. It scores this PR's own refactor as +213 KB shared — an improvement reported as a regression, which is exactly the failure the `firstLoadChunks` docstring describes.

So I removed it rather than ship it. The reasoning is recorded in `BUDGETS_KB` so the next attempt doesn't repeat the measurement. **The blind spot you identified is real and still open** — it just needs a basis that survives Turbopack.

Two other things:

- The header docstring still claimed the script parses `app-build-manifest.json`. That was stale from my #703, fixed here.
- `package-lock.json` still declared `react-markdown` as a root dependency after this PR dropped it from `package.json`. `npm ci` prunes correctly so CI was never wrong — the lockfile just disagreed. Synced.

**Verified on this branch against Next 16.2.12:** build `--webpack` exit 0; `check:bundles` exit 0 with all 14 pages under budget and `/league` at 37.8 KB against the tightened 50; vitest 99 files / 1659 tests passing.

Separately: its **`Boot stack + run journeys`** job failed at 18:12 and I have not looked at that yet.

---

jasonleetucker-code:
**Do not merge yet.** The E2E failure is real and belongs to this PR. I checked rather than assuming, because I had it recorded as a known pre-existing failure — that record was wrong.

**E2E is green on `main`.** Run 30937854570 at `b227ba7d4` succeeded, and every scheduled run since 2026-07-31 has succeeded. The AC-03 baseline of 3 failing specs was cleared by #657/#660. So there is no baseline to write this off against.

**This branch:** run 30937231220 — 1 failed, 1 flaky, 135 passed, 49 skipped.

Both problems are the same shape — an element resolving to **two** nodes:

- `waivers-smoke.spec.js:36` — `getByLabel(/Include rookies/i)` → 2 `<input type="checkbox">`. Failed on the first run **and** the retry.
- `journey-trade.spec.js:145` — `getByText(/Pick a team and scan/i)` → 2 `<h3 class="ds-empty__title">`. Passed on retry, so it was reported flaky.

What narrows it down:

1. Each string appears **exactly once** in source (`app/waivers/page.jsx:683`, `app/arbitrage/page.jsx:304`). Two nodes means one render site produced two.
2. **Neither page is touched by this PR** — `git diff main...HEAD -- app/waivers/page.jsx app/arbitrage/page.jsx` is empty. So the cause is in a shared component, not the pages.
3. `AppShell` renders `{children}` once, and 135 tests passed, so it is not a global double-render.
4. It reproduced under Next **15.5.22** (the run's cleanup log names the version), so #712 is not implicated.
5. One case failed twice, the other passed on retry. A strict-mode violation is deterministic for a given DOM, so a retry passing means the second node was **transient** — present in one render pass, gone in the next.

Putting those together, my hypothesis — and I am flagging it as a hypothesis, I have not reproduced it locally — is that the `dynamic()` conversions open a brief window where a placeholder and the real subtree are both mounted, and the strict-mode locators land inside it. The two `ds-empty__title` nodes are suggestive: an empty state rendered by both the loading and loaded paths at once.

That is worth chasing even setting the tests aside. A duplicated interactive control is a real artifact, not just a selector problem, and `ssr: false` on an interaction-only surface is exactly where a double-mount hides.

The Playwright trace is attached to that run and will show it directly:
```
npx playwright show-trace test-results/waivers-smoke-.../trace.zip
```

Everything else on this PR is in good shape: `Validate PR` is green on `f113df091`, the branch merges cleanly, and the budgets verify against a real Next 16 `--webpack` build with `/league` at 37.8 KB against the tightened 50.

---

PR #710, opened 2026-08-04
Add Sharp Roster Percentage on the shared sharp cohort

jasonleetucker-code:
Which players our verified sharp managers actually roster — built on the **same pool** that powers the Sharp Buy/Sell Tracker, not a second list of sharps.

## One cohort, many surfaces

`CohortMember` / `cohort_members` / `curated_members` / `provisional_members` / `load_ffpc_config` move out of `market.py` into **`src/sharp/cohort.py`**, the single definition of "who is a sharp". `market.py` re-exports every name, so `sharp_market.cohort_members` and the monkeypatch seam existing tests use still resolve.

| surface | resolves its pool via |
|---|---|
| Sharp Buy/Sell Tracker | `cohort.cohort_members` |
| Sharp Roster Percentage | `cohort.cohort_members` |
| roster collection pass | `cohort.cohort_members` |
| activity crawl | `cohort.cohort_members` |

That last one is a fix: `scripts/crawl_sharp_activity.py` was re-deriving the pool itself (`build_manager_records` → `score_managers` → filter), so a change to qualification moved the boards while the crawl that *feeds* them kept selecting the old set. It was the only remaining second list.

Two tests keep it that way: one asserts the function identity across modules, another greps `roster_percentage.py` for qualification vocabulary so a future fork announces itself.

## The missing data layer

Nothing persisted what the cohort currently **owns**. The intel crawler's holdings are scoped to one league's member pool, `records.py` fetches `/rosters` and discards the player list, and FFPC roster contents sat unqueryable inside a metadata blob.

- **`src/sharp/roster_store.py`** — four additive tables. Deduplication is **structural**, not a cleanup pass: one row per roster, one row per `(roster, player)`, so re-collection, duplicate imports and repeat snapshots cannot inflate a count. History is stored as open intervals rather than daily copies. Created with `CREATE TABLE IF NOT EXISTS` and deliberately *not* wired to `PLATFORM_SCHEMA_VERSION` — bumping that re-runs the whole platform migration on every deployed ledger to add four additive tables.
- **`src/sharp/roster_collect.py`** — the third crawl pass (discovery → records → rosters). Two Sleeper calls per *league* regardless of how many sharps play in it; FFPC costs zero. Collapses season chains, so one dynasty league is one roster however many `league_id`s it has had. Every roster reached is stored, eligible or not, with its exclusion reasons attached.
- **`src/sharp/roster_percentage.py`** — the board.

## Two counting decisions worth reviewing

**The denominator is rosters, not people.** A sharp with five legitimate dynasty teams contributes five roster observations. Both numbers ship: `eligibleRosters` counts rosters, `uniqueSharpManagers` collapses linked accounts to humans.

**The denominator is per player.** A linebacker cannot be rostered in a league with no IDP slots. Dividing every player by every sharp roster reports ~20% for an IDP owned in *every* league that can roster him — an artifact of the cohort's format mix. Each player is measured against the rosters whose league is known to field his position family, and a roster that holds him is always inside his own denominator, so a percentage cannot exceed 100%.

Trends compare the **intersection** of the two roster populations and are withheld below 80% overlap, so cohort growth never reads as players gaining ownership. "No change" and "not comparable" render differently.

## Surfaces

`GET /api/sharp/roster-percentage` and `/audit`, declared directly in `server.py`.

`/market/sharp-roster-percentage` carries the full filter set, list sizes (25/50/100/all), the alternate sorts, a transparency strip, sample warnings and a collapsed exclusion breakdown. `frontend/lib/sharp-roster-percentage.js` is a pure materializer — a client-side percentage would agree today and be wrong the moment the per-player denominator rule moves, and a test greps the file to keep it that way.

## Three silent-coverage bugs fixed along the way (`e35114b`)

Each produced a board that looked healthy while omitting part of the cohort, which is why none surfaced in testing.

1. **Discovery recorded memberships only for expanded leagues.** The user→leagues branch queued a league and stored its metadata but wrote no `league_memberships` row, so leagues left on the frontier by the call budget, `maxGenerations` or `maxLeaguesPerRun` had none. The collector reads `platform_memberships` to find a sharp's leagues, so it was bounded to the expanded subgraph — and `cohortCoveragePct` counts *managers* seen, so a sharp represented by 1 of their 4 leagues still read as fully represented. Sleeper's own `/user/{id}/leagues` proves the membership, so the row is a fact rather than an inference.
2. **A budgeted roster pass re-collected the same prefix forever.** Ordering by league id meant a run that hit its call budget collected the same alphabetical prefix on every subsequent run, and leagues past the cutoff were *never* collected — a permanently invisible tail, not a slow rollout. Now uses `record_queue.prioritize_league_ids`, the fair ordering the records crawl already had. Each run reports `leaguesRemaining` so `--budget` can be sized from the journal.
3. **Route registration depended on import order.** `src/sharp/service.py` registers its market routes as an import-time side effect; anything importing it before the app exists makes that a no-op, and the module cache means the later import re-runs nothing — `/api/sharp/market` then 404s with no other symptom. `server.py` now calls the public `register_http_routes()` explicitly. This is what `tests/sharp/test_public_api_allowlist.py` was catching, so that test **now passes**.

## A bug the tests caught

A holding must persist until the observation that **contradicts** it, not until the last one that **confirms** it. Rosters are observed periodically, so a roster seen on day 0 and day 40 has no confirmation at day 10 — yet the player was rostered then. Reading the confirming timestamp silently zeroes every drop that happened between two crawls, surfacing as a 30-day trend of exactly 0.0 no matter what moved. Fixed and pinned.

## Honest gaps, documented rather than papered over

1. **No general-dynasty ownership feed exists in this platform.** Every ranking source publishes values or ranks; Sleeper's trending endpoint publishes 24-hour add counts — a flow, not a stock. So `marketRosterPct` and `sharpRosterAdvantage` are `null`, the page says why, and `set_market_ownership_provider` is the registration seam for a real feed. The sharp-vs-market sorts work but rank nothing until one is registered.
2. **FFPC contributes zero rosters today.** The parser is correct and test-pinned, but all ten configured seeds are `LeagueHome.aspx` pages that yield transactions and standings, no roster table. The FFPC half activates when a roster-bearing URL is configured — no code change needed.
3. **Trends need two collection passes** before they resolve, which is why the timer is daily rather than weekly.
4. **Coverage is still bounded by how far discovery has reached.** `transparency.rostersPerManager` is the honest readout — near 1.0 while sharps typically run several dynasty teams means discovery has not caught up, not that sharps own one team each.

Full list in `docs/sharp-roster-percentage/METHODOLOGY.md`.

## Verification

- **6076 Python tests** and **1675 frontend tests** pass.
- `pytest tests/sharp/ tests/intel/` — the partial run that previously failed on the import-order bug — is green at 447.
- `scripts/validate_sharp_roster_percentage.py` re-derives every published number from the raw rows using code that shares nothing with the engine. Run against a 69-roster / 41-manager population, all ten audit checks clear: numerator recount, one-observation-per-roster, denominator, arithmetic bounds, duplicate rosters, duplicate identity attribution, asset mapping, excluded rosters never reaching a numerator, filters moving both sides, and agreement with the Buy/Sell Tracker's own payload.
- Production build compiles the page at 4.88 kB (134 kB first load); all bundles under budget.

**The audit ran against a constructed fixture, not production data** — this container has no sharp data (`data/intel/` is gitignored and empty). The fixture reproduces the shapes that matter: multi-team managers, shared leagues, IDP vs offense-only formats, taxi/IR players, a season chain and excluded rosters. Re-running the validator on prod after the first crawl is the step that actually closes the audit requirement.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -823,6 +823,27 @@ which is why the timers are staggered:
 3. `scripts/crawl_sharp_rosters.py` (05:50) — finds what they currently
    OWN, which is what the roster-percentage board is made of
 
+Two things the roster pass depends on, both fixed rather than worked
+around:
+
+- `discovery.py` records a `league_memberships` row at USER-expansion
+  time as well as league-expansion time. It used to record one only when
+  a league was expanded, so leagues left on the frontier by the budget
+  had none — invisible to anything asking "which leagues does this
+  manager play in", which silently bounded the roster crawl to the
+  expanded subgraph.
+- The roster pass orders leagues through
+  `record_queue.prioritize_league_ids` (never-collected first, then
+  oldest), the same fair ordering the records crawl uses. Sorting by
+  league id meant a budget-capped run re-collected the same prefix
+  forever and never reached the tail.
+
+`server.py` calls `_sharp_service.register_http_routes()` explicitly
+after importing the module. The import-time side effect alone is not
+enough: anything that imports `src.sharp.service` before the app exists
+makes it a no-op, and the module cache means the later import re-runs
+nothing — `/api/sharp/market` then 404s with no other symptom.
+
 Roster observations live in `sharp_rosters` / `sharp_roster_assets` /
 `sharp_roster_asset_spans` / `sharp_roster_observations`
 (`src/sharp/roster_store.py`), created by a plain
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -116,6 +116,8 @@ npm run regression                   # Full pipeline: preflight + tests
 | `/api/trade/suggestions` | POST | Roster-aware trade suggestions (reads live contract) |
 | `/api/trade/finder` | POST | KTC arbitrage finder |
 | `/api/leagues` | GET | Active league registry (stable `key` → `displayName` + roster settings; **no Sleeper IDs leaked**) |
+| `/api/sharp/roster-percentage` | GET | Sharp Roster Percentage board (global cohort — takes no `leagueKey`) |
+| `/api/sharp/roster-percentage/audit` | GET | Every roster behind one player's count, for manual verification |
 
 ### Rankings vs. league context — the core split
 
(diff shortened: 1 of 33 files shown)


jasonleetucker-code:
Pushed `e35114b` — three coverage gaps, all of which would have shipped a board that looked healthy while omitting part of the cohort.

**Discovery recorded memberships only for expanded leagues.** `src/sharp/discovery.py` wrote a `league_memberships` row only in the league→users branch; the user→leagues branch queued the league but recorded no membership. Leagues left on the frontier by the call budget, `maxGenerations` or `maxLeaguesPerRun` had none. Since the roster collector asks `platform_memberships` which leagues a cohort member plays in, it was silently bounded to the expanded subgraph — a sharp with four dynasty teams could contribute one roster to the denominator and still read as fully represented, because `cohortCoveragePct` counts *managers* seen, not their leagues. Sleeper's own `/user/{id}/leagues` proves the membership, so the row is a fact rather than an inference.

**A budgeted roster pass re-collected the same prefix forever.** `collect_sleeper_rosters` ordered leagues by id, so a run that hit its call budget collected the same alphabetical prefix on every subsequent run and the leagues after the cutoff were *never* collected — a permanently invisible tail, not a slow rollout. Now ordered by `record_queue.prioritize_league_ids`, the same fair ordering the records crawl already uses. Each run also reports `leaguesRemaining` so `--budget` can be sized from the journal instead of guessed.

**Route registration depended on import order.** `src/sharp/service.py` registers its market routes as an import-time side effect; anything importing it before the app exists makes that a no-op, and the module cache means the later import re-runs nothing — `/api/sharp/market` then 404s with no other symptom. `server.py` now calls the public `register_http_routes()` explicitly after importing. This is exactly what `tests/sharp/test_public_api_allowlist.py` was catching, so the pre-existing failure noted in the PR description is now **fixed** rather than merely explained: `pytest tests/sharp/ tests/intel/` is green at 447.

6076 Python tests (up 13) and 1675 frontend tests pass. The ten-check audit still clears.

Remaining items are operational or config-only, and none block merge: the cohort must be non-empty in production before rows appear, trends need two crawl passes before the 7d/30d columns resolve, FFPC needs one roster-bearing URL in `config/sharp/ffpc_sources.json`, and the sharp-vs-market columns stay inert until an ownership feed is registered — there is no such feed in the platform and I did not synthesize one from Sleeper's trending add-counts.

---

PR #714, opened 2026-08-04
Consensus Edge: make the two unmeasured components measurable (and find out one still is not)

jasonleetucker-code:
Started as the docs-only answer to the open `_PERCENTILE_REFERENCE_N` question and grew into the "earn the edge first" work: the engineering that makes the two unmeasured components measurable, so the pre-registered gate has something real to pass. **The `consensus_edge` flag stays OFF.**

Both components carried declared priors for a stated reason. Both reasons were wrong, and correcting them points in opposite directions.

---

## `snapTrend` — measurable now, still unmeasured, for a different reason

Four places said the axis was unbacktestable because `data/playerctx/snapshot.json` is overwritten weekly with no history. The implied fix was a retention project.

**The history was always upstream.** nflverse publishes `snap_counts_{season}.csv` as one row per player *per game* with a `week` column, and appends a dated snapshot to `depth_charts_{season}.csv` on every scrape. `parse_snap_counts` was discarding it with a newest-season filter and no week cutoff. As-of reconstruction was a missing function argument, not a missing dataset.

**But retention would not have helped either**, and that is what decides the work. The panel runs 2026-04-16 → 2026-08-04, entirely between draft and kickoff, so all 111 dates resolve to the same completed season and the same final week:

```
2026-04-16 → AsOf(season=2025, through_week=22)
2026-06-01 → AsOf(season=2025, through_week=22)
2026-08-04 → AsOf(season=2025, through_week=22)
```

`snapTrend` is the identical per-player number on 16 April and on 3 August. Weekly retention across that window would have captured ~16 byte-identical blocks.

What shipped:

- `parse_snap_counts(season=, through_week=)` and `parse_depth_charts(as_of=)`, both defaulting to the previous unbounded behaviour and test-pinned as byte-identical.
- `src/playerctx/asof.py` resolves a date to the window observable on it, from the nflverse schedule already fetched for BDVM's ROS. A week counts only when all its games are done; same-day games do not count, because snaps publish after the game.
- `service.reconstruct_playerctx` replays the snapshot in memory. **Not** a parameter on `refresh_playerctx` — that would leave a historical replay one default argument away from overwriting the file production serves live.
- A `snapTrend` arm on the composite backtest. First run: mean rho **+0.037** over 5 folds, 4/5 positive, beat the market-value benchmark in **0/5** — and it stamps `effectiveSignalObservations: 1` so the fold count cannot vouch for what is one cross-section.

Verified rather than reasoned about:

| check | result |
|---|---|
| week-22 replay of 2025 vs the live unbounded read | **byte-identical**, 1,766 players |
| trend moves with the cutoff | 1766 / 1592 / 1400 players at weeks 22 / 10 / 4 |
| existing composite + momentum arms after the change | **bit-identical** — the arm is purely additive |

**Residual bias, reported not assumed away.** The join anchor is the live-only Sleeper directory, so a player out of the league by run time cannot join. `reconstruct_playerctx` returns per-source join rates (80.7% for the 2025 snap counts) rather than an unqualified player list.

---

## Sharp Flow — the filter is now real, and it still cannot be backtested

`inputs.sharp_movements` documented itself as "qualified-manager trade movements". The query was `WHERE tx_type = 'trade'` and nothing else. The claim held in production only because the crawler visits qualified managers — incidental, and it would fail silently and flatteringly the day anything else wrote to the ledger.

Four defects, all the same shape — code that works until you use it:

- **No cohort filter**, per above.
- **`managerQuality` never supplied**, so every manager weighed 1.0 and the quality term in `aggregate_asset` was a constant.
- **`STATUS_NO_COHORT` declared and unreachable** — a status naming a check no code performed. "A ledger exists but nobody qualifies" read as "no ledger".
- **The query read the per-platform raw id columns.** After the platform migration those keep the *source* ids; on a multi-platform ledger the same player arrives under two asset keys and no manager matches a cohort key.

All four fixed, with the cohort taken from `src.sharp.market.cohort_members` rather than a second definition of who qualifies.

**None of that makes it backtestable.** The corpus is conditioned on *today's* cohort — `crawl_sharp_activity.py` never collected movements from managers who have since dropped out — so it is survivorship-biased on a proxy for the outcome, upstream of every filter. The recorded reason ("unvalidatable until `src/sharp/` gains an as-of cohort") was necessary but **not sufficient**, and stating it alone implied a fix that would not work. A historical Sharp Flow backtest is deliberately not attempted; a number produced that way would be plausible and wrong.

**Adopted instead: the forward-only path that already existed.** `component_sharp_flow` has been snapshotted daily all along, and `label_outcomes` fills cohort-excess returns once the horizon elapses — signal and outcome written weeks apart by two code paths, so nothing is reconstructed and nothing can leak. `scripts/validate_components_forward.py` reads it for every stored component plus the served ranking key, so a Sharp Flow number has something to be judged against.

---

## Bugs the new code exposed

- **Parameter shadowing.** The parse loop bound its row-season local to the name `season`, destroying the caller's argument on the first row so the filter silently used the newest season regardless.
- **Off-by-one-day.** `parse_depth_charts` compared `dt > as_of` on whole strings. `dt` is a full timestamp, so the natural bound `"2025-11-09"` dropped that entire day and fell back to the previous snapshot — while still returning a plausible depth chart. Fixed with a same-length prefix compare; verified the two covering tests fail without it.
- **A field-list coupling that would rot silently.** `identity_join.identity_rows` trims a contract to the fields the joins read, so the backtest can hold a rolling window of days without holding a rolling window of 4 MB contracts. A test pins that the trimmed join equals the full one, so adding a fallback spelling can't quietly drop the join to zero rows.

---

## What this does not claim

Neither component is validated, and this does not move the flag. `snapTrend` becomes a real multi-fold measurement when in-season dates enter the panel — automatically, no further work. Sharp Flow accrues one forward observation per day from the day the board runs. If those come back null, the honest outcome is a documented dead end.

`params_v1.json` edits are comment-only: `paramSetId` is unchanged at `b90ef7db61a10905`, so committed measurements stay attributable. The `consensus-edge-revalidate` gate re-runs board-validation only, which this does not touch.

ADR-026 and ADR-027 in `docs/consensus-edge/DECISIONS.md`.

## One thing to note in the history

The first commit accidentally swept two scheduled-refresh artifacts (`data/scrape_state/sleeper_last_success`, `data/sleeper_last_good.json` — one-line timestamps a local scrape had touched) into the diff via `git add -A`. That is what made the PR conflict with main. Both resolved to main's values in the merge commit; the net diff against main now contains **zero** generated artifacts.

## Tests

Re-run after merging main:

- Python: **6275 passed**, 286 deselected
- Frontend: **1776 passed** (107 files)

_(Pre-merge on this branch alone: 5825 python vs a 5767 baseline, i.e. +58 new tests here.)_


jasonleetucker-code:

This one merged with no conflicts and needed no adjustment — the single documentation line is on `main` verbatim, and `_PERCENTILE_REFERENCE_N` stays 500, which is what the finding itself mandates.

---

PR #719, opened 2026-08-04
Fix two Sharp Roster Percentage defects found by auditing real Sleeper data

jasonleetucker-code:
Follow-up to #710. I ran the whole chain end-to-end against live Sleeper for the first time — discovery (331 managers, 62 leagues, 21 sharp-eligible), records, an 11-manager qualified cohort, the roster crawl, then the validator. Two defects the constructed fixture could not surface.

## 1. The audit reported a false failure on a correct board

`scripts/validate_sharp_roster_percentage.py` hardcoded its own four-element list of position families, omitting `FAMILY_UNKNOWN`. The engine includes it deliberately: a player whose position we cannot determine has a denominator that cannot be *narrowed*, so every roster applies.

Compounding it, the script almost never has a contract — `latest_contract_data` is an in-process global of the running server — and without one **every** player's position is unknown. The two facts combined into a confident `0/10 per-player denominators match` against a board that was right.

- `rp.POSITION_FAMILIES` is now exported and iterated from both sides, so the engine and its independent auditor can't drift again.
- A missing contract is now a **visible failed check**, not a quiet `SKIPPED` tucked inside a `PASS`.
- `--contract PATH` lets an operator actually exercise the position-aware denominator, which is the subtlest rule in the feature and was never being audited.

Check 7 was also asking the wrong question. It failed when our *board* didn't price a player, but the audit item is whether players map to the correct internal ids. Tyrod Taylor — active, rostered by a third of the cohort, outside our top-800 pool — is correct behaviour, not a mapping error. Identity is now checked against the canonical catalog, and unpriced-but-identified players are reported as information.

## 2. A rostered player outside our board rendered as a bare Sleeper id

Same player, second symptom: the page showed **`827`** next to a perfectly correct 33.3%. Our board only names players inside its ranked pool, and nothing hydrated `canonical_assets` on the Sleeper side — the FFPC crawl already did its half.

The roster crawl now fetches Sleeper's directory **once per run** (not per league) and hydrates the catalog; the board consults it before falling back to the tracker's file-based catalog. Best-effort throughout: a directory failure leaves names as they were and never costs a roster observation.

## Verified on real data

12 eligible rosters from 6 sharps, 217 qualifying players, **all 12 audit checks pass at a 40-player sample**:

```
[PASS] contract_available_for_position_aware_checks   937 players joined
[PASS] numerator_matches_underlying_rosters           40/40 recount exactly
[PASS] percentage_is_numerator_over_denominator       40/40 in range
[PASS] one_observation_per_roster_per_player          12 rosters, no dupes
[PASS] denominator_matches_eligible_rosters           40/40 match
[PASS] no_duplicate_roster_rows                       22 rosters, 22 distinct
[PASS] each_roster_has_exactly_one_manager
[PASS] counted_assets_resolve_to_a_known_identity     40/40 resolve
[PASS] board_coverage_of_rostered_players             36/40 priced; 4 unpriced
[PASS] excluded_rosters_never_enter_a_numerator       10 excluded, none counted
[PASS] filters_move_numerator_and_denominator
[PASS] buy_sell_join_agrees_with_the_tracker
```

The most encouraging line isn't in the checks: **kickers measure against only the 3 rosters in leagues with a K slot, not all 12.** That's the per-player denominator working on real league formats, which is exactly what the fixture couldn't prove.

Board output is sane too — real names, `directional` sample warning at 12 rosters, trends correctly withheld (`roster_population_changed`) because only one observation exists so far, exclusions itemised (`incompatible_league_format: 6, incomplete_roster_data: 5, league_not_sharp_eligible: 4`).

6134 Python tests pass.

## Still open, unchanged

No general-dynasty ownership feed exists, so the sharp-vs-market columns remain null. FFPC still contributes zero rosters until a roster-bearing URL is configured. Both are config/data, not code.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Btu1kVCWeT6FkSE2HcMKUn)_


jasonleetucker-code:

This one merged with no conflicts and needed no adjustment. Both fixes are on `main` as written — `POSITION_FAMILIES` exports all five families including `FAMILY_UNKNOWN`, and the Sleeper catalog hydration is intact.

---

PR #720, opened 2026-08-04
Audit tranche 2: drifted mirrors, a hollow guard, and docs that misdescribe live behaviour

jasonleetucker-code:
Continues the unfalsifiable-number audit after #704. Five commits — and **two of my own findings retracted** after measurement, which is as much of the result as the fixes.

| tier | result |
|---|---|
| `pytest -m "not livedata"` | **6027 passed**, 0 failed |
| `pytest -m livedata` | **252 passed**, 25 skipped |
| `npx vitest run` | **1756 passed**, 105 files |
| both CI ruff gates | clean |
| board fingerprint | **byte-identical** on the pinned payload |

## Fixed

**A guard that asserted a type, on the wrong key, comparing nothing.**
`test_legacy_dict_mirrors_the_row_value` read `rdv` from the row and never used it; its only assertion was `isinstance(mirrored, (int, float))` on `_finalAdjusted`. A mirror off by any amount passed.

The wrong-key half matters as much as the missing comparison: `_finalAdjusted` is the legacy scraper **composite**, a different scale from `rankDerivedValue` — it equals the row value on **1 of 812** rows. So fixing only the assertion would have produced a guard that failed on correct data. The dict does carry `rankDerivedValue`, on all 812 rows, matching **812/812**. That's the invariant the test's name always claimed.

**9 pure tests returned to the blocking CI gate.** `test_dlf_scraper.py` sat in `_LIVEDATA_MODULES` despite its own docstring saying it "exercises the offline-safe pure functions". Its single `CSVs/site_raw/` mention asserts a config *string*. 9 tests, 0.14s, no data files — and before this, a regression in DLF's HTML parse or paywall detection (the ingestion path for four registry sources) could not fail a PR.

**Three docs claimed every feature flag defaults OFF** — including `feature_flags.py`'s own docstring, the file the other two point at. **7 of 15 default ON**, including `te_basis_conversion`, which reprices every tight end on the live board.

**`IDP_CONSENSUS_KEYS` had drifted both ways at once** — omitted `dlfRookieIdp` (a real source; 28 of 29 affected rows shift their consensus mean by ≥1 rank) while its comment named "FootballGuys IDP", which exists in neither registry.

**Two mislabels** — `effectiveWeight` shown as "Weight" when the contract says it's "never applied to the blend"; `EDGE_PREMIUM_RANK_LIMIT` documented as 150 when it's 250.

## Retracted after measurement

**`values.full` falling back to the raw composite.** All 281 unpriced rows produce **zero** through the frontend chain — `displayValue`, `finalAdjusted` and `overall` are all null for them. A prior math-audit fix covered both paths. The reported `2026 Early 1st → finalAdjusted: 6136` does not reproduce.

**The corridor clamp's "dead" percentile band.** The measurement holds — 128/128 clamped rows use the flat 0.15 cap. The *conclusion* was wrong, and writing the guard is what refuted it: a synthetic fixture with narrower source disagreement yields `bandPct 0.0217, cappedByMaxBand False`. The percentile block is live machinery that today's unusually wide IDP drift dominates, not code that cannot run — a different thing from the `coverageWeight` precedent. It ships as a corrected comment with **no test**, because a test pinning "the cap always wins" would assert a fact about market drift and go red the first time sources agree more closely.

## Guards are built to resist re-baselining

The flag guard asserts the **claim, not the count** — pinning "7" would go red whenever someone adds a flag, and the cheapest fix would be bumping the number, which is the green-by-construction habit ADR-008 exists to stop. The `IDP_CONSENSUS_KEYS` guard derives its expectation from `RANKING_SOURCES`. Both carry non-vacuity assertions.

**My first draft of the flag commit said "8 of 15" and listed `consensus_edge` among the enabled flags.** I took it from a survey rather than reading `_DEFAULTS`; `consensus_edge` defaults OFF. The guard caught it on its first run, before push.

## Mutation-tested

- Restore `(default OFF)` to ARCHITECTURE.md → red
- Remove `dlfRookieIdp` again → 2 assertions red
- Shift every mirrored value by +7 → red with `dict=9653 row=9646`

## Deliberately not changed

`README.md` prints the live Sleeper league ID. It appears in **19 files** including tracked config, exports and git history, and `SECURITY.md` records the public visibility as deliberate "with the exposure understood". Redacting one line would remove it from nowhere. That needs a real decision — rotate, accept, or scrub history — not a line edit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,11 @@
 # Risk It To Get The Brisket — Dynasty Trade Calculator
 
-Private repo for the dynasty trade calculator stack powering
-[chaseupside.com](https://chaseupside.com).
+Public repo for the dynasty trade calculator stack powering
+[chaseupside.com](https://chaseupside.com). Visibility is a deliberate
+owner decision — see `SECURITY.md`, which records what that exposes
+(valuation engine, league configuration, tracked `data/` + `exports/`
+snapshots). This line said "Private repo" while `SECURITY.md` and
+`HANDOFF.md` both said public.
 
 **Architecture at a glance:**
 - **Backend:** Python 3.12 FastAPI + Uvicorn (port 8000).
@@ -12,7 +16,12 @@ Private repo for the dynasty trade calculator stack powering
 - **Auth model:** session-gated `/api/*` endpoints with a
   Sleeper-login allowlist; public routes hit `/api/public/league/*`.
 - **Feature flags** (`src/api/feature_flags.py`): every new
-  capability from the 2026-04 upgrade ships flag-gated (default OFF).
+  capability from the 2026-04 upgrade ships flag-gated. Defaults are
+  **per-flag**: 7 of the 15 in `_DEFAULTS` ship enabled (`bdvm_engine`,
+  `te_basis_conversion`, `monte_carlo_trade`, `idp_scoring_fit`,
+  `reception_scoring_fit`, `nfl_data_ingest`, `realized_points_api`),
+  several deliberately. For those the env var is a rollback lever, not an
+  opt-in. Read `_DEFAULTS` before assuming a capability is dormant.
 
 **Orientation docs:**
 - [`docs/ARCHITECTURE.md`](docs/ARCHITECTURE.md) — system map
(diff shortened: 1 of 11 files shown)


jasonleetucker-code:

This one merged with no conflicts and needed no adjustment — `dlfRookieIdp` is in `IDP_CONSENSUS_KEYS`, the SourceAuditPanel shows "Weight (applied)" ahead of "Coverage wt (diagnostic)", and both new tests are on `main`.

One thing this PR surfaced but did not fix, carried forward in #723's description so it isn't lost: `src/api/compact_view.py::_SLIM_SOURCE_RANK_META_FIELDS` omits `appliedWeight`, so on the compact view the corrected row does not render while the diagnostic one still does — the same inversion, one surface over.

---

PR #726, opened 2026-08-05
Run the E2E safety net on PRs that touch the app shell

jasonleetucker-code:
Closes the gap that let #709 ship a bug rendering every page twice.

## Why

`e2e.yml`'s header said the suite deliberately does not run on PRs, and gave a good reason: it boots the whole stack and drives a real browser, which is too slow for every change. **That reason still holds for most changes.** It did not hold for the one class of change that just got through.

#709 added a `dynamic()` import to `components/AppShell.jsx`. `dynamic()` is `React.lazy`, which needs a Suspense boundary; with no local one the nearest ancestor is the App Router's, and `{children}` — the entire page — renders inside it. Every route's content became deferred streaming content: staged in a hidden `<div id="S:1">`, moved into place, and the staged copy left behind. **`/waivers` served three `<main>` elements** — the shell's, the page's, and a hidden second full copy of the page.

Nothing else in the repo saw it. The duplicate has no client rects and never reaches the accessibility tree, so it is invisible in a screenshot and to a human clicking around. Six rounds of performance instrumentation — CLS, LCP, INP, long-task/TBT, scroll FPS, per-route bundle size — **all reported unchanged numbers** for a build that rendered every page twice. `pr-validation.yml` runs pytest, vitest and lint; it never opens a browser. The sole detector was a Playwright strict-mode violation in `waivers-smoke.spec.js`, in this workflow, which ran nightly and so caught it a day late on a branch nobody was watching.

## What changed

**1. A path-filtered `pull_request` trigger.** The paths are the SSR/module-graph surface — root layout, shell composition, the auth gate, bundler and framework config, plus the suite itself:

```
frontend/app/layout.jsx          frontend/middleware.js
frontend/app/AppShellWrapper.jsx frontend/lib/public-routes.js
frontend/components/AppShell.jsx frontend/package.json
frontend/components/shell/**     frontend/next.config.*
tests/e2e/**                     .github/workflows/e2e.yml
```

`AppShell.jsx` and `package.json` are both on that list, so **this trigger would have caught #709 at review time.**

`frontend/components/ds/**` is deliberately **excluded**, and the file says why: the design system is touched by a large share of frontend PRs, so including it would quietly turn this into the every-PR gate the header rules out. Dispatch the workflow by hand for a risky `ds` change.

**2. Per-ref concurrency.** The group was a single global `e2e-safety-net` with `cancel-in-progress: true`. The first PR run would have cancelled the nightly *and* every other PR's run — and a cancelled run reports the same "not green" as a failure while having tested nothing. Now keyed on `github.event.pull_request.number || github.ref`.

**3. Both tracking-issue steps require main AND a non-PR event.**

```yaml
if: failure() && github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
if: success() && github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
```

The two clauses guard different things and neither subsumes the other in intent:

- The **ref** clause is about branch *dispatch*, not PRs. Dispatching this workflow on a branch is a first-class use — it is how a branch gets arbitrated against the green baseline — but its result must not speak for `main`. That is #716.
- The **event** clause states the PR-run intent at the point of use. A PR failure is already visible on the PR's own checks; commenting on the tracker would bury the real signal under one comment per red PR run, the exact "permanently red" noise the close step was added to end. It is redundant with the ref clause today (a PR run's ref is `refs/pull/<n>/merge`), and it is kept so that adding a `push:` trigger later cannot silently make PR runs eligible.

The close step is the more dangerous of the two in both directions: a false all-clear closes a real open failure, while a false alarm merely files a wrong one.

*(First revision of this PR replaced the ref guard with the event guard instead of conjoining them. That was unnecessary — the ref check already excluded PR runs — and it re-opened #716, since `workflow_dispatch` on a branch satisfies `!= 'pull_request'`. Fixed in the second commit.)*

## Expect intermittent reds, and read them before blaming the PR

This is documented in the workflow header. The app has a **pre-existing** SSR-streaming race in which the same `#S:1` staging copy is occasionally left behind — measured at **1/15 loads on `/arbitrage` and 1/45 on `/waivers`, identical rates on a build of `main` and on the #709 branch**. It is `main`'s defect, not any one PR's, and is tracked separately in #716.

Before attributing a red run here to the PR under review, reproduce the rate against a build of the base branch. **Do not "fix" it by loosening those locators** — they are the only detector this repo has for a bug class every performance metric is blind to.

## Testing

- New `tests/deploy/test_e2e_workflow_triggers.py` (8 tests) `yaml.safe_load`s the workflow and pins the three triggers, the ten path globs, the deliberate *absence* of `components/ds/**`, the per-ref concurrency expression, and the ref guard on both issue steps. Each assertion was verified to fail against the form it guards against before being kept. It lives beside the existing workflow-wiring tests in `tests/deploy/` and runs in `pr-validation.yml` with the rest of pytest — nothing in the repo referenced `e2e.yml` before this.
- No change to the job body — same steps, same stack boot, same Playwright invocation. Only triggering, concurrency and the two issue-step guards moved.
- This PR touches `.github/workflows/e2e.yml`, which is on its own path list, so the new trigger exercises itself here.

---

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -56,7 +56,13 @@ name: E2E Safety Net
 # Failure handling mirrors scheduled-refresh.yml: an idempotent
 # tracking issue labelled ``e2e-failures`` is opened (or commented
 # on) so a broken journey can't sit unnoticed between manual runs.
-# Both issue steps are scoped to NON-PR runs — see the guards on them.
+#
+# That tracker is about MAIN, and only main writes to it.  Both issue
+# steps require `github.ref == 'refs/heads/main'` AND a non-PR event —
+# the ref clause keeps a branch DISPATCH from speaking for main (#716),
+# the event clause keeps a PR run from doing the same.  Neither clause
+# subsumes the other in intent, so both are stated; see the notes on the
+# steps themselves.
 
 on:
   schedule:
@@ -309,13 +315,27 @@ jobs:
         # on the open ``e2e-failures`` issue if there is one, else
         # open a single labelled issue and keep reusing it.
         #
-        # NON-PR RUNS ONLY.  This issue tracks "is main's safety net
-        # broken".  A PR run failing says something about the PR, not
-        # about main, and the failure is already visible on the PR's own
-        # checks — commenting here would bury the real signal under one
-        # comment per red PR run, which is exactly the "permanently red"
-        # noise the close-step below was added to end.
-        if: failure() && github.event_name != 'pull_request'
+        # MAIN ONLY, and the two clauses below guard DIFFERENT things —
+        # do not collapse them into one.
+        #
+        # The ref clause is about BRANCH DISPATCH, not about PRs.
+        # Dispatching this workflow on a branch is a first-class use (it
+        # is how a branch gets arbitrated against the green baseline),
+        # and its result is reported in the run itself.  It just does not
+        # get to speak for main.  See #716, and see the close step below
+        # for why that one is the dangerous direction.
+        #
+        # The event clause states the PR-run intent at the point of use.
+        # A PR run failing says something about the PR, not about main,
+        # and the failure is already visible on the PR's own checks —
+        # commenting here would bury the real signal under one comment
+        # per red PR run, which is exactly the "permanently red" noise
+        # the close step below was added to end.  It is redundant with
+        # the ref clause TODAY (a `pull_request` run's ref is
+        # `refs/pull/<n>/merge`, never `refs/heads/main`), and it is kept
+        # so that adding a `push:` trigger later cannot silently make PR
+        # runs eligible.
+        if: failure() && github.ref == 'refs/heads/main' && github.event_name != 'pull_request'
         shell: bash
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -323,7 +343,7 @@ jobs:
           set -Eeuo pipefail
           RUN_URL="${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
           BODY=$(cat <<EOF
-          Nightly E2E safety-net run failed.
+          E2E safety-net run failed on \`main\` (trigger: ${GITHUB_EVENT_NAME}).
 
           - Workflow run: ${RUN_URL}
           - Commit: ${GITHUB_SHA}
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
**E2E is red here, and it is `main`'s — arbitrated against the baseline rather than assumed.** `Validate PR` is green.

The workflow header asks that a red run be reproduced against the base branch before it is attributed to the PR, so I dispatched the suite on `main` ([run 30970233761](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/30970233761), `f2029b1e`). It failed:

| | This PR (`4642a7cd`) | `main` (`f2029b1e`) |
|---|---|---|
| `journey-perfect-draft.spec.js:75` — buyable combination of real rookies | ❌ | ❌ |
| `journey-perfect-draft.spec.js:109` — never spends more than the budget | ❌ | ❌ |
| `journey-trade.spec.js:145` — `/arbitrage` renders trades or empty state | ✅ | ❌ |
| `waivers-smoke.spec.js:36` — rookie toggle | flaky (retried green) | flaky (retried green) |

This PR's failures are a strict **subset** of `main`'s, and the extra one on `main` is `/arbitrage` — the pre-existing `#S:1` streaming race at its documented ~1/15 rate (#716). `main` opened #732 for its own failure.

Why the perfect-draft failures are new to CI but not new to `main`:

| | |
|---|---|
| `journey-perfect-draft.spec.js` reached `main` via #723 | 2026-08-05 **00:04Z** |
| Last E2E run on `main` before today | 2026-08-04 **19:54Z** |

The spec landed ten minutes *after* the last run that could have exercised it, so **it has never run against `main`**. The path-filtered trigger this PR adds is the first thing to execute it — which is the trigger doing exactly what it was added for, on its first outing.

The diff here is one YAML file the application never reads, so the app under test was byte-identical to `main`'s. I did not touch the perfect-draft locators.

What I checked before concluding it is not a plain endpoint failure: against a live backend, `GET /api/draft/roster-context?leagueKey=dynasty_main&teamName=…` returns a full context (`waiverLadder` 90 rungs, `cutLadder` populated, `openRosterSpots` 1), `GET /api/draft-capital` returns 72 picks all carrying `rookieBoardValue`, and the panel's 145 unit tests pass. The panel returns `null` on `!result`, so the remaining suspect is the solve going empty under the committed E2E snapshot specifically. That is `main`-side work and does not belong in this PR.

### Also fixed here since review opened

I introduced a regression in the first commit and it is worth calling out rather than burying in the diff. To stop the tracker steps firing on PR runs I *replaced* their guard instead of extending it:

```diff
- if: failure() && github.ref == 'refs/heads/main'
+ if: failure() && github.event_name != 'pull_request'
```

Unnecessary — a `pull_request` run's ref is `refs/pull/<n>/merge`, so the original guard already excluded PR runs. And regressive: `workflow_dispatch` on a branch satisfies `!= 'pull_request'`, so a green branch dispatch would once again close an open `e2e-failures` issue and issue `main`'s all-clear from code never on `main`. That is #716. Both guards are now the conjunction, and `tests/deploy/test_e2e_workflow_triggers.py` pins the triggers, the path list (including the deliberate absence of `components/ds/**`), the concurrency group and both ref guards — verified to fail against the broken form before being kept. Nothing referenced `e2e.yml` from a test before this.

---

Issue #730, opened 2026-08-05
Streaming SSR intermittently leaves its #S:1 staging copy in the DOM (page markup renders twice)


jasonleetucker-code:
Filing this because it currently has **no tracker**. It was diagnosed inside #716 — the bot-filed `e2e-failures` tracking issue — and #716 was then closed, so the diagnosis is buried in a closed issue and the automation will only file again on a *failed* `main` run.

## Mechanism

React's streaming SSR writes deferred content into a hidden `<div id="S:1">` staging container and then moves it into place. Occasionally the staged copy is **left behind**, so the page's markup exists twice in the DOM.

The duplicate has no client rects and never reaches the accessibility tree. It is invisible in a screenshot, in an a11y snapshot, and to a human clicking around. It is still duplicate DOM and duplicate element ids.

## Rate

Measured by the session that diagnosed it (reported on #716 — **not independently reproduced by me**), loading each route repeatedly and counting loads that render two copies. Builds of `main` and of a feature branch:

| route | needle | `main` | branch |
|---|---|---|---|
| `/arbitrage` | "Pick a team and scan" | 1/15 | 1/15 |
| `/waivers` | "Include rookies" | 1/45 | 1/45 |

**Identical on both**, which is what makes it ambient on `main` rather than any one PR's regression.

## It is not AppShell

Worth recording, because AppShell is where the last investigation landed and it was a different bug. PR #709 briefly made this fire on *every* load of *every* route by using `dynamic()` in `AppShell` — `React.lazy` needs a Suspense boundary, and with no local one the nearest ancestor is the App Router's, which encloses `{children}`. That was fixed there by loading imperatively via `await import()`.

The ambient race is separate and **survives static imports**. What remains in play is the App Router's own per-segment boundary, created by the 12 `loading.jsx` files — including `frontend/app/arbitrage/loading.jsx` and `frontend/app/waivers/loading.jsx`. Neither affected page uses `useSearchParams`, and `next.config.mjs` sets no `experimental`/PPR flags. Versions: `next 16.2.12`, `react 19.2.8`.

## The only two detectors

Playwright strict-mode violations ("resolved to 2 elements"):

- `tests/e2e/specs/journey-trade.spec.js` — `getByText(/Pick a team and scan/i)`
- `tests/e2e/specs/waivers-smoke.spec.js` — `getByLabel(/Include rookies/i)`

[...]

PR #733, opened 2026-08-05
Make a green E2E run mean the product is healthy

jasonleetucker-code:
#718 scoped the auto-issue trackers to mainline runs — it fixed **who** may speak for `main`. It did not fix **how much one green run proves**, which I said on #716 rather than leave the tracker looking sound. This is that half.

Three independent ways a green run could mean *"the suite didn't look"*, all `docs/ORCHESTRATION.md` §6.15's class. Each verified against `origin/main` before being touched.

## 1. A retried failure read as green — and closed the tracker

`playwright.config.js` — `retries: process.env.CI ? 1 : 0`. A test can fail on attempt 0, pass on attempt 1, and leave the run's status at `passed`, exit 0. Grepping every workflow and all of `tests/e2e/`: **nothing read the flaky count.**

So `e2e.yml`'s close step retired the `e2e-failures` issue on a run that contained failures — draining *every* open one, since it iterates `.[]` — under its stated premise that *"a green run of THIS workflow is the all-clear by construction."*

Fixed with `failOnFlakyTests`. **The predicate is in the config, not in `stack-death-reporter.js`, and that placement is the entire point:** a `--reporter=` flag replaces the reporter array and unloads every guard in that file, so putting it there would have rebuilt the exact defect fixed in §2 below. A CLI flag cannot unload a config key, and `--fail-on-flaky-tests` can only set it *true*.

Proven rather than assumed — scratchpad run, no stack needed:

| case | result |
|---|---|
| `failOnFlakyTests` present, flaky test | exit **1** |
| key **removed**, same flaky test | exit **0**, banner still prints |

That second row is the proof that the config key — not my banner — is what fails the run.

`retries` stays at 1: the retry is what produces `trace: "on-first-retry"`, and the defect was never the retry, only its silence.

## 2. The prod smoke unloaded its own guards

`prod-e2e-smoke.yml` passed `--reporter=line`, replacing the config's reporter array and unloading `stack-death-reporter.js` — the stack-death abort **and** the coverage floor — for every production smoke run it has ever made. The reporter's own header warns about this in capitals.

The trap in the obvious fix: the floor defaults to 100, but that job runs one spec. Measured on run `30953000224` — `32 passed`, tests `[1/32]`–`[32/32]`, **zero skipped**. Bounds now set to match. Side effect: `playwright-report/` is populated, so the failure-artifact step stops uploading an empty directory.

## 3. A product degradation became a green skip

`journey-settings-overrides.spec.js` skipped when the custom-mix badge never appeared and the app had logged its base-contract fallback — measured firing *routinely*. That is what `tests/e2e/README.md` convention 5 forbids: *"Skip cleanly on absent INFRA — never on absent DATA."*

Replaced with a **more** specific assertion, not a weaker one: `/rankings` fires its own overrides POST on hydration, and that second round-trip is what the badge depends on and what the skip was forgiving. A failure now says *"the second overrides POST returned 503"* instead of *"badge missing, cause unknown."* It also recovers `guard.assertClean()` — `test.skip()` throws, so on the skip path every console error on that journey was discarded too.

## Two live defects found while verifying

- **`Number("3O")` is `NaN`, and both `passed < NaN` and `skipped > NaN` are `false`** — one typo in a workflow `env:` block would silently disable the coverage floor while its banner claimed otherwise. Unreachable while nothing set those vars; §2 sets them by hand in YAML. `coverageBound()` now refuses to run instead.
- **The reporter's documented baseline was wrong.** It claimed "~149 passed / 29 skipped". Run `30945387957` measured **139 passed / 49 skipped** — skips had drifted to 49 against a ceiling of 60 and nothing fired, and a floor of 100 let 39 tests vanish silently. Baseline corrected and cited; floor raised to 120.

## Guards, so this cannot quietly come back

In `test_e2e_harness_guards.py`, which runs in the **fast pytest gate** — so these fire on PRs even though the suite itself does not:

- no workflow may pass `--reporter`
- no workflow may set `E2E_ALLOW_FLAKY` / `E2E_NO_STACK_GUARD`
- `retries` and `failOnFlakyTests` must stay paired (non-zero CI retries without the key is the silent regression)
- a subset run must declare numeric bounds; a full-suite run must not set them at all

Per this file's own doctrine, **each was verified by reintroducing its defect and watching it fail** — 5 for 5, then green again.

## Expected consequence, stated up front

The nightly will now go red **roughly one night in ten**, because there is a real intermittent defect: React streaming leaving its `<div id="S:1">` staging copy in the DOM, so page markup exists twice. Filed as **#730** — it had no tracker at all, because it was diagnosed inside #716 which was then closed.

Two honesty notes: the per-load rates behind that estimate are another session's measurement reported on #716, **not reproduced by me**; and the estimate is derived from them, not itself a measurement.

That noise is the point. The alternative is the current state, where the same defect fires just as often and the workflow reports success.

## Verification

- `ruff check .` clean; `ruff format --check` clean (884 files)
- `node --check` on all three JS files; `yaml.safe_load` on the workflow
- `test_e2e_harness_guards.py`: 13 passed
- 5/5 defect-reintroduction checks fire, then revert to green
- 4/4 Playwright self-test cases (flaky→1, key-removed→0, coverage floor→1, NaN→named error)
- full `pytest -m "not livedata"` running; I'll report the result on this PR

## Out of scope

Fixing the `#S:1` race (#730) — likely a React 19.2 / Next 16.2 streaming bug and open-ended; this makes it *visible*, which is the prerequisite. And `E2E Safety Net` still cannot be a required check: it has no `pull_request` trigger by deliberate design. Worth noting the repo has **no branch protection at all**, so not even `Validate PR` is required — owner action OA-02.

---

### DIFF
--- a/.github/workflows/prod-e2e-smoke.yml
+++ b/.github/workflows/prod-e2e-smoke.yml
@@ -87,6 +87,24 @@ jobs:
         shell: bash
         env:
           E2E_BASE_URL: ${{ steps.url.outputs.url }}
+          # This job runs ONE spec, not the full suite, so the reporter's
+          # default floor — sized for the nightly's 139 passed — would
+          # fail every prod smoke run.  Measured on run 30953000224
+          # (5349fea29, 2026-08-04 21:35Z): "32 passed (1.3m)", tests
+          # numbered [1/32]..[32/32], zero skipped = 16 tests x 2 chromium
+          # projects.
+          #
+          # 30 rather than 32: a floor equal to the exact count fires on
+          # any legitimate consolidation of two tests, and a floor that
+          # gets "adjusted" every quarter stops being believed.  30 still
+          # catches what this exists for — a spec-path typo or a project
+          # filter matching nothing takes the count to 0 or 16, not 31.
+          E2E_MIN_PASSED: "30"
+          # Zero, and it CAN be zero: public-league.spec.js has no
+          # conditional skips left — its three data-condition gates were
+          # converted to assertions.  Any skip here means a gate came
+          # back, which is exactly what we want to hear about.
+          E2E_MAX_SKIPPED: "0"
         run: |
           set -Eeuo pipefail
           # Chromium desktop + mobile viewport.  We intentionally skip
@@ -94,12 +112,27 @@ jobs:
           # require ``playwright install webkit`` and don't buy us
           # anything over the chromium mobile viewport for the public
           # /league layout coverage.
+          #
+          # NO `--reporter` FLAG.  It REPLACES playwright.config.js's
+          # whole reporter array, which unloaded stack-death-reporter.js
+          # and with it BOTH the stack-death abort AND the coverage floor
+          # — for every prod smoke run this workflow made before
+          # 2026-08-05.  The reporter's own header warns about this in
+          # capitals; this workflow did it anyway, which is the same
+          # never-fires class the guard was written to catch.
+          # tests/e2e/test_e2e_harness_guards.py now fails the PR if any
+          # workflow reintroduces the flag.
+          #
+          # Side effect worth knowing: the config's reporters are `list`
+          # + `html`.  `list` is what `--reporter=line` was approximating,
+          # and `html` now actually populates tests/e2e/playwright-report/,
+          # which the failure-artifact step below has been uploading empty
+          # since it was written.
           npx playwright test \
             -c tests/e2e/playwright.config.js \
             --project=desktop-1366 \
             --project=mobile-chromium \
-            tests/e2e/specs/public-league.spec.js \
-            --reporter=line
+            tests/e2e/specs/public-league.spec.js
 
       - name: Upload Playwright trace artifacts on failure
         if: failure()
--- a/tests/e2e/README.md
+++ b/tests/e2e/README.md
@@ -172,6 +172,24 @@ pass; the journey specs above are their Next.js-era replacements.
    (`screenshot: only-on-failure`, `video: retain-on-failure`,
    `trace: on-first-retry`) — check `test-results/` or the CI
    artifact.
+8. **A flaky pass is a failure.**  `retries: 1` in CI exists for the
+   trace, not for a second opinion: `failOnFlakyTests` in
+   `playwright.config.js` fails any run where a test passed only on
+   retry.  Before that key landed, Playwright reported such a run as
+   `passed` / exit 0 and nothing read the flaky count — so `e2e.yml`'s
+   close step would retire the open `e2e-failures` issue on a run that
+   had watched a critical journey fail.
+   The key lives in the config, not in `stack-death-reporter.js`,
+   because `--reporter=…` on the command line replaces the reporter
+   array and unloads every guard in that file; a CLI flag cannot unload
+   a config key.  The reporter prints the explanation and deliberately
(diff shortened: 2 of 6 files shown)


jasonleetucker-code:
Ran the two branch dispatches the plan called for. One passed, one failed, and the failure is worth reading carefully because **it is not caused by this PR** — and it corrects a number I put in the description.

## prod-e2e-smoke on the branch — ✅ success

[Run 30971429447](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/30971429447). This validates §2 end-to-end: reporters restored, and `E2E_MIN_PASSED: "30"` / `E2E_MAX_SKIPPED: "0"` are correctly sized. A wrong floor would have failed the run, which is exactly why this had to be measured before the 4-hourly cron inherited it.

## E2E on the branch — ❌ failed, and that is the informative part

[Run 30971427996](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/30971427996): **4 failed / 1 flaky / 52 skipped / 137 passed (8.3m)**.

**The mechanism works.** The banner fired on a real runner and named the flaky test:

> `desktop-1366 › waivers-smoke.spec.js › signed-in: /waivers page › rookie toggle is present and toggleable`

**§3's change passed.** `journey-settings-overrides` is absent from the failure list — so removing the degradation skip and asserting the `/rankings`-side overrides round-trip holds on the runner. That was the one open question about §3, and it is now measured rather than argued.

### Why the red is not this PR's

`main` at **`f2029b1e5` — this PR's exact base commit — is already red.** [Run 30970233761](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/30970233761) failed and opened **#732**, before any of this branch's changes existed.

Breaking down the five problems in my branch run:

| failure | mine? |
|---|---|
| `journey-perfect-draft.spec.js:75` | **no** — spec added to `main` after my last green baseline; this PR does not touch it |
| `journey-perfect-draft.spec.js:109` | **no** — same |
| `journey-trade.spec.js:145` (`/arbitrage`) | **no** — the #730 `#S:1` race |
| `waivers-smoke.spec.js:36` \[mobile] | **no** — #730, hard-failed *both* attempts |
| `waivers-smoke.spec.js:36` \[desktop] | the flaky one — newly fatal, which is the intended behaviour |

**Four of the five are hard failures**, red on both attempts and completely independent of `failOnFlakyTests`. They would have failed this run with or without this PR. Only the fifth is newly fatal.

### Correcting my own estimate

The description says the nightly will go red *"roughly one night in ten"*. **The observed data contradicts that and I am not leaving it to stand.** The `#S:1` race hit **three** tests in a single run — `/waivers` on both projects plus `/arbitrage` — not the ~1-in-10 that the per-load rates on #716 imply. Those rates were another session's measurement, which I flagged as unreproduced; this run is evidence they understate it materially on a loaded runner.

The honest statement is: **the suite is red now, on `main`, for reasons that predate this PR**, and #730 is a bigger problem than its filed rates suggest.

### Why this still merges

It does not make `main` redder. #732 already tracks `main`'s failure, filed by the mainline-only guard from #718 — and note its body reads *"failed on `main` (trigger: workflow_dispatch)"*, which is that PR's wording fix working as designed. What this PR changes is that the **one laundered case stops being laundered**; the other four were always going to be red.

Holding it back would keep a retried failure reading as a clean pass while the tracker gets closed on it — which is the entire defect.

---

PR #741, opened 2026-08-05
Name the streaming duplicate correctly, unblock 28 dead recap links, and prepare the operator items

jasonleetucker-code:
Three things, in one branch because they share one root question: what is actually broken on `main`, and who can fix it.

---

# 1. The streaming duplicate (#730) is not a product defect

#730 and the flaky banner I shipped in #733 both call it one. **That is wrong**, and the wrong name was sending every reader after the wrong bug — including me.

Read from the pinned `react-dom@19.2.8` source, not inferred: **React 19.2 made the Suspense reveal deferred.** `$RC` no longer moves the staged content and deletes the container in one tick — it marks the boundary `"$~"`, queues the pair on `$RB`, and schedules `$RV` via `requestAnimationFrame`, thereafter a `setTimeout` throttled to `$RT + 300 - now` (up to **2300ms** in one window). `$RV` is the only thing that removes `<div hidden id="S:n">`.

So a full copy of the boundary's subtree is **supposed** to be in the DOM for that window. No app change removes it; no user sees it.

**Reproduced, not argued** — chromium under `Emulation.setCPUThrottlingRate`, a loaded CI runner by other means:

| | |
|---|---|
| boundary marker `"$~"` | **30/30** loads (unthrottled) |
| staging container present | **20/25** (8× throttle) |
| `<main>` == 3 **and** toggle → 2 elements | **2/25** — the CI symptom |
| same loop after `waitForStreamSettled()` | **0/25** |

The fix is `waitForStreamSettled()` in `helpers/journey.js`. It is **not `.first()` in disguise**, verified in both directions: control passes; injecting a permanent duplicate (cloning the page `<main>`) makes the wait return at once and the strict locator **still fail**. A permanent duplicate is a real defect — it is what #709's `dynamic()` produced — and those two locators remain its only detector.

**Latent test bug this exposed.** `public-league.spec.js` picked its mobile/desktop branch with `await mobileSelect.isVisible().catch(() => false)`. When the staged copy duplicated the `<select>`, the strict-mode violation was **swallowed by the `.catch`**, "could not tell" became "desktop", and the test walked the desktop branch at 390px where `.desktop-only` is `display:none`. All twelve buttons were then legitimately absent → `missing tabs: <all 12>`, which reads like the nav was deleted. That is the nightly's mobile-chromium flake. Now settles the stream and decides from the viewport, which cannot throw.

Corrected the misattribution in all five places it had spread to.

**Investigated and deliberately not changed:** `sw.js`. The theory was that caching HTML navigations could freeze a mid-stream document into a permanent duplicate. It cannot — `cache.put()` consumes the cloned body to completion. I had written that change before checking; reverting it is the honest outcome.

---

# 2. The archive advertised 28 recap links it could not build

**Live on production** — `prod-e2e-smoke` red since 06:41Z.

`list_matchups` admits a pair on "two entries and either side scored" and never checks owner attribution. `_side_block` returned `None` whenever `resolve_owner` came back empty, and `build_matchup_recap` bails if either side is `None`, so `server.py` 404s silently. `identity.py` deliberately does not register rosters whose owner is orphaned or retired — its own comment says such data "falls through to the same `""` attribution path … and filtered out". **Nothing filtered it.** The asymmetry is the defect, not the retirement list.

Measured against live Sleeper: **28 of 158**, all 2024, roster_ids 9 and 10.

Fixed by making the *detail* tolerant rather than making the index hide games — real 2024 matchups stay reachable, attributed to the roster instead of to a person. Retired managers still stay out of dropdowns and franchise pages, because that filtering lives in `identity.py`, untouched.

| after the change | |
|---|---|
| index advertises | 158 |
| detail returns `None` | **0 of 158** (was 28) |
| sides with fallback | 2 → `Former manager` / `Team 9`, `Team 10` |

The synthetic owner id is unique per roster **on purpose**: the UI marks the winner with `winnerOwnerId === side.ownerId`, so two retired owners facing each other would both compare equal on `""` and both render as the winner. `ownerResolved: false` gates the "Franchise page →" link, which would otherwise trade one dead link for another.

Regression test asserts the invariant over the **whole index**, not a sample — the failing subset was one season/roster pair a sample would miss — and guards the other direction too. Verified by reintroducing the `return None`.

---

# 3. The operator items: everything up to the credential

None can be closed from a session. Two register bugs turned up while checking.

**Bug: OA-02 named the wrong checks.** It said requiring `E2E Safety Net` or `Deploy Production` would deadlock every PR. Those are **workflow** names; GitHub matches **job** names. Neither string would ever match a check run, so a ruleset requiring them would block every PR forever waiting for a check that cannot report. Real names: `Validate PR`, `Boot stack + run journeys`, `Validate Build Inputs`, `Deploy To Production`.

**Bug: branch protection was missing from `UNIMPLEMENTED_BACKLOG` §10** — it lived only in the audit, so the two registers disagreed, and §10 is the one `HANDOFF.md` points at.

| item | prepared | still needs you |
|---|---|---|
| branch protection | `deploy/github/main-protection-ruleset.json` + `apply-branch-protection.sh` (idempotent, `--dry-run`) | an admin-scoped `GH_TOKEN` |
| cert renewal | `deploy/verify-cert-renewal.sh` | `sudo` on the VPS |
| token rotation | `docs/runbooks/rotate-intel-refresh-token.md` | SSH + the GitHub secret |

The rotation runbook **opens with a preflight, not a rotation**: `server.py` reads `os.getenv("INTEL_REFRESH_TOKEN","").strip() or SIGNAL_ALERT_CRON_TOKEN`, and `.env.example` ships the key **commented out**. Production may be authenticating on the fallback — making this an *add*, not a rotate — and rotating the fallback instead has a far wider blast radius. **Which is true cannot be determined from the repo.**

The cert script checks the thing the timer's existence proves nothing about: certs use the **webroot** plugin, so certbot never touches nginx, and without `renew_hook` a renewed certificate lands on disk unused while the site serves the expired one — a renewal that "succeeds" and still takes the site down.

---

## Verification

- Repro + fix measured locally against a booted stack (tables above).
- `waivers-smoke` + `journey-trade`: **15 passed**; `public-league` sub-nav walk passes on **both** projects.
- `tests/public_league`: **345 passed**; frontend vitest **1930 passed**.
- `ruff check .` / `ruff format --check .` clean; harness guards 13 passed.
- Both new shell scripts pass the `bash -n` gate `pr-validation` runs.

## Not in scope

Perfect Draft's panel failure — another session is on it (#735, triage on #732). Commenting there with what my investigation adds rather than duplicating it.


jasonleetucker-code:
CI status, and the blocker stated rather than left implicit.

**`Validate PR`: ✅ success.**
**`Boot stack + run journeys`: ❌ — 2 failed / 52 skipped / 140 passed.**

The two failures are `journey-perfect-draft.spec.js:75` and `:109`, which this PR deliberately does not touch (#735's session owns them, triage on #732). Measured against the nightly on `main` (`a6b426b76`):

| | main nightly | this PR |
|---|---|---|
| `journey-trade.spec.js:145` `/arbitrage` | failed | ✅ |
| `waivers-smoke.spec.js:36` desktop | failed | ✅ |
| `waivers-smoke.spec.js:36` mobile | flaky | ✅ |
| `public-league.spec.js:383` ×2 | failed | ✅ |
| `public-league.spec.js:93` sub-nav | flaky | ✅ |
| `journey-perfect-draft.spec.js` ×2 | failed | failed |
| **totals** | **6 failed, 2 flaky, 134 passed** | **2 failed, 0 flaky, 140 passed** |

Zero flaky is worth noting on its own: the `failOnFlakyTests` banner from #733 did not fire, so `waitForStreamSettled` removed those two as well rather than converting them into hard failures.

## Blocker

Not fixing the remaining two — they are pre-existing on `main` and belong to an active session. I did dig far enough to hand over something useful, and to rule things out:

**Ruled out: the backend, under CI's own data.** #735's triage checked `/api/draft/roster-context` against a live backend. I checked it against the **committed E2E snapshot**, which is the input that differs:

```
GET /api/draft/roster-context?leagueKey=dynasty_main&teamName=Blaine
  waiverLadder 90 · cutLadder 4 · openRosterSpots 5 · contextVersion 2026-08-04.v3
```

Complete. So the endpoint is not the cause under CI either, and the client-side `!result` branch remains the suspect — as their triage concluded.

**One observation I could not take further:** `/api/draft-capital?leagueKey=dynasty_main` reports `season: 2027`, with 72 picks all carrying `rookieBoardValue` and `rookieSource: ours_filtered`. If the page derives its rookie pool for the 2027 class while the board carries the 2026 one, `rookies` would be empty and `optimizeDraft` would return nothing — which is exactly the `!result` branch. **I have not verified that join**, and chasing it further would mean taking over their fix rather than helping it.

**Correction to something I said on #732:** I floated that `data/public_league/` being untracked might be feeding Perfect Draft different data in CI. Having now probed the actual endpoints, I have no evidence it touches that path — treat that as a discarded lead, not a hint.

---

jasonleetucker-code:
**Held, not abandoned** — deliberately not merging while E2E is red, even though the red is entirely the two Perfect Draft tests this PR scoped out.

Rationale: merging now would put `main` in a state where the nightly is red *for known reasons*, which is precisely the signal-quality problem #733 existed to fix. Waiting means the next nightly after both land is a genuine green rather than one that needs a footnote.

Accepted cost of waiting, stated plainly: the **28 dead production recap links** (§2) stay live until this merges. `prod-e2e-smoke` also stays red on them.

Unblocks as soon as `journey-perfect-draft.spec.js:75` / `:109` pass — nothing here depends on how they are fixed. I'll rebase onto whatever lands and merge.

---

jasonleetucker-code:
**Correcting my own hold reasoning.** I said above I'd wait for Perfect Draft to land first. That premise was wrong: **#743 cannot land first** — it is on hold waiting for the streaming fix, which is the half *this* PR carries. Measured:

| PR | Run | Fails on | Fixed by |
|---|---|---|---|
| **#741** (this) | `31014610123` | `journey-perfect-draft:75`, `:109` | #743 |
| **#743** | `31011698743` | `journey-trade:145`, `waivers-smoke:36` | #741 |
| **#747** | `31016166424` | necessarily perfect-draft | #743 |

A clean three-way deadlock. #747 is a third session's independent re-implementation of this PR's streaming fix — same file, same two specs — and it cannot be green either: its only non-test change is a comment block in `e2e.yml`, so it runs the same unfixed draft spec.

Waiting was therefore not a strategy, it was a stall. What the E2E signal is telling us is real and correct: **no single PR here can be green, because the two fixes have to be on one commit.**

Worth noting what this PR's own fixes did in that run, since it is the evidence that the streaming half works: `waivers-smoke:46` rookie toggle ✓ (mobile), `journey-trade:145` /arbitrage ✓, `public-league:406` recap route ✓, `public-league:93` sub-nav walk ✓ — every detector that was red is now green, and the only reds left are in code this PR does not touch.

Dispatched **[run 31019506229](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31019506229)** on a throwaway `claude/e2e-integration-741-743` — this branch merged with `claude/perfect-draft-team-latch`. Clean merge, zero file overlap. It answers the one question nobody has measured: do the two together go green? Result posted here and on #747 either way, red included.

Unchanged and still true: the **28 dead production recap links** (§2) stay live until this merges, and `prod-e2e-smoke` stays red on them.

---

jasonleetucker-code:
## The E2E red is not yours, and it is not Perfect Draft any more

Auditing the open backlog, so flagging rather than touching this branch — another session is clearly still on it (`c38a7d5d`), and `ASSISTANT_COORDINATION.md` says one assistant per branch.

**The three-way deadlock is over.** #743 merged as `d4ddc12f`, and in your latest run all three Perfect Draft specs pass:

```
✓ 30  journey-perfect-draft.spec.js:95   recommends a buyable combination of real rookies (5.9s)
✓ 31  journey-perfect-draft.spec.js:146  never plans to spend more than the budget (2.9s)
✓ 32  journey-perfect-draft.spec.js:167  says which league state it is in (8.0s)
```

**The new red is a different thing entirely**, and it is not a product failure at all. [Run 31025224262](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31025224262):

```
0 failed · 2 flaky · 140 passed (4.1m)

✘ journey-rankings.spec.js:109  search input filters the board by name fragment
✓ journey-rankings.spec.js:109  (retry #1)
✘ journey-rankings.spec.js:152  global search ('/' shortcut) …
✓ journey-rankings.spec.js:152  (retry #1)

Error: apiRequestContext.post: read ECONNRESET
```

Zero failures. The job is red purely because `failOnFlakyTests` promoted the retried green — which is the gate working as designed, on a flake that has nothing to do with this PR.

Both resets are the `authedPage` fixture's `POST /api/test/create-session`, which runs before every authed spec and had **no retry**. A keep-alive race between Playwright's request context and uvicorn recycling an idle connection therefore failed whichever authed spec happened to lose it. `journey-rankings.spec.js` is not a file this PR touches.

**#752 fixes it at the fixture**, scoped so it cannot launder anything: only a transport error (no HTTP answer at all) is retried, bounded at 3; any response including a 5xx is returned untouched; an exhausted retry rethrows so `stack-death-reporter.js` still matches its `CONNECTION_ERROR`. `failOnFlakyTests` is untouched and its guard still passes.

Once #752 lands, merging `main` here should give you a clean run.

## Separately — your §1 is now partly on `main`

#747 landed `awaitStreamSettled()` in `helpers/journey.js`, which you already accounted for in `565834f8`. Verified that what remains here is **not** duplicated on `main`:

| | on `main` | here |
|---|---|---|
| `matchup_recap._side_block` returns `None` on unresolved owner | **still live** — the 28 dead prod links | fixed |
| `public-league.spec.js` decides mobile/desktop via `isVisible().catch(() => false)` | **still live** | fixed |
| `stack-death-reporter.js` calls the transient "a PRODUCT defect" | **still live** | corrected |
| `AppShell.jsx` "its own defect" note | **still live** | corrected |
| the four operator-item files | absent | added |

So the §2 recap fix and the narrative corrections are unique to this branch and worth merging on their own.

---

jasonleetucker-code:
## Review of the three operator items — one finding on the ruleset

Reviewed the two new shell scripts and the payload as part of the backlog audit. Both scripts are solid: `set -Eeuo pipefail`, explicit `require_command`, a real `--dry-run` that validates the payload without sending, idempotent create-vs-update, no secrets in any committed file, and `verify-cert-renewal.sh` is read-only apart from `certbot renew --dry-run`, which is non-destructive by definition. The OA-02 workflow-name-vs-job-name correction is right and worth having caught — `Validate PR`, `Boot stack + run journeys`, `Validate Build Inputs`, `Deploy To Production` all check out against the workflow files.

### The ruleset does not prevent the scenario its own comment names

`main-protection-ruleset.json` states the risk as:

> `deploy.yml` ships every push to main straight to production. So an **accidental direct push** is an unreviewed path to the live site.

But the payload also carries:

```json
"bypass_actors": [
  { "actor_id": 5, "actor_type": "RepositoryRole", "bypass_mode": "always" }
]
```

`actor_id: 5` is the repository **admin** role, and `list_repository_collaborators` shows a single account — the owner, who is that admin. With `bypass_mode: always`, an accidental `git push origin main` from the owner still **succeeds**. GitHub does not prompt or warn; the push just goes through and deploys.

So as written the ruleset protects `main` from everyone except the only person who can push to it. The two rules that *do* still bite — `deletion` and `non_fast_forward` — are real value, and requiring `Validate PR` genuinely fixes the zero-checks-on-a-conflicted-PR hole. But the headline scenario is not covered.

This is a judgement call rather than a defect, so I am not proposing a change to the payload:

* **Keep the bypass** — accepts that the guard is about tooling hygiene and the conflicted-PR hole, not about stopping the owner. Worth saying so in the `_comment` so the next reader is not misled by the current framing.
* **Drop the bypass** — an accidental push is then actually rejected. A genuine hotfix means disabling the ruleset first, which is two deliberate clicks rather than an accident. Given `deploy.yml` auto-deploys, that asymmetry seems like the right trade.

I lean toward dropping it, precisely *because* the deploy is automatic — but it is your repo and your 3am hotfix, so it should be your call rather than mine.

### One nit

`apply-branch-protection.sh` writes the API response to a fixed `/tmp/ruleset-response.json`. `mktemp` would avoid both the predictable-path symlink surface and clobbering a concurrent run. Cosmetic on a single-user VPS.

---

jasonleetucker-code:
**Status: green except for a pre-existing flake that is not this PR's, now tracked as #753.**

`Validate PR` ✓. E2E has **0 failed / 140 passed** on `44145a3f2` — the red is `failOnFlakyTests` firing on `journey-rankings.spec.js`, a spec this PR does not touch. Full evidence and the ruled-out causes are in #753.

I tried to fix it rather than merge around it, and should be straight about the result: **one of my two changes was inert.**

- Raising the per-test timeout 90s → 150s does **not** apply to this failure. The assertion inside `gotoRankingsBoard` carries its own explicit `timeout: 60_000`, so the per-test budget never governs it. The underlying observation was still real — 60s + 30s in the helper exactly equalled the old 90s test timeout, leaving its twelve callers nothing for their own assertions — so the change stays, but it did not fix what I hoped.
- Hardening `auth-fixture.js` **did** address the second observed failure mode (the `create-session` POST throwing) and closes a genuine hole: it had no timeout, no retry, and turned every non-2xx into `test.skip()` while claiming "likely E2E_TEST_MODE not set". A 429 or transient 5xx silently skipped tests across the **nine** specs using it, and a skip reads as fine in every summary. `e2e.yml`'s own header already noted "create-session 429 makes signed-in specs skip" — known, written down as an aside, never fixed.

The real defect in #753 is that `/rankings` renders **zero** rows for a full 60s on a cold context — `element(s) not found`, not "not visible" — then works on retry. That points at the documented `buildRows` fail-fast path (empty rows + "no players" banner when a payload arrives without rank stamps), i.e. a failed or degraded `/api/data` fetch, not a rendering bug. Confirming it needs the `e2e-server-logs` artifact and the per-test `error-context.md`, which I can't download from this environment.

I did not reproduce it locally in any configuration: `journey-rankings` 36/36 with `--repeat-each=6 --retries=0`, and the full suite 192 passed / 0 failed / 0 flaky. This container is faster than the GitHub runner, so the load condition never arises here.

**Not re-running to chase a green.** Two attempts already went 3-flaky then 2-flaky; a third would be retry-farming, which is precisely what `failOnFlakyTests` exists to prevent — and laundering it here would be a poor way to treat a guard I added in #733.

This PR's own content is verified: `tests/public_league/` 345 passed, `tests/api/` 1747 passed / 14 skipped, harness guards 21 passed, `ruff check` clean. It still carries the **28 dead production recap links** fix, which stays live until someone merges.

Leaving the merge decision open rather than taking it unilaterally, since merging is a production action here — `deploy.yml` ships every push to `main` straight to prod.

---

PR #742, opened 2026-08-05
Close S-3 (C09), and fix the live BUY/SELL regression #740 left behind

jasonleetucker-code:
**This body replaces the original**, which described a rank-space implementation that has since been deleted. See the [rework comment](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/742#issuecomment-5193862907) for the history.

Batch **C4**. Closes Critical **C09 (S-3)**. C08/C10/C19 are closed on `main` and credited to **#740**, not to this pass.

> `src/api/data_contract.py` and `src/public_league/public_contract.py` are **byte-identical to `main`**. This PR computes no value and changes no board. Everything here is the consumer side.

---

## 1. A live regression on `main`: the trade page's BUY/SELL edge is dead for every player

#740 moved the market gap into value space and did the careful thing: it published the new number as `marketGapValueRatio` and stamped `marketGapMagnitude = None` on every row, so a consumer still reading the old field would fail *visibly*.

One failed silently anyway.

`trade-logic.js::getPlayerEdge` still read `marketGapMagnitude`. `Number(null)` is `0` — finite, and below the `MIN_EDGE_RANK_GAP = 3` floor — so the guard returned early for **every player**. `buildRows` never forwarded the replacement field either, so nothing downstream could have seen it.

Verified against current `main`:

| | |
|---|---|
| `data_contract.py:8292` | `row["marketGapMagnitude"] = None` |
| `dynasty-data.js:1130,1274` | forwards `marketGapMagnitude` only |
| `trade-logic.js:965` | `Number(row.marketGapMagnitude)` |

Four call sites were affected: `PlayerPopup`, two in `trade-sections.jsx`, and `league-analysis.js`.

Its tests stayed green because they feed rank-space fixtures (`2`, `5`, `6`) against a rank-space threshold (`3`). Nothing exercised it against a contract-shaped row.

**Measured after the fix: 287 rows signal, where it was 0 for every row.**

Also renamed the result's `rankGap` → `valueGapPct`. It is no longer a count of ranks, and `Math.round()` on a 0–1 ratio collapsed every gap to zero — `PlayerPopup` was rendering *"Consensus ranks 0 spots higher"*.

## 2. S-3 / C09 — the panels gated, sorted and labelled on the wrong quantity

Both `/edge` premium panels displayed the sign of `marketGapDirection` while gating **and** sorting on `sourceRankSpread` — how much the sources disagree with *each other*. Two different quantities. A clean, large gap on a player every source agreed about was **excluded**; a negligible gap on a player they were arguing over sorted **first**. The gate also barely gated.

Two further instances in `edge-helpers.js` *rendered* it, printing `"Sell +45 ranks"` — a disagreement width, described to the user as the gap.

**The display half was still live after the first pass, and is fixed here.** `DataTable` is not `presorted`, so it re-sorts by `defaultSort` — which was still `{key: "spread"}` — and the caller's ordering never reached the screen. The rendered column was `colSpread()` for the same reason. New `colMarketGap()` (the offense twin of the `colIdpGap()` the IDP panels have always had) now backs both the column and the sort.

`colSpread()` and `PREMIUM_SUMMARY_SPREAD` stay where spread genuinely *is* the subject: the disagreements panel and the disagreement stat.

## 3. Thresholds: four became two

`config/thresholds.json` is now the source, Python *loads* it so it cannot drift, and `test_threshold_parity.py` enforces the JS mirror — replacing a hand-written "mirror these Python constants" comment, which is a thing that cannot fail a build.

`PREMIUM_SUMMARY_VALUE_RATIO = 0.15` is p70 of the live distribution (0.1565 across 414 rows), giving 127 candidates against the 148 the old gate passed — re-aimed, not emptied. The trade page reuses `MARKET_GAP_MIN_VALUE_RATIO` rather than minting a fourth number: if the board will not label a gap, a trade surface must not signal on it.

The parity test's unit guard asserts the live unit **and** that no retired rank-space name has come back. That guard is what makes this regression class fail loudly next time.

---

## Testing

| Gate | Result |
|---|---|
| `npx vitest run` | **117 files, 1986 passed** |
| `ruff format --check .` / `ruff check .` | clean (990 files) |
| `audit_status.py` | no drift — 21 closed / 19 open / 1 deferred / 2 needs-review |
| `pytest -m "not livedata"` | see check run |

New guard: `frontend/__tests__/components/edge-page-market-gap.test.jsx` renders the page, so it catches a missing import by construction, and its fixture is built so ordering by gap and ordering by spread are **exact opposites**. **Mutation-tested both ways** — dropping the import fails 6/6 with a real `ReferenceError`; restoring `defaultSort: "spread"` + `colSpread()` fails 4/6.

That test exists because the repair pass found a `ReferenceError` shipping in this PR: `premiumBy` called `marketGapAtLeast`/`marketGapRatioOf` and `page.jsx` imported neither, so `/edge` threw during render. The repo has **no ESLint at all** and `build:nocheck` skips Next's lint, so nothing in CI could see it.

## Deliberately left alone, and said so

`idpMarketEdge` still computes client-side in ordinal space, and `MARKET_GAP_MIN_DIFF` stays in the JSON as an ordinal threshold for it. The backend emits no IDP gap, so deleting the path would remove a working surface to satisfy a rule. Tracked as the **IDP-anchor follow-up**: give the backend an IDP retail anchor, stamp the gap, then delete the client path and its threshold together.

`/edge`'s IDP panels also sort by scraping a number out of a human-readable title string (`idpGapOf`). Same reason; it should read a stamped magnitude once one exists.


jasonleetucker-code:
## Do not merge as-is — #740 fixed the same defect, better

**#740 landed a competing fix for S-1/S-2/S-4 while this PR was open, and its approach is superior to mine.** Merging this would either revert that work or ship two competing market-gap implementations. Flagging rather than resolving, because which one survives is an architectural call.

### Why #740's approach wins

I fixed the gap by moving the comparison into **rank space** (rank ÷ observed board depth) and then subtracting a measured **per-position basis** — TE +121, PICK −109 per-mille — because the tight-end reading is arithmetically correct and only *looks* like a signal.

#740 uses `sourceRankMeta[key].valueContribution` instead: post-ladder, common-scaled 0–9999, and **already past ADR-015's `convert_te_value`**. That removes the *cause*. Their measured medians by position — QB +0.008, TE +0.084, WR +0.110, RB +0.112 — show the TE anomaly simply gone, where mine measures it and subtracts it.

Removing the cause beats correcting for it. `src/api/rank_space.py` and the de-meaning in `_compute_market_gap` should be dropped.

### What this PR still has that main does not

Verified against `origin/main` at `cef17703b`:

| | on main | in this PR |
|---|---|---|
| **S-3** — `/edge` panels gate *and sort* on `sourceRankSpread` while displaying the gap's sign | **still open** (`page.jsx:143,158`) | fixed at 4 sites |
| `config/thresholds.json` + Python loader + parity test | absent | present |
| `marketGapUnknown` abstention reasons | absent (0 occurrences) | present — 386 IDP rows currently collapse to a bare `"none"` |
| `market_gap` harness surface | absent | 14 rows |

S-3 is the one with direct user impact still live on main: a clean, large gap on a player every source agrees about is **excluded** from the buy/sell panels, while a negligible gap on a player they're arguing over sorts **first**. The gate also barely gates — 376 of 414 rows passed it. And `edge-helpers.js` *renders* the disagreement width as `"Sell +45 ranks"`, describing it to the user as the gap.

### Suggested rework

1. Drop `src/api/rank_space.py` and the rank-space/de-meaning rewrite of `_compute_market_gap`; keep main's value-space version.
2. Keep the S-3 fix, re-pointed at main's magnitude (now a **relative value ratio**, not rank-space per-mille).
3. Keep `config/thresholds.json` + the parity test, re-deriving the gap thresholds against main's `MARKET_GAP_MIN_VALUE_RATIO` unit.
4. Keep `marketGapUnknown` and the harness surface.
5. Re-measure end to end — the numbers in the PR body above are all rank-space and no longer describe what would ship.

Local state at `d57ff2f92` is green (6653 Python, 1943 frontend, ruff clean) and CI passed at 14:10Z, but that was measured against my implementation, not the merged one.

---

jasonleetucker-code:
Heads-up before this merges: **the market-gap portion of this PR is already on `main`.**

Not closing it and not touching the branch — per the title this is Critical batch C4, and the rest of it may well be work nothing else has done. Only the "stop selling every tight end" part is duplicated.

## The overlap

#740 merged as `cef17703` and fixed the same defect the same way — comparing in value space instead of raw ordinals. `git merge-base --is-ancestor cef17703 <this branch>` is **false**, so this branch was cut before that landed and does not contain it. Shared files:

- `frontend/lib/display-helpers.js`
- `frontend/lib/thresholds.js`
- `src/api/data_contract.py`

What's live on `main` now: `marketEdge` / `marketGapLabel` / `idpMarketEdge` diff `sourceRankMeta[*].valueContribution`; `_compute_market_gap` does the same server-side so `/rankings` and `/edge` can't disagree; gate is `MARKET_GAP_MIN_VALUE_RATIO = 0.05`, calibrated to 68.8% non-HOLD against the old rank gate's 69.2%; `marketGapMagnitude` retired to `None` with `marketGapValueRatio` added rather than silently changing a published field's units.

Measured effect there: top-250 SELL went from 35 labels of which 32 were TE — every top-250 SELL on the board was a tight end — to 130 spread WR 47 / RB 34 / TE 32 / QB 13.

## What I am *not* claiming

**Not that #740's implementation is better than this one.** Nobody has compared them. #740 landed because it was green against current `main` at the time, not because anyone judged it superior. If this branch's version is better — a different threshold derivation, better handling of the IDP side, something #740 missed — that is worth arguing for, and swapping it in is a legitimate outcome.

What matters is that it's an argument someone makes deliberately, rather than a second implementation landing on top of a first by accident.

## Practically

Rebasing onto current `main` will surface this as a conflict in all three files rather than a silent double-apply, which is the good case. The rest of C4 should rebase cleanly.

## Context

This is the **eighth** duplicated repair across five sessions today, and the third independent solution of *this particular defect* (#740, #722, and this one). It's also the first one that happened *after* the earlier fix had already merged.

`scripts/check_work_claims.py` landed for exactly this (#746). Run against `frontend/lib/display-helpers.js` it currently names both this branch and `claude/fantasy-football-master-audit-umvex5`. Running it before the work would have caught this.

---

jasonleetucker-code:
## Resolved — #740's engine taken, S-3 closed, and a live regression fixed

Reworked per the assessment above. **The PR body is now stale** — everything in it describes the rank-space implementation that has been deleted. This comment is the accurate description.

### #740's engine taken wholesale

`src/api/data_contract.py` and `src/public_league/public_contract.py` are now **byte-identical to main (0 diff lines)**, so this PR's board *is* main's board — a tautology by construction rather than a claim that needs a diff to support it. `src/api/rank_space.py`, its tests, and the de-meaning pre-pass are deleted.

### A live regression on main, fixed

#740 handled the unit change carefully: it published the new number as `marketGapValueRatio` and stamped `marketGapMagnitude = None` on every row so a stale consumer would fail *visibly*. One failed silently anyway.

`trade-logic.js::getPlayerEdge` still read `marketGapMagnitude`. `Number(null)` is `0` — finite, and below the `MIN_EDGE_RANK_GAP = 3` floor — so it returned early for **every player** and the trade page's BUY/SELL edge was dead across the board. `buildRows` never forwarded the new field either, so nothing downstream could see it.

Its tests stayed green because they feed rank-space fixtures (`2`, `5`, `6`) against a rank-space threshold. Nothing exercised it against a contract-shaped row.

**Measured after the fix: 287 rows signal, where it was 0 for every row.**

Also renamed the result's `rankGap` → `valueGapPct`: it is no longer a count of ranks, and `Math.round()` on a 0–1 ratio collapsed every gap to zero. `PlayerPopup` was rendering that as *"Consensus ranks 0 spots higher"*; it now says *"values this player N% above"*.

### S-3 / C09 closed — the finding #740 left open

Both `/edge` premium panels filtered **and sorted** on `sourceRankSpread` while displaying the gap's sign, so a clean large gap on a player every source agreed about was excluded and a negligible gap on a contested player sorted first. The gate passed 148 of 414 two-sided rows, on the wrong axis. Two further instances in `edge-helpers` also *rendered* it — `"Sell +45 ranks"`, a disagreement width described to the user as the gap. All four now gate, sort and label on `marketGapValueRatio`.

### Thresholds — the count went *down*

Four thresholds gated this one concept; there are now **two**. `PREMIUM_SUMMARY_VALUE_RATIO = 0.15` is p70 of the live distribution (0.1565 across 414 rows), giving 127 candidates against the 148 the old gate passed — re-aimed, not emptied. The trade page reuses `MARKET_GAP_MIN_VALUE_RATIO` rather than minting a fourth number. The parity test's unit guard now asserts the live unit *and* that no retired rank-space name has come back — that guard is what makes this batch's regression class fail loudly next time.

### Registry: 21 closed, 19 open, 1 deferred, 2 needs-review

- **C08 (S-2), C10 (S-4), C19 (S-1)** → closed, credited to **#740**, not this pass, with the rank-space-vs-value-space reasoning recorded so nobody redoes the inferior version.
- **C12 (V-4)** → closed on main by the TEP auto-sentinel work; batch C6 step 6.4 no longer needs it. Found by diffing statuses against main *before* regenerating the generated file — a blind regenerate would have reverted it.
- **C09 (S-3)** → closed here.

### Gates

Frontend 1954 passed (115 files) · ruff clean (911 files) · coercion gate clean (baseline 694 → 692, the two S-3 sites burned down) · drift probe clean · board `--expect-no-value-change` OK · surfaces reproduce. Python suite running; will report.

The `market_gap` surface was rewritten for value space and includes a tight end with a huge ordinal gap (ranks 40 vs 180/200) whose value ratio is **0.025** — the S-2 artifact, demonstrated absent rather than argued about.

---

PR #743, opened 2026-08-05
Make Perfect Draft reachable: five defects behind one blank panel

jasonleetucker-code:
Fixes the `journey-perfect-draft` failures tracked in #732. `main`'s E2E has been red on its own nightly since this spec landed.

## Why this was invisible until now

`journey-perfect-draft.spec.js` reached `main` via #723 at 00:04Z; the last E2E run on `main` was 19:54Z the evening before. Across all 30 recorded runs of `e2e.yml` there is **no run on any `claude/perfect-draft-*` branch** — so the spec had never executed in CI, anywhere. #726's path-filtered PR trigger is what finally ran it, on its first outing, and it found real bugs.

Reproduced locally (2 failed / 1 passed, matching CI exactly) and fixed in order. Each defect was only visible once the one before it was out of the way, which is why this is five and not one.

## The app — a cold-load race, and it affects production

**1. The panel latched the team.** `useState(myTeamName)` captured the *first* name it ever saw. `teamName || myTeamName` looks like a fallback but only rescues the empty-string case, so a wrong non-empty name stuck forever. On a cold load it is reliably wrong:

`useLeague()` reports no league until `/api/leagues` answers, so `/draft` hydrates from the unsuffixed legacy storage key, misses, and falls back to `createDefaultWorkspace()` — whose first team is `"Russini Panini"`, not a team in the league. The lazy panel resolves inside that window, asks the backend for that team, gets **400 `unknown_team`**, classifies it as a failure and returns `null`. Permanently — the latched name never changed, so `useRosterContext` never refetched when the real workspace arrived. It takes its own team `<Select>` down with it, so a user cannot correct it without a reload.

Now derived from the workspace, with an explicit pick as an override, cleared on `league:changed` so a selection can't follow you into a league that has no such team.

**2. The rookie-pool auto-sync ran once, ever**, gated on `hydrated` — which flips true after that same pre-league hydration. So the 72-rookie pool was written onto the throwaway placeholder workspace (and persisted to the legacy key); the real league-scoped workspace then hydrated with no pool, and the sync refused to run again. Every rookie ended up without a `boardValue`, so the solve priced nothing and the panel reported that **no rookie was worth buying — on a board showing 72 rookies**. Being a race, it presented as an intermittent empty plan rather than a reliable one.

Now keyed on `draftStorageKey`, which also covers league switching. Re-running is cheap: the merge already no-ops when the incoming pool matches.

## The Next bridge — dev + E2E only

In production nginx routes `/api/*` straight to the backend, so none of these are reachable there. They bite exactly where Next serves `/api/*` itself.

**3. `/api/leagues` had no route at all.** Next answered its HTML 404 page, the JSON parse failed, `leagues` stayed empty, and `selectedLeagueKey` — which only returns a candidate present in that list — was permanently `""`. That is what un-scoped the draft workspace in (1) and (2).

**4. `roster-context` forwarded no cookie**, so the authenticated endpoint answered 401 and the panel vanished on it.

**5. `draft-capital` forwarded no cookie.** It answers 200 to anyone but **redacts** the rookie fields for public callers: `rookieName` and `rookieBoardValue` came back null on all 72 picks, which surfaces as no failure anywhere.

`proxyGet` grows an opt-in `cookie` param rather than always sending credentials. **21 other bridges still call it bare**, several against authenticated endpoints (`/api/bdvm/*`, `/api/sharp/*`) — deliberately out of scope here, but worth a follow-up.

## The spec — it could not have passed as written

- **`seedWorkspace` seeded budgets the app immediately zeroed.** The capital auto-sync keys teams by **Sleeper team name** while the fixture seeds the **manager names** `roster-context` matches on; the two namespaces barely overlap, so every seeded row took the unmatched branch and was zeroed. The panel then *correctly* reported that nothing is worth buying with $0. Seeding `feedBudget` distinct from `initialBudget` marks the row user-edited, which the sync preserves by design — the same state as a manager typing their carry-over balance.
- **`expectNoBadValueTokens(panel)`** passed a Locator to a `(page, selector)` helper, so `querySelectorAll` received an `HTMLElement` and threw. It had never run, because the assertions above it had never passed.
- **Added**: assert the Budget tile is non-zero *before* asserting rows, so a zeroed budget fails by name instead of as a 60s row timeout that says nothing about why.

**No locator was loosened.**

## Testing

- `journey-perfect-draft` **3/3, five consecutive runs** — it was a race, so one green run proves nothing.
- **Full E2E suite: 193 passed, 0 failed** (including `/arbitrage`, the #716 streaming race, on this run).
- **vitest: 1932 passed** across 111 files.
- The two new panel tests were **mutation-tested**: re-latching `teamName` turns the first one red. The first version of that test was thrown away for passing against *both* versions — `encodeURIComponent("Russini Panini")` does not match the `+` that `URLSearchParams` actually emits, so the mock had been answering 200 to everything.

## Not fixed here

The two team-name namespaces (capital feed vs. contract rosters) are a real seam — a workspace loaded from the capital feed carries names `roster-context` cannot resolve for 11 of 12 teams. The de-latch makes that recoverable rather than fatal, but threading `ownerId` through the workspace is the actual fix and is a bigger change. Noted on #732.

---

### DIFF
--- a/frontend/__tests__/components/PerfectDraftPanel.test.jsx
+++ b/frontend/__tests__/components/PerfectDraftPanel.test.jsx
@@ -644,3 +644,73 @@ describe("live-bid target is reconciled with the board", () => {
     expect(await screen.findByText(/Sold — pick the next player/i)).toBeInTheDocument();
   });
 });
+
+describe("the team it asks about", () => {
+  it("follows the workspace when a placeholder team is replaced by the real one", async () => {
+    // The cold-boot sequence on /draft, which is what took this panel off the
+    // page entirely on `main`.
+    //
+    // `useLeague()` reports no league until /api/leagues answers, so the page
+    // hydrates from the unsuffixed legacy storage key, misses, and falls back
+    // to `createDefaultWorkspace()` — whose first team is a placeholder that
+    // is not in the league. The panel used to LATCH that name in
+    // `useState(myTeamName)`, ask the backend for a team that does not exist,
+    // take the 400 as a failure and vanish — permanently, because the latched
+    // name never changed, so the roster-context hook never refetched even
+    // after the real workspace arrived. It took its own team <Select> with
+    // it, so the user could not correct it without a reload.
+    const placeholder = {
+      settings: { myTeamIdx: 0 },
+      teams: [{ name: "Russini Panini" }, { name: "Beta" }],
+    };
+
+    // URLSearchParams encodes the space as "+", so match on the bare token
+    // rather than encodeURIComponent — getting this wrong makes the mock
+    // answer 200 to everything and the test proves nothing.
+    fetch.mockImplementation((url) =>
+      Promise.resolve(
+        String(url).includes("Russini")
+          ? jsonResponse(400, { error: "unknown_team", leagueKey: "pd_main" })
+          : jsonResponse(200, okPayload()),
+      ),
+    );
+
+    const { container, rerender } = render(
+      <PerfectDraftPanel stats={STATS} workspace={placeholder} />,
+    );
+    // Wait for the 400 to be CLASSIFIED, not merely requested: asserting on
+    // the initial loading render would pass no matter what the server said.
+    await waitFor(() =>
+      expect(fetch).toHaveBeenCalledWith(
+        expect.stringContaining("teamName=Russini"),
+        expect.anything(),
+      ),
+    );
+    await waitFor(() =>
+      expect(container.querySelector(".perfect-draft-panel")).toBeNull(),
+    );
+
+    // The league resolves and the real league-scoped workspace hydrates.
+    rerender(<PerfectDraftPanel stats={STATS} workspace={WORKSPACE} />);
+    expect(await screen.findByRole("table")).toBeInTheDocument();
+  });
+
+  it("keeps an explicit pick instead of snapping back to the workspace team", async () => {
+    // The flip side: following the workspace must not undo a deliberate
+    // choice, or the team selector would be unusable.
+    fetch.mockResolvedValue(jsonResponse(200, okPayload()));
+    render(<PerfectDraftPanel stats={STATS} workspace={WORKSPACE} />);
+    await screen.findByRole("table");
+
+    const teamSelect = screen.getByRole("combobox", { name: /team to optimize/i });
+    fireEvent.change(teamSelect, { target: { value: "Beta" } });
+
+    await waitFor(() =>
+      expect(fetch).toHaveBeenCalledWith(
+        expect.stringContaining("teamName=Beta"),
+        expect.anything(),
+      ),
+    );
+    expect(teamSelect).toHaveValue("Beta");
+  });
+});
--- a/frontend/app/api/draft-capital/route.js
+++ b/frontend/app/api/draft-capital/route.js
@@ -8,10 +8,18 @@ export async function GET(request) {
     // not the server-resolved default.  Omitted → backend resolves as
     // before (unchanged for the unscoped DraftCapitalSection).
     const leagueKey = request?.nextUrl?.searchParams?.get("leagueKey");
-    const { data, status } = await proxyGet(
-      "/api/draft-capital",
-      leagueKey ? { searchParams: { leagueKey } } : undefined,
-    );
+    // Forward the session cookie. This endpoint answers 200 to anyone, but
+    // REDACTS the rookie fields for public callers — `rookieName` and
+    // `rookieBoardValue` come back null on all 72 picks. That is not a
+    // visible failure anywhere: /draft's auto-sync writes the empty pool
+    // onto the workspace, every rookie ends up without a positive
+    // `boardValue`, and the Perfect Draft solve drops the whole pool and
+    // renders a panel with zero rows. Only bites where Next serves /api/*
+    // itself (dev, E2E); nginx routes past this file in production.
+    const { data, status } = await proxyGet("/api/draft-capital", {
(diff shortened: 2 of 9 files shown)


jasonleetucker-code:
**Both defects this PR was opened for now pass in CI.** `journey-perfect-draft` is green, and `signed-in-smoke` is green after the switcher-disambiguation commit. The one remaining red is `waivers-smoke.spec.js:36` — the pre-existing `#S:1` streaming race (#716).

I did **not** take that on faith. It hard-failed on two consecutive runs of this branch where it had previously only been flaky-then-green, and this PR does change shell-level rendering (the league switcher renders for the first time), so "it's the known race" was a hypothesis worth falsifying rather than an excuse. Measured it the way the workflow header asks — same stack, same seeded snapshot, `--retries=0`, 15 runs per arm:

| Build | Pass | Fail | Rate |
|---|---|---|---|
| This branch (`1231aa56`) | 13 | 2 | **13%** |
| `main` (`aecc9adf`), frontend swapped + rebuilt | 12 | 3 | **20%** |

The rate is **not higher on this branch** — if anything marginally lower, and the difference across 30 runs is well inside noise. The failure is `main`'s, and this PR does not aggravate it.

Worth recording separately: **the workflow header's documented rate is stale.** It says 1/45 (~2%) on `/waivers`; both arms here measure ~13–20%, roughly an order of magnitude worse. Whatever changed since that was measured, #716 is a much bigger source of red than its own documentation claims — which matters, because at that rate a single retry is not enough to hide it and the suite will go red on unrelated PRs. That belongs on #716, not here.

To be explicit about what I did *not* do: **the `Include rookies` locator is untouched.** It is one of the two detectors this repo has for the render-twice class that every performance metric missed, and loosening it to get this PR green would destroy the only thing that catches that bug.

Baseline for the arm-swap: `git checkout origin/main -- frontend/`, plus removing the new `frontend/app/api/leagues/route.js` (a pathspec checkout does not delete files absent from the source tree — verified `/api/leagues` returned 404 through Next before measuring, which is the signature of main's state). Working tree restored to `1231aa56` afterward, verified clean.

---

jasonleetucker-code:
Re-ran the E2E job. Red again, and **again only on #716 detectors** — this time `journey-trade.spec.js:145` `/arbitrage` (desktop) and `waivers-smoke.spec.js:36` on **mobile-chromium**. Different detectors from the previous run; `journey-perfect-draft` and `signed-in-smoke` green in both.

I am not going to keep re-running, because the arithmetic says it won't converge. Combining the measured per-test rate (13% on this branch, 20% on `main`) with the four independent detector points the suite carries — `waivers` × {desktop, mobile}, `/arbitrage` × {desktop, mobile} — an all-green run is roughly a coin flip at best, and the three runs so far are consistent with that. **While #716 stands at its real rate, a green E2E run on any shell-touching PR is a matter of luck, not correctness.**

That is the honest blocker on this PR, and it is not something this PR can fix:

- Every failure across three runs is the pre-existing `#S:1` streaming duplicate.
- Both defects #743 exists to fix pass in CI, repeatedly.
- The measurement above shows this branch does not aggravate the race.
- The only way to make this PR green from inside it would be to loosen the very locators that are the sole detector for the render-twice class — which the workflow header forbids, for good reason.

So the decision is a maintainer one: merge on the evidence that the red is entirely #716, or hold #743 until #716 is fixed. I'd merge — it fixes a feature that is currently unreachable in production on a cold load, and holding it does not make #716 any better.

Separately, #716 deserves to be re-scoped: its documented 1/45 is off by an order of magnitude, and at the real rate it is no longer a background annoyance but an effective block on the safety net doing its job. Happy to take that on as its own change.

---

jasonleetucker-code:
**On hold pending #716**, per maintainer decision. Not blocked on anything in this diff — both defects it fixes pass in CI repeatedly, and the red is entirely the streaming race.

Picking up #716 now on its own branch off `main`. Once it lands and `main`'s E2E is genuinely green, I'll rebase this onto that clean baseline and re-run, so the green here means something instead of being a coin flip.

---

jasonleetucker-code:
Posting the integration result I promised on #747, since it is evidence for this PR: **green.**

[Run 31019506229](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31019506229) on `claude/e2e-integration-741-743` — this branch at `1231aa56` merged with #741. Job `success`, Playwright suite `success`, "Alert on workflow failure" and "Upload server logs on failure" both `skipped`, report artifact **319 KB / 1 file** against the 103–104 MB of failure videos both red runs produced. #733's coverage floor also gates on ≥120 passed / ≤60 skipped, so a green exit means the suite genuinely ran rather than silently collapsing to nothing.

So the two PRs were never in conflict — each was only ever red on the other's missing half. Nothing in this diff needed to change for that; it was already correct.

You have since rebased onto `f80eb5487` and #747 has landed, so your own CI is now the authority and my run is superseded. I mention it only because it independently confirms the call you made in your third comment — the red here was entirely #716, and holding was the right read of the evidence you had.

I've rebased #741 onto the post-#747 `main` and **dropped my duplicate `waitForStreamSettled` in favour of your `awaitStreamSettled`**, as proposed. `tests/e2e/helpers/journey.js` now carries **zero diff** from `main` on my branch, and `journey-trade.spec.js` / `waivers-smoke.spec.js` / `e2e.yml` are yours verbatim — I took your side on every conflict. #741 keeps only what is uniquely its own: the 28 dead production recap links, the `public-league.spec.js` viewport fix that goes with them, and the operator items. No further overlap between us.

Thanks for the `e2e.yml` header rewrite — it is a better record of the diagnosis than mine was.

---

PR #751, opened 2026-08-05
Wait for the sub-nav before deciding which one to drive

jasonleetucker-code:
The last thing keeping `main`'s E2E red, and the last thing keeping #732 open.

## The red was a flake, not a failure

After #747 and #743 landed I dispatched `e2e.yml` on `main` so the workflow's own close step could retire #732 on its own evidence. It came back red:

```
0 failed
1 flaky
  [mobile-chromium] › public-league.spec.js:93 › renders league page,
                       switches tabs, and never touches private endpoints
141 passed
```

Zero failures, and **zero `waivers`/`arbitrage` duplicates — #747's streaming fix held on `main`.** The run exits 1 because `playwright.config.js` sets `failOnFlakyTests`: a green earned on a retry is deliberately rejected so the close step can't retire the tracker on a shaky run. That guard did exactly its job.

This flake was always there. It only surfaced once the streaming flake that used to dominate every run was gone.

## Root cause

The spec picked its branch by sampling visibility **exactly once**:

```js
const useMobile = await mobileSelect.isVisible().catch(() => false);
```

Both controls are always in the DOM — `LeagueClient.jsx:323` renders the `<select>`, `:336` renders `<SubNav>` inside `.desktop-only` — and CSS decides which applies. Sampled on 390×844 right after `domcontentloaded`:

```
t~0ms    selectExists: true,  selectVisible: FALSE
t~200ms  selectExists: true,  selectVisible: true
```

Land in that window on a phone and `useMobile` is `false`, so the walk takes the desktop branch — where `.desktop-only` is `display: none`, which drops those buttons out of the **accessibility tree**, and `getByRole("button", …)` reads exactly that tree. It finds **0** for all twelve tabs, every one lands in `missingTabs`, and the assertion fails with "league sub-nav is missing tabs".

**Measured:** sampling immediately after `domcontentloaded` picks the wrong branch **4/15** on a phone viewport; with the wait, **0/15**. The spec doesn't sample quite that early, which is why the real rate is lower and why it never reproduced locally (0/12) — CI load widens the window.

Readiness only. The tab-presence assertion is untouched and still fails on a genuinely dropped or renamed tab.

## Not `expect(a.or(b)).toBeVisible()`

That was my first attempt and it is wrong. On desktop **both** controls match their selectors, so `or()` resolves to 2 elements and strict mode throws before it can wait for anything. Measured **8/8 pass on mobile and 8/8 fail on desktop** — the exact regression my own verification plan said to watch for. Polling "either is visible" tolerates both existing, which is the real situation.

## The reporter's advice was stale, and it is the copy that prints

`stack-death-reporter.js` still told every reader the first thing to check was React leaving its `<div id="S:1">` staging copy behind, and called it a **product defect**. #747 established both halves are wrong and fixed `e2e.yml`'s header — but missed this copy, which is the one that actually prints on a red run, and the one that sent an investigation down a dead end.

It now names the two causes that want **opposite** responses:

1. **A readiness race** — streamed routes, or a control that exists before it measures visible. Not a defect. Fix the *wait* (`awaitStreamSettled()`, or a poll for "which control is live").
2. **A persistent duplicate** that survives `awaitStreamSettled()` — that *is* a product defect (#709 shipped one). Fix the app.

The "do not add `.first()`" warning stays; it is still right, and it is why the detectors work.

## Testing

- `public-league` tab walk: **8/8 on `mobile-chromium` and 8/8 on `desktop-1366`**, `--retries=0`.
- Full E2E suite green **twice** (194, 203) with **no flaky count** — `failOnFlakyTests` means "0 failed" is not the bar.
- `test_e2e_harness_guards.py` + `tests/deploy/`: 84 passed (the harness guards police the reporter I edited).
- `ruff format --check` / `ruff check` clean with the pinned 0.6.9.

## After merge

Dispatch `e2e.yml` on `main`. A green run lets the close step retire **#732** on its own evidence rather than anyone's assertion. If it goes green and #732 stays open, that's a real finding about the close-step guard restored in #726.

---

### DIFF
--- a/tests/e2e/specs/public-league.spec.js
+++ b/tests/e2e/specs/public-league.spec.js
@@ -106,6 +106,38 @@ test.describe("public /league page", () => {
     // via a <select> dropdown; on desktop, each tab is a <button>.  Detect
     // which control is currently visible and drive it accordingly.
     const mobileSelect = page.getByLabel("Select league section");
+    const desktopTab = page.getByRole("button", { name: TABS[0], exact: true }).first();
+    // Wait for whichever control this viewport actually uses BEFORE deciding,
+    // rather than sampling visibility once at an arbitrary instant (#732).
+    //
+    // Both controls are always in the DOM — LeagueClient.jsx:323 renders the
+    // <select>, :336 renders <SubNav> inside `.desktop-only` — and CSS decides
+    // which applies. Measured on 390x844 right after domcontentloaded, the
+    // select EXISTS but is not yet visible at t~0 and becomes visible by
+    // ~200ms. A one-shot `isVisible()` landing in that window returns false on
+    // a phone, so the walk took the desktop branch; there `.desktop-only` is
+    // `display: none`, which drops those buttons out of the ACCESSIBILITY TREE,
+    // and `getByRole("button", …)` reads exactly that tree — so it found 0 for
+    // all twelve tabs and every one landed in `missingTabs`.
+    //
+    // Load-dependent, so it never reproduced locally (0/12) and only surfaced
+    // on a busy runner. Readiness only: the tab-presence assertion below is
+    // untouched and still fails on a genuinely dropped or renamed tab.
+    // Polled rather than `expect(a.or(b)).toBeVisible()`: on desktop BOTH
+    // controls match their selectors, so `or()` resolves to 2 elements and
+    // strict mode throws before it can wait for anything. Measured — that
+    // version passed 8/8 on mobile and failed 8/8 on desktop.
+    await expect
+      .poll(
+        async () =>
+          (await mobileSelect.isVisible().catch(() => false)) ||
+          (await desktopTab.isVisible().catch(() => false)),
+        {
+          message: "neither sub-nav control (mobile <select> / desktop tabs) became visible",
+          timeout: 60_000,
+        },
+      )
+      .toBe(true);
     const useMobile = await mobileSelect.isVisible().catch(() => false);
 
     // ── What this test does and does NOT prove ────────────────────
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Closing — **superseded by #741**, which landed on `main` as `83900f39` while this was waiting on CI.

Both halves of this PR are already there, and one of them is better than what I wrote.

**The sub-nav decision.** I polled until either control became visible. #741 decides from `page.viewportSize()` against the 768px breakpoint instead. That is strictly better: it is a fact about the run rather than a probe that can race or throw, so there is nothing left to wait for. It also names a cause mine did not — the `.catch(() => false)` was swallowing a *strict-mode violation* from the streamed duplicate, not just an early sample — and settles the stream first.

**The reporter banner.** `main`'s copy already carries the corrected React 19.2 mechanism, the explicit "this is React's behaviour, not a defect in this app", the `awaitStreamSettled()` remedy, the persistent-vs-transient distinction, and the do-not-add-`.first()` warning. There is nothing left for this branch to fix — including the `locator.or()` advice I had put in mine, which I had already measured failing 8/8 on desktop (both controls match, so strict mode throws before it can wait).

This PR's own run [31026906945](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31026906945) did confirm the fix worked — `public-league.spec.js:93` passed on both projects, desktop 1.9s and mobile-chromium 760ms — but that evidence now just corroborates #741's version.

What that run turned up that #741 did **not** cover is a separate budget defect, carried forward in **#754**: the `gotoRankingsBoard` 60s wait exhausting on its own, which no per-test cap raise reaches.

---

PR #752, opened 2026-08-05
Stop a dropped connection in the session mint from failing a whole E2E job

jasonleetucker-code:
Found while auditing the open PR backlog: **this is what is currently blocking #741**, and it can block any PR.

## The failure

`mintSession` runs in the `authedPage` fixture before every authed spec, and had no retry. A single transport error there fails the spec outright.

PR #741, [run 31025224262](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31025224262):

```
0 failed · 2 flaky · 140 passed (4.1m)

✘ journey-rankings.spec.js:109  search input filters the board by name fragment
✓ journey-rankings.spec.js:109  (retry #1)
✘ journey-rankings.spec.js:152  global search ('/' shortcut) …
✓ journey-rankings.spec.js:152  (retry #1)

Error: apiRequestContext.post: read ECONNRESET
Error: apiRequestContext.post: read ECONNRESET
```

Zero real failures. `failOnFlakyTests` then correctly turned the retried green into `exit code 1`, and #741 — whose diff is `src/public_league/matchup_recap.py`, a recap page and some comments — was blocked by two rankings specs it never touched.

The reset is a keep-alive race between Playwright's request context and uvicorn recycling an idle connection. It is undocumented anywhere in the repo; the only existing acknowledgement of `ECONNRESET` is `stack-death-reporter.js`'s `CONNECTION_ERROR` regex.

## Why this is not a relaxation of the flaky gate

The retry is scoped so it cannot launder a real failure, and the boundary is asserted rather than asserted-in-a-comment:

| situation | behaviour |
|---|---|
| transport error — the request got **no answer at all** | retried, bounded at 3 |
| **any** HTTP response, 5xx included | returned untouched, never retried |
| a non-transport throw (bad URL, API misuse) | propagates on attempt 1 |
| every attempt reset | rethrows, so `stack-death-reporter.js` still matches and still prints its banner |

A backend that answers wrongly fails exactly as before. A backend that is genuinely dead still fails, with the right banner. What is removed is a single dropped connection **during setup** — a POST that is not a behaviour under test — which was previously indistinguishable from a real defect.

`failOnFlakyTests` in `playwright.config.js` is **untouched**, and `TestFlakyGateIsArmed`, the guard that asserts it is still set, still passes. This fixes the flake instead of tolerating it, which is the posture that gate exists to enforce.

## Shape

`mintSession` moves to `tests/e2e/helpers/mint-session.js` so it carries no `@playwright/test` import — the fixture itself cannot be `require`d without Playwright's runner, which would have made the guard below impossible to run cheaply.

`tests/e2e/test_auth_fixture_retry.js` pins all four rows of that table and runs in the **fast pytest gate** via `test_e2e_harness_guards.py`, alongside the existing harness guards, for the reason that file states: this class of defect makes authed specs fail *everywhere at once* and so reads as a broken product rather than a broken harness.

## Testing

- `pytest tests/e2e/test_e2e_harness_guards.py` — **14 passed** (13 existing + the new one)
- `node tests/e2e/test_auth_fixture_retry.js` — 5 assertions passed
- `ruff format --check` / `ruff check` — clean
- **Mutation-tested**: changing `mint-session.js` to retry *every* error rather than only transport errors makes the guard fail on the "a real error must not be retried" assertion.

## Limitation, stated

I have not reproduced the `ECONNRESET` locally — it is a timing race on a loaded CI runner, and the diagnosis rests on the run log above plus the absence of any retry in the fixture. If the reset turns out to have a server-side cause in `/api/test/create-session`, this change makes the symptom rare rather than fixing that cause, and the exhausted-retry path still surfaces it.

---
_Generated by [Claude Code](https://claude.ai/code/session_01BYcUBrcQvMWvK4MJ5pTstc)_

### DIFF
--- a/tests/e2e/helpers/auth-fixture.js
+++ b/tests/e2e/helpers/auth-fixture.js
@@ -21,6 +21,10 @@
 const base = require("@playwright/test");
 const { pageUrl } = require("./journey");
 
+// Session minting lives in its own module so it can be exercised without
+// Playwright's runner present.  See mint-session.js for why it retries.
+const { mintSession } = require("./mint-session");
+
 exports.test = base.test.extend({
   authedPage: async ({ page, baseURL }, use) => {
     const secret = process.env.E2E_TEST_SECRET;
@@ -28,9 +32,7 @@ exports.test = base.test.extend({
       base.test.skip(true, "E2E_TEST_SECRET not set — skipping signed-in tests");
       return;
     }
-    const resp = await page.request.post(`${baseURL}/api/test/create-session`, {
-      headers: { Authorization: `Bearer ${secret}` },
-    });
+    const resp = await mintSession(page, baseURL, secret);
     if (!resp.ok()) {
       base.test.skip(
         true,
--- a/tests/e2e/helpers/mint-session.js
+++ b/tests/e2e/helpers/mint-session.js
@@ -1,0 +1,67 @@
+/**
+ * Mint the E2E test session, retrying ONLY a dropped connection.
+ *
+ * Split out of `auth-fixture.js` so it carries no `@playwright/test` import and
+ * can be exercised by a plain `node` script — the fixture itself cannot be
+ * required without Playwright's runner present.
+ *
+ * WHY THIS RETRIES — and why that is not laundering a failure.
+ *
+ * Every authed spec runs this in its fixture, before a single assertion.  On
+ * 2026-08-05 two of them (`journey-rankings.spec.js:109` and `:152`) failed
+ * with `apiRequestContext.post: read ECONNRESET` and both passed on Playwright's
+ * retry, so the run reported **0 failed, 2 flaky, 140 passed** — and
+ * `failOnFlakyTests` correctly turned that into a red job, blocking a PR whose
+ * files were nowhere near either spec.
+ *
+ * The reset is a keep-alive race between Playwright's request context and
+ * uvicorn recycling an idle connection.  It carries no product signal: this
+ * POST is setup, not a behaviour under test, and a connection that dies before
+ * it is answered says nothing about the app.
+ *
+ * The distinction that keeps `failOnFlakyTests` honest:
+ *
+ *   - a transport error (no answer)      → retry, bounded
+ *   - ANY HTTP response, including 5xx   → returned as-is, never retried
+ *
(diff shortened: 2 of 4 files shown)


jasonleetucker-code:
Closing — **superseded by #741**, which landed on `main` as `83900f39` shortly after this was opened. I tried to merge this first (it is the correct fix for the `ECONNRESET` that failed #751's first attempt) and got a conflict in `tests/e2e/helpers/auth-fixture.js`, which is how I found out.

`main` now has a `mintSession` that is a **superset** of this one:

| | this PR | `main` today |
|---|---|---|
| transport error (`ECONNRESET` &c.) | retried, bounded at 3 | retried, bounded at 3, with a 15s per-attempt `timeout` and exponential backoff |
| 429 / 5xx | returned untouched | **retried** — the rate limiter and a busy server are transient too |
| 404 | returned untouched → `test.skip` | `test.skip`, and it is the *only* status that skips |
| 401/403 on a wrong secret | returned untouched → `test.skip` | **throws**, with a message saying the server answered so this is a real misconfiguration |

That last row is the one worth calling out, because it is a hole this PR left open and #741 closed. The old fixture skipped on *any* non-ok status while its message claimed "likely E2E_TEST_MODE not set" — and a skip reads as fine in every summary this repo prints. Nine specs use the fixture, so one blip could quietly retire most of the signed-in suite.

The guard here (`test_auth_fixture_retry.js`) cannot be carried across as-is: it asserts "any HTTP response, 5xx included, is returned untouched", which is deliberately **not** `main`'s behaviour any more. Rewriting it against the new semantics would be worth doing on its own merits, but it is a different test from this one and does not belong on a branch that conflicts.

The diagnosis in this PR body was right and is worth keeping in the record: the reset is a keep-alive race between Playwright's request context and uvicorn recycling an idle connection, it carries no product signal, and it was failing whole jobs on specs the PR never touched. That is now fixed on `main`.

Related budget defect from the same runs, not covered by #741 and open in **#754**: `gotoRankingsBoard`'s inner 60s wait exhausting on its own — which is what this PR's own run [31027451127](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31027451127) went red on, rather than on a reset.

---

Issue #753, opened 2026-08-05
/rankings renders ZERO board rows for 60s on a cold context, then works on retry


jasonleetucker-code:
Recurring first-attempt failure in `journey-rankings.spec.js`. It always passes on retry, so before #733 added `failOnFlakyTests` it was invisible — Playwright's own status for a retried pass is `passed`, exit 0.

## The measurement

```
Error: rankings board should render rows
expect(locator).toBeVisible() failed
Locator: locator('.ds-table-wrap table tbody tr.rankings-row-clickable').first()
Expected: visible
Timeout: 60000ms
Error: element(s) not found
  at gotoRankingsBoard (tests/e2e/helpers/journey.js:277)
```

**`element(s) not found`, not "not visible"** — after a full 60 seconds the board has rendered **zero** rows. This is not a slow paint; nothing is there at all. A fresh browser context on retry renders it fine.

## Observed runs (all `desktop-1366`, always 0 failed / N flaky)

| Run | Flaky tests | First-attempt failure point |
|---|---|---|
| [31025224262](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31025224262) attempt 1 | `:81`, `:109`, `:134` | `gotoRankingsBoard` — 60s, no rows |
| [31025224262](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31025224262) attempt 2 | `:109`, `:152` | `auth-fixture.js:31` — the `create-session` POST threw |
| [31027429080](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31027429080) | `:81`, `:109` | `gotoRankingsBoard` — 60s, no rows |

The varying test set is a symptom, not the shape of the bug: twelve tests across three specs share `gotoRankingsBoard`, so whichever ones happen to hit a bad cold load are the ones reported.

## What is NOT the cause (ruled out, not assumed)

- **Not parallelism.** `playwright.config.js` sets `workers: 1` and `fullyParallel: false` in CI.
- **Not the React 19.2 streaming duplicate (#716).** That produces a strict-mode "resolved to 2 elements". This is *zero* elements.
- **Not a stale selector.** The same locator passes on retry in the same run, and passes 36/36 locally.
[...]

PR #754, opened 2026-08-05
Pin the wait-outlives-its-test invariant (and withdraw the timeout raise that did not fix it)

jasonleetucker-code:
**Net diff is one new test.** No timeout, assertion, locator or selector changes.

This branch opened with a timeout raise. The failing run's artifacts disproved it, so it is withdrawn in the second commit and only the guard remains. Both commits are kept so the reasoning is on the record rather than rewritten out.

## What I got wrong, and what disproved it

I argued `gotoRankingsBoard`'s 60s wait was exhausting because the board is slow under CI contention, and raised it to 90s. **#753 called this inert, and #753 is right.**

Its author asked the next person to start from the `e2e-server-logs` artifact and the per-test `error-context.md`. I pulled both from run [31027451127](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31027451127). The page snapshot at the moment of failure:

```yaml
- heading "No player data available" [level=3]
- paragraph: The backend may still be initializing.
```

`rankings/page.jsx:1647` renders that under `!loading && !error && rows.length === 0`. So at the moment of failure the fetch had **completed** (`loading` false), raised **nothing** (`error` null), and the payload materialized to **zero rows** — the documented `buildRows` fail-fast. The page had settled into a terminal empty state. It would have stayed there for any timeout; 90s buys 30 more seconds of the same empty state.

Full evidence, including the root-cause lead, is posted on #753 — that is where it belongs.

## What survives

`TestNoWaitCanOutliveItsTest`: no declared `timeout:` in `specs/` or `helpers/` may reach the per-test cap that contains it.

Such a wait can never reach its own deadline — the test is killed first, and the run reports "Test timeout of Nms exceeded" at whatever line was executing. The message written to explain what was being waited for never prints. That is how a budget bug reads as a product regression, and it happened twice on 2026-08-05 from opposite directions:

* `journey-trade.spec.js:205` polled `/arbitrage` with a 90s budget inside a 90s cap — run [31026906945](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31026906945) died as a **test** timeout.
* `gotoRankingsBoard` waited 60s inside that same cap after the page load had spent part of it — run 31027451127 died as the **inner** wait.

#741 fixed the first by raising the cap 90s → 150s. The guard makes that repair permanent instead of incidental; it does not change the cap.

The per-file ceiling accounts for `test.setTimeout()` overrides, which is why the perfect-draft and settings-overrides 90s waits are correctly **not** flagged.

## Testing

- `pytest tests/e2e/ tests/deploy/ -q` — **85 passed**.
- `ruff format --check .` / `ruff check .` with the pinned **0.6.9** — clean (989 files).
- **Mutation-tested both directions.** Restoring the 90s cap makes the guard name exactly `journey-tools-health.spec.js:74`, `journey-trade.spec.js:205` and the rankings wait — and correctly stays silent on the two files that raise their own ceiling. Re-indenting the config key makes it fail loudly ("this guard reads it by indentation, so reformatting that block silently disarms it") rather than passing vacuously.

## What this does not do

It does not make `main` green. #753 is the live defect keeping it red, and this PR deliberately does not paper over it.

https://claude.ai/code/session_01EeQ4SkH9khPRx2b3WunSTA

### DIFF
--- a/tests/e2e/helpers/journey.js
+++ b/tests/e2e/helpers/journey.js
@@ -274,8 +274,26 @@ async function gotoRankingsBoard(page, { minRows = 50 } = {}) {
   // which reads like a dead pipeline and was a dead cookie.
   await page.goto(pageUrl("/rankings"), { waitUntil: "domcontentloaded" });
   const rows = page.locator(SEL.boardRow);
+  // 90s, not 60s.  60s was measured EXHAUSTING on run 31027451127
+  // (journey-rankings.spec.js:109, desktop-1366), and the distinction
+  // matters: it failed with this helper's own message and "Timeout:
+  // 60000ms", i.e. the inner wait ran out on its own — the per-test cap
+  // was not what pre-empted it.  Raising `timeout` in playwright.config.js
+  // (150s since #741) therefore does not cover this case, and the two
+  // fixes are complementary rather than a re-litigation of that one.
+  //
+  // Same reading of the evidence as #741's: the board is SLOW under CI
+  // contention, not broken — the retry passed, and the same run measured
+  // POST /api/trade/finder at 66s against 7.7s on an unloaded runner, a
+  // ~9x stretch on a CPU-bound backend path.  /rankings pulls the ~4 MB
+  // contract through the same process.
+  //
+  // Budget arithmetic, kept in step with the config: this helper's worst
+  // case is 90s + the 30s row-count poll below = 120s, which leaves 60s
+  // of the 180s cap for the test body — the same headroom #741 sized the
+  // cap around.  Move one and move the other.
   await expect(rows.first(), "rankings board should render rows").toBeVisible({
-    timeout: 60_000,
+    timeout: 90_000,
   });
   await expect
     .poll(() => rows.count(), {
--- a/tests/e2e/helpers/journey.js
+++ b/tests/e2e/helpers/journey.js
@@ -274,26 +274,8 @@ async function gotoRankingsBoard(page, { minRows = 50 } = {}) {
   // which reads like a dead pipeline and was a dead cookie.
   await page.goto(pageUrl("/rankings"), { waitUntil: "domcontentloaded" });
   const rows = page.locator(SEL.boardRow);
-  // 90s, not 60s.  60s was measured EXHAUSTING on run 31027451127
-  // (journey-rankings.spec.js:109, desktop-1366), and the distinction
-  // matters: it failed with this helper's own message and "Timeout:
-  // 60000ms", i.e. the inner wait ran out on its own — the per-test cap
-  // was not what pre-empted it.  Raising `timeout` in playwright.config.js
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
## Independently verified — merging once `Validate PR` lands

Reviewed as part of the open-PR backlog audit. I re-ran your evidence rather than taking it, and it holds.

**Net diff is one new test class**, confirmed against `origin/main`: 92 insertions, `tests/e2e/test_e2e_harness_guards.py` only. No timeout, assertion, locator or selector changes — the withdrawal in `73d0e4b5` is complete.

**Guard passes and is not vacuous.** `pytest tests/e2e/test_e2e_harness_guards.py` → **14 passed**; ruff format + check clean.

**Mutation-tested both directions myself**, in a clean worktree:

| Mutation | Result |
|---|---|
| restore `timeout: 90_000` | fails, naming `specs/journey-tools-health.spec.js:74` and `specs/journey-trade.spec.js:205` — both real |
| re-indent the config key to 4 spaces | fails **loudly**: *"could not read exactly one per-test `timeout:` … reformatting that block silently disarms it"* |
| restored | passes |

The second one is the part worth calling out. A guard that reads config by indentation would normally be a liability — the usual failure is that it silently matches nothing and passes forever. Asserting `len(found) == 1` converts that into a loud failure, and the two anti-vacuity assertions (`len(files) > 10`, `scanned > 20`) close the same hole on the scan side. That is the right shape.

### One correction to the body

> Restoring the 90s cap makes the guard name exactly `journey-tools-health.spec.js:74`, `journey-trade.spec.js:205` **and the rankings wait**

It names the first two only. `gotoRankingsBoard`'s wait is `60_000`, which is *under* a 90s ceiling, so the guard correctly does not flag it — and that is consistent with your own conclusion that the rankings failure was never a budget problem. The guard behaves right; only that sentence overstates it by one.

### Its E2E red is #753, not this PR

Run [31029389899](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/31029389899): **0 failed · 1 flaky · 141 passed**, the flake being `journey-rankings.spec.js:28` → `gotoRankingsBoard`, 60s, `element(s) not found`. That is #753 exactly. A pytest-only diff cannot influence an E2E run, so this is inherited from `main` and the PR is right not to paper over it.

### Unrelated confirmation, since it came up in #753

The `auth-fixture.js` hardening that landed in `44145a3f2` is holding: **no `ECONNRESET` in either of the last two full E2E runs** (31027451127, 31029389899), where the run before it had two. Your version is also strictly better than the one I had proposed in #752 — mine retried transport errors only, yours also covers 429/5xx and narrows the `test.skip()` to 404, which closes a signal-integrity hole across all nine specs that mine left open. Closing #752 was the right call.

---

PR #758, opened 2026-08-05
Corroborate #747 at 2,400 loads, correct the loading.jsx claim, and name the duplication invariant

jasonleetucker-code:
**Read #747 first — it is the diagnosis, and it is right.** This branch was in flight when #747 landed and reached the same conclusion from the other end: #747 read `react-dom@19.2.8`'s `$RC`/`$RV` reveal machinery, this measured the running app. They agree. Rebased onto it, its `e2e.yml` rewrite taken wholesale, and the one claim of mine it disproves deleted.

What is left is worth having:

## 1. The same answer at 100× the sample

**2,400 loads**, six routes, two independent oracles: **zero persistent duplicates**. #747 measured 0/25 after `awaitStreamSettled()`; this is 0/2400. Corroboration by a different method, at a sample size that matters for a claim about a *rate*.

The second oracle counts the **specs' own locators** — `getByLabel(/Include rookies/i, {exact:false})`, lifted from `waivers-smoke.spec.js:38` — because the original detector was a strict-mode violation, and contradicting a measured number with a differently-shaped instrument is how `firstLoadChunks()` earned its retraction. Verified live rather than assumed: every needle baselines at exactly 1.

## 2. One correction that still stands

Both #747's summary and `e2e.yml` say routes with a `loading.jsx` "are wrapped by the App Router in a Suspense boundary and STREAMED". True — but the implication that those are *the* streaming routes is not:

| route | `loading.jsx` | stages mid-load |
|---|---|---|
| `/bdvm` | **no** | **20%** |
| `/waivers` | yes | 15.5% |
| `/arbitrage` | yes | 25% |
| `/trending`, `/finder` | no | 0% |

`/bdvm` has no loading file and streams *more often* than `/waivers`, which has one. Verified against the filesystem, not inferred from the harness.

So the population needing `awaitStreamSettled()` is **not** "the ~10 routes with a loading file", and a future spec on a route without one is not thereby safe. Any route whose tree suspends can stream; the loading file only supplies the fallback UI when it does. **This does not affect #747's fix**, which waits unconditionally and no-ops where nothing streamed.

## 3. A detector that names its subject

#747 keeps the strict locators, which is right — a *persistent* duplicate (the #709 class) is still a real defect and they are its only detector. But that detection is **incidental**: it lives in a spec about a rookie toggle and fires only because a locator happens to resolve twice, so it names the wrong subject when it goes off.

`tests/e2e/specs/ssr-duplication.spec.js` states the invariant directly, after `awaitStreamSettled()`. It asserts only route-independent properties — no surviving staging container, no duplicate element IDs, exactly one `<h1>` — and deliberately skips the `<main>` count, which is route-dependent (2 on `/waivers`, 1 on `/arbitrage`). A permanent duplicate is deterministic, so one navigation per route catches it; rates were the transient reveal's problem and that is no longer a defect.

`frontend/scripts/measure-duplication.mjs --loads 200` remains for when a rate *is* wanted.

## 4. Two defects fixed

**A flake that was turning this PR red.** `/tools/trade-coverage audits every Sleeper team` failed with `Received array: []`, then passed on retry — and the harness guard rejects a retried green. The spec polled `.trade-coverage-row` until the count matched, then read a *different* selector once. The obvious reading is a partial row, but that's impossible: the `<li>` and its name `<span>` are emitted together (`page.jsx:201-203`). The real mechanism is the list re-rendering between the two observations, since the page fans out one request per team. Fixed by polling the value being asserted — **stricter, not looser**.

**A column in my own instrument that silently went null.** `loading.jsx` was a hardcoded table that `--routes` never populated, so every run with an explicit route list reported `null` for the whole column — the same silent-nothing failure the needle detector exists to avoid. Now read from disk, resolved against the script rather than cwd.

## Testing

- 2,400 harness loads, both oracles, production build.
- 10/10 specs pass locally (`ssr-duplication`, `journey-tools-health`, `waivers-smoke`), **zero skips**.

Local trap, now documented in the ledger: `/api/test/create-session` is rate-limited at 60/min, so running the harness against the same backend as the E2E suite starves the auth fixture and specs **skip** — which reads as a non-failure. Use `RATE_LIMIT_BYPASS_IPS=127.0.0.1`. CI boots a fresh stack per job and is unaffected.

---


jasonleetucker-code:
## E2E is red on a suite-level flake, not on this PR's content

Three E2E runs on this branch, **three different tests** flaked once each — and no run had a hard failure:

| run | flaky test | failure mode |
|---|---|---|
| 1 | `journey-tools-health:100` (trade-coverage) | `Received array: []` |
| 2 | `journey-perfect-draft:146` | `.perfect-draft-panel` absent after 90s |
| 3 | `journey-settings-overrides:45` | `POST /api/rankings/overrides` → **503** |

Every run: **147 passed, 0 failed, 1 flaky.** The flaky test is never the same one twice.

**What the three have in common is the expensive backend path**, not the page: the trade-coverage audit fans out one request per team; the Perfect Draft panel needs `/api/draft/roster-context` (4 MB contract + lineup solver + scarcity); and the overrides POST rebuilds the whole ranking pipeline synchronously on the event loop. That reads as contention on a loaded runner, which is the shape #732 has been tracking for days — `waivers-smoke` retried green, `/arbitrage`, `public-league`, and the two Perfect Draft tests, all on `main`.

I fixed #1 in this PR (it was a genuine two-selector race — polled `.trade-coverage-row`, then read `.trade-coverage-team-name` once, across a list that re-renders as each team's request lands). #2 is the known race root-caused in #743, which **is** in this branch along with #750; #743's own write-up notes it needed five consecutive green runs to trust, so a residual tail is consistent with that. #3 I could not reproduce locally — that path is auth-gated and returns 401 without a session, so my probe couldn't exercise it.

**On whether this PR caused it.** `ssr-duplication.spec.js` adds 6 page loads (3 routes × 2 projects), and `/rankings` and `/waivers` do pull the private contract, so it is not zero added load — I'm flagging that rather than dismissing it. But it passed **6/6 in both runs that contained it**, the flaky test differs every run, and the same failure shapes predate this branch on `main`. I have not trimmed the spec, because doing so on a hunch would weaken a detector on the heaviest route to chase a cause I have not established — and `/rankings` is exactly where a render-twice regression would show first.

Locally on this branch: `journey-perfect-draft` **3/3**, and `ssr-duplication` + `journey-tools-health` + `waivers-smoke` **10/10 with zero skips**.

One thing worth knowing for anyone reproducing locally: `/api/test/create-session` is rate-limited at 60/min, so running the measurement harness against the same backend as the suite starves the auth fixture and specs **skip** — which reads as a pass. Use `RATE_LIMIT_BYPASS_IPS=127.0.0.1`.

Happy to dig into the overrides 503 as its own piece of work if you'd rather that block this PR; it looks like the more interesting of the three.

---

jasonleetucker-code:
Absorbed. The core SSR duplication invariant/tests from this pre-squash branch were ported to current main (`tests/e2e/specs/ssr-duplication.spec.js`). Closing the historical PR; the old measurement material remains in Git history.

PR #759, opened 2026-08-05
Stop non-data routes from fetching the player contract (and fix the dead search button)

jasonleetucker-code:
Ledger item "Planned #3". `/login` fetched the multi-MB `/api/data` contract and materialized ~1,100 rows to render a password box. Six routes did.

**The win is the network request, not just the render.** `prefetchBaseContract()` is called *inside* `useDynastyData` (`useDynastyData.js:111`), and only `PrivateAppShell` mounts that hook — so taking the other branch skips the GET itself, not merely the `buildRows` pass.

## Measured, in a signed-in browser

| route | contract requests | `<main>` rendered |
|---|---|---|
| `/login` | **0** | 121 chars |
| `/more` | **0** | 1,616 chars |
| `/design` | **0** | 3,854 chars |
| `/admin` | **0** | 2,148 chars |
| `/tools/source-health` | **0** | 141 chars |
| `/rankings` *(control)* | 1 | — |
| `/settings` *(control)* | 1 | — |
| `/tools/trade-coverage` *(control)* | 1 | — |

Watched as real network traffic rather than asserted from the code. The control group is the probe's own self-check: without it, five zeros could equally mean the probe was blind. Render is checked alongside, because a gated page that *crashed* would also fetch nothing.

## A second list, not more entries in `PUBLIC_ONLY_ROUTE_PREFIXES`

That constant is a **privacy boundary** — its own comment says the contract "leaks private rankings, edge signals, trade targets, and source-override state that public visitors must not see." Every route added here is private; only a signed-in user reaches `/admin`. Folding them together would leave the next person auditing "what is public" reading `/admin` off the public list and getting a wrong answer.

So: two independently-named predicates, composed. `PublicAppShell` becomes `NoPlayerDataAppShell`, since it now answers to both reasons.

**Two candidates were audited and rejected**, and are pinned as such so a future "obvious" addition fails a test that explains itself:

- **`/settings`** calls `useDynastyData()` *directly*, so gating the shell wouldn't even stop its fetch — it would only strip its search.
- **`/tools/trade-coverage`** reads `useApp().rawData` to build the audit.

**Gating on `authenticated` instead of a route list was also rejected.** `useAuth` is tri-state and `AppShellWrapper` passes `authenticated === true`, so it's `false` during the auth probe. That would serialize the contract fetch behind an auth round-trip on *every* page — trading a wasted fetch on `/login` for a slower first paint everywhere.

## Fixes a live defect on the way

`searchEnabled` is `privateDataEnabled && authenticated`, and `openSearch()` returns early without it — but `TopBar` and `MobileTopBar` rendered the search button on `authenticated` **alone**. So a signed-in user on `/league` has been seeing a search button that does nothing, and gating six more routes would have spread it.

`searchEnabled` now reaches both nav bars through the existing `AppShellSearchBridge`. Search reads the player contract, so hiding the affordance where there is no contract is the honest outcome, not a compromise. Defaults `true`, so callers predating the prop are unaffected.

## Testing

- **2010 tests pass** (118 files).
- New `app-shell-route-gates.test.js` pins both predicates, prefix-vs-separator matching (`/leagues` must not match `/league`), and the two deliberate exclusions.
- `TopBar.test.jsx` gains search-affordance gating for both nav bars, including the default-on case.
- `public-league.test.js`'s static guard is updated for the rename and gains an assertion that `/league` still routes to the dataless shell. Its `toBeTruthy()` on the regex match is load-bearing — a failed match would make the following `not.toMatch()` pass vacuously and the guard would silently stop guarding.
- Bundle budgets unchanged: `/login` 7.6/15 KB, `/more` 8.1/10 KB (the tightest in the repo).

Branched off `main` rather than stacked on #758, so this is independently mergeable.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01SaMgGfyHGEs38eGJ3kBf6H)_

### DIFF
--- a/frontend/__tests__/app-shell-route-gates.test.js
+++ b/frontend/__tests__/app-shell-route-gates.test.js
@@ -1,0 +1,113 @@
+/**
+ * AppShell route gates — which routes hydrate the player pipeline.
+ *
+ * Two INDEPENDENT predicates, deliberately not merged:
+ *
+ *   isPublicOnlyRoute    — a PRIVACY boundary. The /league subtree is
+ *                          reachable logged-out, and /api/data carries
+ *                          private rankings, edge signals and trade
+ *                          targets. Adding a route here is a security
+ *                          statement.
+ *   isNoPlayerDataRoute  — a COST decision. These are private routes
+ *                          that simply have no player data to render.
+ *
+ * Folding the second list into the first would make the next person
+ * auditing "what is public" read /admin off the public list and get a
+ * wrong answer. These tests exist mostly to keep them apart.
+ *
+ * Pure functions, so tested directly — a wrong entry breaks a page
+ * rather than slowing it, which is worth a cheap unit test.
+ */
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("next/navigation", () => ({ usePathname: () => "/" }));
+
+import { __routeGates } from "@/components/AppShell";
+
+const { isPublicOnlyRoute, isNoPlayerDataRoute, NO_PLAYER_DATA_ROUTE_PREFIXES } =
+  __routeGates;
+
+// Mirrors AppShell's own composition. If this drifts from the component,
+// the assertions below stop describing production.
+const privateDataEnabled = (pathname) =>
+  !isPublicOnlyRoute(pathname) && !isNoPlayerDataRoute(pathname);
+
+describe("isPublicOnlyRoute — the privacy boundary", () => {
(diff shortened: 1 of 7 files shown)


jasonleetucker-code:
Absorbed. Current main already contains the non-data/private-route player-contract gating (`NoPlayerDataAppShell`) and associated behavior. Closing this pre-squash PR rather than merging duplicate old history.

PR #760, opened 2026-08-05
Fix the board FPS harness — it was measuring itself, and windowing DOES fix the board

jasonleetucker-code:
**This PR's original conclusion was wrong and this body has been rewritten.** It previously said "row windowing does not fix it" and reverted a working implementation on that basis. The number behind that conclusion came from a harness that measured its own scroll loop. With the harness fixed, windowing takes the board from **22 FPS to 59.5 at 1×** and meets the ≥50 target.

## The defect in the instrument

The driver was:

```js
for (let i = 0; i < 30; i++) {
  window.scrollBy(0, 400);
  await new Promise((r) => setTimeout(r, 16));   // <-- this
}
```

That is 30 discrete jumps paced by a timer, not a scroll. The frame counter therefore measured how fast that loop could turn over — bounded **above** by the timer at ~60/s no matter how fast the page was. It is exactly why the 65-row board (32.1) and the 964-row board (28.6) came out indistinguishable, which was the entire basis for "row count is not the driver".

## The fix

CDP `Input.synthesizeScrollGesture` drives the scroll from outside the page; the page only records `requestAnimationFrame` timestamps. **No timer anywhere in the loop.** 9000 px at 2400 px/s. The report now leads with median FPS but also carries p95 frame time and the count of frames over 32 ms, because jank is long frames rather than a low mean.

A trivial-page control runs first at every throttle rate and is printed. That is what makes a board number readable, and it is decisive here: the control holds **60.0 FPS at 1×, 4× AND 6×**. The harness can produce 60 on this hardware, CPU throttling alone does not degrade it, so the board's collapse is the board's.

## Re-measured (3 runs, median)

| board | rows | DOM nodes | 1× | 4× | 6× |
|---|---|---|---|---|---|
| capped | 200 | 6,597 | 32.8 | 9.0 | — |
| "Show all" | 964 | 30,501 | 22–27 | 5.7–6.5 | 4.1 |
| **~60 rows laid out** (window proxy) | 964 | — | **59.5** | **23.3** | **15.0** |
| *control: trivial page* | — | ~1,800 | 60.0 | 60.0 | 59.7 |

**1×: 22 → 59.5, MEETS the ≥50 target.** 4×: 6.0 → 23.3 — still short of ≥30, but ~4× better. The old table's "windowing bought ~2 FPS" was the timer floor absorbing the entire effect.

The window proxy is `display: none` on `:nth-child(n+61)` applied to the live board — the same thing a window achieves by unmounting, measured without needing the implementation.

## It is LAYOUT, not paint, and no CSS reaches it

Every cheap intervention ablated with `!important` overrides on the live board, same driver, baseline re-measured first and last so drift (~10%) is visible. Every CSS arm sits inside that drift:

| suspect removed | 1× | 4× |
|---|---|---|
| baseline | 26.5 / 22.8 | 6.5 / 5.4 |
| `tr:hover .sticky-name` (cursor parked off-table) | 23.4 | 6.3 |
| `.sticky-name` `box-shadow` | 25.2 | 5.2 |
| `.sticky-name` `position: sticky` | 22.4 | 6.2 |
| `thead th` `position: sticky` | 21.4 | 4.4 |
| all sticky, everywhere | 20.6 | 4.7 |
| `table-layout: fixed` | 24.2 | 6.0 |
| `content-visibility: auto` on rows | 22.7 | 6.5 |
| `contain: layout paint style` on rows | 23.4 | 5.7 |
| **`visibility: hidden` rows 61+** — skips PAINT | **29.5** | **9.4** |
| **`display: none` rows 61+** — skips LAYOUT too | **59.5** | **23.3** |

The last two rows are the finding. Skipping *paint* for 900 rows buys ~3 FPS; skipping *layout* buys 33. **The cost is table layout and no containment property reaches it** — which is why `content-visibility` and `contain` did nothing. Neither is honoured on `<tr>` the way it is on a block element.

That closes "can CSS avoid virtualization here" with a measured no.

## And it is not JavaScript

The live board's rendered HTML was snapshotted with every stylesheet inlined and every `<script>` stripped, then reloaded as a static document and scrolled with the identical driver:

| | rows | nodes | 1× | 4× |
|---|---|---|---|---|
| live board (React, hydrated) | 964 | 30,501 | 27.9 | 6.2 |
| static snapshot, **no JS at all** | 964 | 30,475 | 27.5 | 6.9 |

Identical. Nothing runs during the scroll. So the old body's advice — "start by profiling what the page does per frame" — would have led straight to a dead end, and the harness that produced it would have been trusted the whole way.

## Two harness bugs fixed en route, both the same class as the one being retracted

- `MIN_FRAMES` was 30. At 4× the board delivered 29 frames across the whole gesture — under 8 FPS — and the guard **refused the run**. A guard firing on the finding. Now 8, with thin samples flagged in the output rather than suppressed.
- `thinSample` was computed inside the function serialized into the page, where a module-scope constant is a runtime `ReferenceError` rather than a build error. Stamped by the caller now, with a note.

## What this PR now contains

The corrected harness plus the ledger retraction. It does **not** re-land the windowing implementation — that is planned item #2, now live again, and it belongs in its own PR with the one genuine obstacle re-solved: `journey-rankings.spec.js:81` counts rows before/after filtering and asserts the count drops, which pins to the window size under windowing. That assertion should be re-pointed at `aria-rowcount` (the honest observable once a window exists), not deleted. The repo's standing rule against loosening board assertions still applies; what changed is that the trade is now worth making, because the change delivers its goal.

Also worth carrying forward: the 4× target is still missed (23.3 vs ≥30). Windowing is necessary and not sufficient — per-row node reduction (~32 nodes/row today) is the complement, and it now has a measured reason to exist.

New flags: `--capped` (measure without "Show all", which is how the row-count question gets a paired comparison) and `--no-hover` (park the cursor off the table).

### DIFF
--- a/docs/performance-optimization.md
+++ b/docs/performance-optimization.md
@@ -995,7 +995,122 @@ One known tail: `/draft` hits ~0.18 in roughly 1 run in 5, attributed
 to a `DIV.muted` block collapsing 361×166 → 0 at ~900ms.  Pre-existing,
 not round 6's doing, and not yet fixed.
 
-## 2026-08-05 — row windowing: BUILT, MEASURED, REVERTED
+## 2026-08-06 — RETRACTION: row windowing DOES fix it, and the harness lied
+
+**The section immediately below is wrong and is kept only as a record.**
+Its conclusion — "row count is not what makes the board scroll slowly" —
+was produced by a harness that measured its own scroll loop. Fixing the
+harness reverses the finding completely.
+
+### The defect in the instrument
+
+`measure-board-fps.mjs` drove the page with `window.scrollBy()` +
+`await new Promise(r => setTimeout(r, 16))`, 30 times. That is not a
+scroll; it is 30 jumps paced by a timer, and the frame counter therefore
+measured how fast that loop could turn over — bounded **above** by the
+timer at ~60/s regardless of how fast the page was. Every absolute number
+below is a property of the loop as much as of the board, which is exactly
+why the 65-row board (32.1) and the 964-row board (28.6) came out
+indistinguishable. They are not.
+
+Rewritten to drive the scroll with CDP `Input.synthesizeScrollGesture`
+and sample `requestAnimationFrame` timestamps with **no timer anywhere in
+the loop**, 9000 px at 2400 px/s, 3 runs each, median:
+
+| board | rows | DOM nodes | 1× | 4× | 6× |
+|---|---|---|---|---|---|
+| capped | 200 | 6,597 | 32.8 | 9.0 | — |
+| "Show all" | 964 | 30,501 | **22–27** | **5.7–6.5** | **4.1** |
+| *control: trivial 600-div page* | — | ~1,800 | **60.0** | **60.0** | **59.7** |
+
+The control is the part that makes the rest readable, and it is new: a
+trivial page holds **60 FPS at every throttle rate, including 6×**. So
+the harness can produce 60 here, CPU throttling alone does not degrade
+it, and the board's collapse is entirely the board's.
+
+### Windowing reaches the 1× target outright
+
+Simulated by removing all but ~60 rows from layout on the live board
+(`display:none` on `:nth-child(n+61)`) — the same thing a window achieves
+by unmounting, without needing the implementation:
+
+| arm | 1× | 4× | 6× |
+|---|---|---|---|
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Absorbed/superseded. Current main already has the board FPS measurement harness and row-windowing machinery; the historical performance claim is pre-squash and must not be re-merged. Closing as absorbed.

PR #761, opened 2026-08-05
Measure first-load JS with an instrument that can be checked

jasonleetucker-code:
> **Body revised 2026-08-06.** The original claimed the conditional `useDynastyData` import was worth the full 57.7 KB raw / 19.4 KB gz. That was an overstatement and is corrected in "What is actually cuttable" below. Nothing about the instrument or the 641.7 KB measurement changes — the correction is to an inference I drew from it, which was about to justify a much riskier change than the bytes warrant.

Planned item #1 ("the framework baseline every route pays") was blocked on not having an instrument anyone could trust. Two previous ones were wrong: `firstLoadChunks()` read a manifest Next 16 stopped emitting, and its replacement treated everything in `static/chunks/` as always-loaded — which is how it scored round 6's refactor as **+213 KB shared** while `/league`'s slice actually *fell* 163.4 → 37.8 KB. It reported an improvement as a regression, confidently.

## The measurement

35 routes, `--webpack` build, Next 16.2.12:

| | raw | gzipped |
|---|---|---|
| **ALWAYS LOADED** (intersection of every route) | **641.7 KB** | **202.0 KB** |
| — framework (Next runtime, react-dom, polyfills, webpack) | 528.6 KB | 162.7 KB |
| — app-owned | 113.2 KB | 39.3 KB |

Heaviest route slices on top: `/draft` 142.7 KB raw, `/rankings` 108.9, `/` 93.5, `/trade` 89.0. These **supersede the ledger's framework figures**, which were Next 15.5.22 numbers from the retracted instrument.

## Why this one can be trusted

It reads the `<script src>` tags out of each route's **prerendered HTML** (`.next/server/app/**.html`) — Next's own statement of what the browser fetches before the page is interactive — and intersects across routes. A chunk appears there only if that route actually preloads it, so a dynamic import cannot masquerade as shared.

`_global-error.html` is excluded: it bypasses the root layout, loads a strict subset, and would silently shrink the intersection to a set no real route has.

## The control, and why the first one was worthless

The instrument asserts that large chunks no route preloads are excluded. This build has 513 KB of them: `ad2866b8.*` (193.4 KB, html2canvas behind `await import()` in ScreenshotFab), `framework-*` (185.2 KB) and `main-*` (134.4 KB), the latter two being pages-router artifacts the app router never loads. The old metric counted all three against every route.

My first version of that control asserted no always-loaded chunk had `html2canvas` in its URL. Webpack content-hashes that chunk to `ad2866b8.<hash>.js`, so the string never appears in a filename and the check **could not fail**. It passed, and it meant nothing. A control that cannot fail reads as verification and is worse than none. The replacement is anchored on a property of the build rather than a filename, and the script exits non-zero when it cannot measure — same posture as `assertGateIsMeasuring()` in `check-bundle-sizes.mjs`.

## What is actually cuttable

**Framework is 528.6 KB of the 641.7 and is not this repo's code.** Next's client runtime (217.1 KB) + react-dom (195.2 KB) + polyfills (110.0 KB, `noModule` so modern browsers skip it — call the modern floor 531.7 KB raw).

**App-owned is 113.2 KB raw / 39.3 KB gz.** The biggest single piece is chunk `5827` at 36.0 KB raw / 12.2 KB gz — the shell chrome (TopBar, MobileChrome, NavMenu, both switchers, ScreenshotFab, StaleDataBanner). So "defer shell components on interaction" **caps out around 12 KB gz**, worth knowing before someone spends a week on it.

### A runtime gate is not a bundling gate — and the correction

`PUBLIC_ONLY_ROUTE_PREFIXES` stops `/league` from *calling* `useDynastyData()`. It does not stop it from *downloading* it: the import in `AppShell.jsx` is static, so the pipeline is in the graph unconditionally. Verified against this build — chunks `8805`, `6023`, `2910` and `2063` all appear in `/league`'s route HTML. **57.7 KB raw / 19.4 KB gz downloaded by a page forbidden to use it.**

That figure is correct as a statement of what `/league` wastes. **What was wrong is the implication that making AppShell's import conditional would recover it.** Traced statically from `app/layout.jsx`, only one of those chunks has AppShell as its reason:

| chunk | reached from |
|---|---|
| `2910` | `components/AppShell.jsx` — **the only edge** |
| `8805` | `useDynastyData.js` **and** `components/useLeague.js` |
| `6023` | `useDynastyData.js`, `useTeam.js`, `useTerminal.js` |
| `2063` | `useLeague.js` |

The surviving edges are the **chrome, not the data layer**: `TopBar.jsx:21-22` and `MobileChrome.jsx:21-22` statically import `LeagueSwitcher` → `useLeague` → `lib/dynasty-data`, and `TeamSwitcher` → `useTeam`. The nav bar renders on every route, so those three chunks stay put no matter what AppShell does. Removing them means deferring visible chrome and making the switchers pop in after paint — a different trade, and not the one this section originally described.

**So the ceiling on a conditional AppShell import is chunk 2910 alone: 12.0 KB raw / 4.5 KB gz.** #763 (stacked on this branch) then takes 6.6 KB raw / 2.3 KB gz of that out of the every-route set by moving `lib/waiver-logic` behind the already-lazy CommandPalette, leaving **5.3 KB raw / 2.2 KB gz** as what remains.

At that size it is worth naming the hazard rather than pursuing it: a lazily-loaded `PrivateAppShell` **component** would remount every page in the app. Both shell branches render at the same position with `{children}` inside (`AppShell.jsx:141-146`), so when the chunk lands the element type at that position changes — page state discarded, effects re-run, fetches fired twice, scroll reset. A null-rendering leaf bridge avoids it, at real complexity for ~2 KB gz. Both the corrected attribution and the hazard are recorded in `docs/performance-optimization.md` on #763.

### And do not reach for `dynamic()`

`AppShell.jsx:16-72` documents with measurements that `dynamic()` there puts `{children}` inside a Suspense boundary and permanently duplicates every page's DOM. The imperative `await import()` idiom in `useLazyComponent` is the one to use.

## Scoping #759 honestly

Gating six routes out of the data pipeline removes the multi-MB `/api/data` fetch and the ~1,100-row `buildRows` pass — the dominant cost, and real — but **zero bytes of JS**.

### DIFF
--- a/docs/performance-optimization.md
+++ b/docs/performance-optimization.md
@@ -995,6 +995,100 @@ One known tail: `/draft` hits ~0.18 in roughly 1 run in 5, attributed
 to a `DIV.muted` block collapsing 361×166 → 0 at ~900ms.  Pre-existing,
 not round 6's doing, and not yet fixed.
 
+## 2026-08-05 — first-load JS, measured truthfully at last
+
+Planned item #1 ("the framework baseline every route pays") was blocked on
+not having an instrument that could be trusted. It has one now:
+`frontend/scripts/measure-first-load.mjs`. The numbers below are its
+output, and anyone can re-run it.
+
+### The measurement
+
+35 routes, `--webpack` production build, Next 16.2.12:
+
+| | raw | gzipped |
+|---|---|---|
+| **ALWAYS LOADED** (intersection of every route) | **641.7 KB** | **202.0 KB** |
+| — framework (Next runtime, react-dom, polyfills, webpack) | 528.6 KB | 162.7 KB |
+| — app-owned | 113.2 KB | 39.3 KB |
+
+Heaviest route-specific slices on top of that: `/draft` 142.7 KB raw
+(41.8 gz), `/rankings` 108.9 (37.7), `/` 93.5 (29.1), `/trade` 89.0 (30.9),
+`/waivers` 74.1 (26.5).
+
+**These supersede the ledger's earlier framework figures**, which were
+measured on Next 15.5.22 by the retracted instrument. Cite these instead.
+
+### Why this instrument can be trusted where the last two could not
+
+It reads the `<script src>` tags out of each route's **prerendered HTML**
+(`.next/server/app/**.html`) — Next's own statement of what the browser
+fetches before the page is interactive — and intersects across routes. A
+chunk appears there only if that route actually preloads it.
+
+That is exactly the property the retracted metrics lacked.
+`firstLoadChunks()` read a manifest Next 16 stopped emitting; its
+replacement treated everything in `static/chunks/` as always-loaded, which
+is how it scored round 6's refactor as +213 KB shared while `/league`'s
+slice actually fell 163.4 → 37.8 KB.
+
+**The control makes that concrete, and it can fail.** This build has
+513 KB of chunks on disk that NO route preloads —
+`ad2866b8.*` (193.4 KB, html2canvas behind `await import()` in
+ScreenshotFab), `framework-*` (185.2 KB) and `main-*` (134.4 KB), the
+latter two being pages-router artifacts the app router never loads. The
+old metric counted all three against every route. The instrument asserts
+at least one such chunk exists and exits non-zero if not.
+
+*A note on how that control was arrived at, because the first version was
+worthless:* it originally asserted no always-loaded chunk had
+`html2canvas` in its URL. Webpack content-hashes that chunk to
+`ad2866b8.<hash>.js`, so the string never appears in a filename and the
+check **could not fail**. It passed, and it meant nothing. Grepping for
+`html2canvas` finds only the import specifier inside the shell chunk. A
+control that cannot fail reads as verification and is worse than none.
+
+### Where the 641.7 KB actually goes, and what is worth cutting
+
+* **Framework is 528.6 KB raw / 162.7 KB gz of it, and it is not
+  app-shaped.** Next's client runtime (217.1 KB) + react-dom (195.2 KB) +
+  polyfills (110.0 KB, `noModule` so modern browsers skip it — call the
+  modern floor 531.7 KB raw). Nothing in this repo's control.
+* **App-owned is 113.2 KB raw / 39.3 KB gz**, and the single biggest piece
+  is chunk `5827` at 36.0 KB raw / 12.2 KB gz — the shell chrome: TopBar,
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
## Correction to this PR's body: the 57.7 KB is not all AppShell's to remove

The body says the static `useDynastyData` import in `AppShell.jsx` ships **57.7 KB raw / 19.4 KB gz** to `/league`, and implies making it conditional would recover that. The 57.7 KB figure is correct as a statement of what `/league` downloads and cannot use. **The implication is wrong** — most of it does not hang off AppShell.

Traced statically from `app/layout.jsx`, only one of the four chunks has AppShell as its reason:

| chunk | reached from |
|---|---|
| `2910` | `components/AppShell.jsx` — **the only edge** |
| `8805` | `useDynastyData.js` **and** `components/useLeague.js` |
| `6023` | `useDynastyData.js`, `useTeam.js`, `useTerminal.js` |
| `2063` | `useLeague.js` |

The surviving edges are the **chrome, not the data layer**: `TopBar.jsx:21-22` and `MobileChrome.jsx:21-22` statically import `LeagueSwitcher` → `useLeague` → `lib/dynasty-data`, and `TeamSwitcher` → `useTeam`. The nav bar renders on every route, so those three chunks stay put no matter what AppShell does. Removing them means deferring visible chrome and having the switchers pop in after paint — a different trade, and not the one the body was describing.

So the ceiling on "make the player pipeline import conditional" is chunk 2910 alone: **12.0 KB raw / 4.5 KB gz**, not 57.7 / 19.4.

#763 (stacked on this branch) then takes 6.6 KB raw / 2.3 KB gz of that 2910 out of the every-route set by moving `lib/waiver-logic` behind the already-lazy CommandPalette, leaving **5.3 KB raw / 2.2 KB gz** as the remaining ceiling. That is now small enough to weigh against a real hazard: a lazily-loaded `PrivateAppShell` *component* would remount every page in the app, because both shell branches render at the same position with `{children}` inside (`AppShell.jsx:141-146`), so the element type at that position changes when the chunk lands.

Nothing about the instrument or the 641.7 KB measurement changes — the correction is to the inference drawn from it, which I wrote and which was about to justify a much riskier change than the bytes warrant. Both the corrected attribution and the hazard are recorded in `docs/performance-optimization.md` on #763.

---

jasonleetucker-code:
Pre-squash historical instrumentation. Its first-load measurements are no longer current evidence. Closing the old PR; if first-load instrumentation is needed again, re-derive and remeasure on current main.

PR #762, opened 2026-08-06
Root cause of the rotating E2E flake: the Next bridge serves an unstamped snapshot as the contract

jasonleetucker-code:
Closes the investigation on **#732**. The rotating E2E flake is fully explained, and **the backend was never "broken"** — which is why four investigations (including two of mine) landed on the wrong subsystem.

Two commits, two halves of one cause: **the stall**, and **what the stall was allowed to turn into**.

## The chain

1. **The boot warm holds a lock against live traffic.** `_warm_overlays_in_background` calls `fetch_sleeper_overlay(force_refresh=True)`; `force_refresh` **skips the cache read entirely** (`sleeper_overlay.py:1197`) and takes the per-league build lock for the whole build — ~47–70 Sleeper URLs at 8 workers, 8 s timeout each. An `/api/data` request in that window **blocks for the remainder of that build**.
2. **The Next bridge aborts at 4 s and silently falls back to disk.** The page fetches `/api/dynasty-data` on the *page* origin (`route.js:31`, `:249`). The committed seed carries **zero rank stamps**.
3. **`buildRows` fail-fasts on exactly that shape** (`lib/dynasty-data.js:1358`) — deliberately, so a drift-prone local blend never renders. Blank board, no error.
4. **It's sticky:** the module-scope contract cache then serves that payload to every mount, and AppShell mounts `useDynastyData` on *every* route. One stall degrades the whole app.

**Why "five runs, five different specs":** the readiness gate passes on `has_data`, which `_prime_latest_payload` sets **before** the warm it spawns finishes. "Backend ready" never meant "`/api/data` answers within 4 s", so whichever spec is running when a request lands in the lock window is the one that fails.

## The fixes

**1 — Never block the contract on the overlay.** `fetch_sleeper_overlay` takes `max_wait_sec`. Request-path callers get, in order: fresh payload → stale payload → `None`. Background callers (boot warm, ros scrape, signal-alerts) keep the unbounded wait, because a warm that skipped its build would leave the cache cold forever — pinned by a test. Budget is **1.5 s**, sized against the bridge's 4 s idle timeout rather than guessed. The overlay is additive: `get_data` already serves the board without it.

**2 — Never serve an unstamped snapshot as a contract.** Verified end-to-end against a dead backend:

| | before | after |
|---|---|---|
| bridge | **HTTP 200** + unusable payload | **HTTP 503** |
| body | (silently empty board) | `{"reason":"disk_snapshot_unstamped","players":1075}` |

## The diagnostic had the same two bugs — which is why nobody found this

1. **It probed the wrong process.** `page.request.get("/api/...")` resolves against `baseURL` (`:8000`); the page loads from `E2E_PAGE_ORIGIN` (`:3000`) and takes its contract from the bridge. **The tell was `legacyPlayers=1093`, matching neither on-disk snapshot** (1075 / 1074).
2. **It counted stamps only over `playersArray`**, which `view=app` pops by design (`server.py:2150`) — so it printed *"the scrape pipeline is not stamping"* on **every** failure regardless of cause.

Now fixed, and **it has already proven itself on this PR's own CI run**:

```
backend (baseURL)     -> HTTP 200, legacyPlayers=1093, rankStamps=740
bridge (page origin)  -> HTTP 200, legacyPlayers=1093, rankStamps=740
=> The payload looks serveable... points at the client, not the contract.
```

`rankStamps=740` where the old probe reported `0`. That run also shows **four 503s** in the console — the new bridge guard refusing the unstamped snapshot instead of silently serving it.

## Testing

- 3 new overlay tests (budget returns `None` rather than blocking; prefers stale over waiting; unbudgeted callers keep the unbounded wait).
- `test_sleeper_overlay_concurrency` 7/7, `test_sleeper_overlay` + `test_sleeper_live_draft` 29/29, **`test_league_routing` 27/27** (exercises `get_data` itself).
- Frontend: 1994 tests. New bridge test pins the guard, both encodings, the positive-integer rule, and — load-bearing — that the committed seed really is unstamped, so the 503 path is reachable.

## Corrections to my own earlier claims

- I said the `has_data` gate was "a symptom amplifier, not the cause". **Wrong** — it passes before the warm releases the lock, which is a genuine contributing cause. Not changed here, because fixing the stall makes the race harmless; gating on the warm is a reasonable follow-up.
- My #732 comment blamed a "30-minute stale-serve window". That leg is largely unreachable during an active suite — the stale path serves immediately and kicks a background refresh, so a constantly-firing suite keeps the entry young. It only opens if that refresh *fails*.

---

### DIFF
--- a/docs/performance-optimization.md
+++ b/docs/performance-optimization.md
@@ -108,16 +108,32 @@ page-to-page navigations. Root causes, in impact order, all verified in
 source and by measurement before changing anything:
 
 1. **Every page load ran the full ranking pipeline server-side —
-   twice.** The `/settings` default `tepMultiplier: 1.15` makes every
+   twice.** The `/settings` default `tepMultiplier: 1.15` made every
    stock user "customized" (`tepMultiplierIsCustomized` treats any
    finite number as an override), so every `fetchDynastyData` took the
    override path: `GET` base + `POST /api/rankings/overrides?view=delta`,
    and the POST rebuilt the whole pipeline synchronously ON the event
    loop, uncached (~0.75s locally, seconds on prod hardware). The hook
    mounts twice per page (AppShell + page), doubling all of it.
-   *The default was deliberately NOT changed* — posting 1.15 flat is a
-   different valuation than omitting it (ADR-015 basis curve), so the
-   fix is memoization, not semantics.
+   *The default was deliberately NOT changed at the time* — posting 1.15
+   flat is a different valuation than omitting it (ADR-015 basis curve),
+   so the fix was memoization, not semantics.
+
+   > **STALE AS OF 2026-08-06 — tense corrected above, and this matters
+   > beyond bookkeeping.** The default is now `null`, not 1.15
+   > (`frontend/components/useSettings.js:51`, changed by `67caac3b`,
+   > "Make the measured TE-basis curve reachable"), and
+   > `tepMultiplierIsCustomized` returns false for null. **A stock signed-in
+   > user therefore issues NO `POST /api/rankings/overrides` on a normal
+   > page load at all.**
+   >
+   > Read as present tense this paragraph produces a wrong causal model,
+   > and it did: it was used this session to argue that `/rankings` was
+   > blocked on an overrides recompute, which sent a CI-flake investigation
+   > down a dead end. Anything reasoning about what a page load costs
+   > should start from `useSettings.js`, not from here. Pinned by
+   > `frontend/__tests__/bridge-timeout-budgets.test.js`, which fails if
+   > the default stops being null.
 2. **Cold Sleeper overlay rebuild = ~90 sequential HTTP round-trips
    inline on `/api/data`** whenever the 15-min TTL expired (~7 of 8
    windows, since the warm-behind only runs on the 2h scrape cadence),
--- a/frontend/__tests__/bridge-refuses-unstamped-snapshot.test.js
+++ b/frontend/__tests__/bridge-refuses-unstamped-snapshot.test.js
@@ -1,0 +1,133 @@
+/**
+ * The Next bridge must not serve an unstamped disk snapshot as a contract.
+ *
+ * WHY THIS EXISTS
+ * ---------------
+ * `frontend/app/api/dynasty-data/route.js` aborts the backend fetch after
+ * a 4s idle timeout and falls back to `loadFromDisk()`. The committed
+ * scraper seed (`exports/latest/dynasty_data_*.json`) carries ~1075
+ * players and ZERO rank stamps of either name.
+ *
+ * `buildRows` fail-fasts on exactly that shape (lib/dynasty-data.js) —
+ * deliberately, so a drift-prone local blend never renders. So the
+ * fallback was manufacturing precisely the input the client is built to
+ * reject: the user got a blank board, with no error, while the backend
+ * was healthy.
+ *
+ * It is also sticky. The module-scope base-contract cache then serves the
+ * poisoned payload to every subsequent mount for its TTL, and AppShell
+ * mounts `useDynastyData` on EVERY route — so one 4s stall degrades the
+ * whole app. That is what made the E2E suite fail on a different spec
+ * each run, which in turn produced two wrong root-cause diagnoses.
+ *
+ * These tests pin the two halves of the fix: the stamp check understands
+ * BOTH encodings, and an unstamped snapshot is refused rather than served.
+ */
+import { describe, expect, it } from "vitest";
+import fs from "node:fs";
+import path from "node:path";
+
+const ROUTE = path.resolve(
+  __dirname,
+  "..",
+  "app",
+  "api",
+  "dynasty-data",
+  "route.js",
+);
+const source = fs.readFileSync(ROUTE, "utf8");
+
+// The route is a Next server module (next/server imports, filesystem
+// access at module scope), so it is asserted structurally rather than
+// imported. The behavioural half is covered by the extracted predicate
+// below, which is a verbatim copy of the route's logic — kept honest by
+// the source assertions.
+function hasRankStamps(payload) {
+  const arr = Array.isArray(payload?.playersArray) ? payload.playersArray : [];
+  for (const r of arr) {
+    if (r && Number.isInteger(r.canonicalConsensusRank) && r.canonicalConsensusRank > 0) {
+      return true;
+    }
+  }
+  const dict = payload?.players;
+  if (dict && typeof dict === "object") {
+    for (const p of Object.values(dict)) {
+      const rk = p?._canonicalConsensusRank;
+      if (Number.isInteger(rk) && rk > 0) return true;
+    }
+  }
+  return false;
+}
+
+describe("bridge refuses an unstamped disk snapshot", () => {
+  it("guards the disk fallback with a stamp check", () => {
+    expect(source).toMatch(/function hasRankStamps/);
+    // The fallback must be conditional on it, not unconditional.
+    expect(source).toMatch(/if \(hasRankStamps\(parsed\)\)/);
+  });
+
+  it("answers 503 rather than serving an unusable contract", () => {
+    expect(source).toMatch(/disk_snapshot_unstamped/);
+    expect(source).toMatch(/status: 503/);
+  });
+});
+
+describe("hasRankStamps understands both payload encodings", () => {
+  it("accepts an array-encoded payload (full/array view)", () => {
+    expect(
+      hasRankStamps({ playersArray: [{ canonicalConsensusRank: 1 }] }),
+    ).toBe(true);
+  });
+
+  it("accepts a legacy-dict payload with the UNDERSCORED field", () => {
+    // server.py pops `playersArray` from the runtime view by design, so a
+    // healthy `view=app` payload has only the dict — and the dict names
+    // the field `_canonicalConsensusRank` (data_contract.py:8346).
+    // Checking only the array is what made the old E2E diagnostic report
+    // "no rank stamps" on every healthy runtime response.
+    expect(
+      hasRankStamps({ players: { "Justin Jefferson": { _canonicalConsensusRank: 1 } } }),
+    ).toBe(true);
+  });
+
+  it("rejects the committed seed's shape: rows present, no stamps anywhere", () => {
+    expect(
+      hasRankStamps({
+        players: { A: { name: "A" }, B: { name: "B" } },
+        playersArray: [],
+      }),
+    ).toBe(false);
+  });
+
+  it("rejects null/zero ranks, matching the client's 'positive integer' rule", () => {
+    expect(hasRankStamps({ playersArray: [{ canonicalConsensusRank: null }] })).toBe(false);
+    expect(hasRankStamps({ playersArray: [{ canonicalConsensusRank: 0 }] })).toBe(false);
+    expect(hasRankStamps({ players: { A: { _canonicalConsensusRank: 0 } } })).toBe(false);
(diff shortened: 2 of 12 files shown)


jasonleetucker-code:
## Status: two real defects fixed, but this does NOT make the suite green — and I want that stated rather than discovered

Both commits are verified improvements, and the new diagnostic is already earning its keep. But a flake survives, so here is the honest scoreboard.

### What is measurably better

The diagnostic now reports, on this PR's own CI run:

```
backend (baseURL)     -> HTTP 200, legacyPlayers=1093, rankStamps=740
bridge (page origin)  -> HTTP 200, legacyPlayers=1093, rankStamps=740
```

**`rankStamps=740` where the old probe reported `0`** on every failure. The "the scrape pipeline is not stamping" false alarm — which sent four investigations at the wrong subsystem — is gone.

### What is not fixed

Still one flaky spec per run, still a different one each time (`journey-rankings` last run, `journey-settings-overrides` this run). So the bridge fallback and the overlay lock were **contributing** causes, not the whole story.

### The next lead, from evidence rather than guesswork

Both runs show the same console signature during the failing load:

```
502 (Bad Gateway) ×2
503 (Service Unavailable) ×3
```

The 502s **predate both of my commits** (they are in the earlier run too), and a 502 means the upstream was unreachable — which points at the backend's **event loop being blocked**, not at the overlay lock I just bounded.

Two things make that concrete:

1. This run's failing spec is `journey-settings-overrides`, which toggles a source and fires `POST /api/rankings/overrides`.
2. `CLAUDE.md` documents that endpoint as rebuilding **the whole ranking pipeline synchronously on the event loop**, uncached.

So the working hypothesis is: an expensive synchronous recompute stalls the loop, concurrent requests (including the board's own contract fetch) fail or time out, and whichever spec is mid-flight loses. That fits "different spec every run" as well as the overlay lock did, and it is the same shape of bug — a blocking call on a shared loop.

**I have not verified it**, and I am not going to guess a third time. Verifying it means reading `backend.log` from the `e2e-server-logs` artifact of a failing run to see whether an overrides recompute overlaps the 502s.

### What I would ask of this PR

Judge it on what it does: a 4 s backend stall no longer becomes a silently blank board, `/api/data` no longer blocks on a decorative overlay, and the diagnostic now names the responsible process instead of accusing the pipeline. It is not a claim that the suite is green.

---

jasonleetucker-code:
## Residual flake located and measured: the overrides recompute stalls the event loop for ~10s — and `run_in_threadpool` does not prevent it

I said in my last comment that I would not guess a third time and would go get evidence. Here it is. **Measured locally, not inferred.**

### The measurement

Hammer `/api/health` (a trivial endpoint) while one `POST /api/rankings/overrides` is in flight:

| | |
|---|---|
| `/api/health` baseline | **7 ms** median |
| overrides POST duration | 13.27 s |
| `/api/health` **during** it | median 83 ms, **max 10,218 ms** |
| status codes during | `{200: 16, 503: 1}` |

A trivial endpoint went from 7 ms to **over ten seconds**, and one request outright 503'd.

### Why this is the flake

- The Next bridge in front of `/api/data` abandons at a **4 s** idle timeout. A 10 s stall is comfortably past it.
- **Playwright runs specs in parallel.** One worker's overrides POST starves the other worker's page load — which is exactly "a different spec fails every run", with no need for any of the causes already fixed in this PR.
- It fits the console signature in both failing runs: `502 Bad Gateway` ×2 and `503` ×3 during the load. Those **predate both commits here**.
- Last run's failing spec was `journey-settings-overrides` — the spec that *fires* the recompute.

### The part worth knowing: the existing mitigation does not mitigate

`post_rankings_overrides` **already** wraps the rebuild in `run_in_threadpool` (`server.py:4255` and `:4266`). The stall happens anyway.

That is the GIL. `run_in_threadpool` moves work off the event-loop *thread*, which fixes **I/O-bound** blocking — but the ranking rebuild is CPU-bound pure Python, so it holds the GIL and the event-loop thread cannot make progress either. The work is "offloaded" in the sense that it is on another thread, and not at all in the sense that the server stays responsive.

This is worth flagging beyond this PR: `CLAUDE.md` describes `_serialize_overlaid_response` as fixed by the same `run_in_threadpool` treatment. If that path is likewise CPU-bound Python, it will have the same property. I have **not** measured that one — stating the concern, not the finding.

*What would refute this reading:* if the heavy work released the GIL (C extension, `json.dumps` internals), `/api/health` would have stayed fast. It did not.

### Why I am stopping here rather than fixing it

The real options are a process pool, precomputing the common override shapes, or making the pipeline cheaper — all architectural, all with a much larger blast radius than this PR. There is already a response memo (`cache_key` at `:4243`), so a *repeated* identical override is cheap; it is the first of each shape that costs 13 s, which is exactly what an E2E suite generates.

**This PR should be judged on what it does:** a backend stall no longer becomes a silently blank board, `/api/data` no longer blocks on a decorative overlay, and the diagnostic now names the responsible process. The remaining flake now has a measured cause and a named owner instead of four wrong theories.

---

jasonleetucker-code:
## Found the overrides 503. It is this bridge, not the backend — and it is a timeout.

Same defect class as this PR's first commit: a Next bridge failure wearing the backend's clothes. Pushed as a third commit.

### The mechanism

`frontend/app/api/rankings/overrides/route.js:42` aborted the backend fetch after **15 s** and returned its own 503. On this PR I measured a cold `view=delta` recompute at **13.27 s** — 88% of that budget, on hardware faster than a GitHub runner. Over the line, the bridge synthesizes a 503 that is byte-for-byte indistinguishable from the backend's own, so the spec reports `expected 200, received 503` and every investigation goes to `server.py`.

### It provably is not `server.py`

`post_rankings_overrides` has exactly two 503 branches, and **neither can fire** in the environment that failed:

1. **`not latest_data`** (`:4038`) — `_prime_latest_payload` calls `_swap_to_empty()` for falsy data (`:2032`), which sets `latest_contract_data = None`. `has_data` is `latest_contract_data is not None` (`:4570`), so this state implies `has_data: false` and CI's readiness gate would never have opened. After boot, `latest_data` is only reassigned in `run_scraper` with a truthy result.
2. **Scoring-profile mismatch** (`:4045`) — guarded by `if loaded_profile and loaded_profile != league_cfg.scoring_profile`. Both leagues in `config/leagues/registry.json` are `superflex_tep15_ppr1`, so it cannot be unequal.

The resolver's own 503 is out too: this handler calls `_resolve_league_for_request` **without** `require_loaded_contract=True`, which is the only path that raises it.

I should have checked which origin the spec talks to before adding those log lines. The spec navigates via `pageUrl("/settings")` (:3000) and waits on the **page's** request, so it never touches :8000 directly — the same wrong-origin mistake this PR's diagnostic fix exists to prevent, made again one layer up.

### What changed

- **Budget 15 s → 45 s**, chosen against the measurement (3.4× the observed cold recompute), not by feel.
- **The synthesized 503 now names its origin**: `bridge_timeout` vs `backend_unreachable`, plus `origin: "next-bridge"`, `elapsedMs`, and a message stating in words that `server.py` did not produce it.

`frontend/__tests__/overrides-bridge-timeout.test.js` pins both halves. Full suite: 119 files / 2000 tests pass.

### What this is not

**It is not a latency fix.** 13 s is a real user-facing wait, and the recompute still blocks the event loop — the existing `run_in_threadpool` at `server.py:4255`/`:4266` does not help, because the work is CPU-bound pure Python holding the GIL (measured on this PR: `/api/health` went 7 ms → 10,218 ms during one overrides POST). This raises the ceiling so a slow response stops being *reported* as a broken one. Making it fast is separate and architectural.

The two `server.py` log lines from the previous commit are kept deliberately. They are now known **not** to be this flake's cause, so their silence in a future failing run is itself the confirmation.

---

jasonleetucker-code:
## Sourced the 502, and killed a story of my own before it shipped

Fourth commit. The E2E run on `d0a260bd` failed on a **different** spec (`journey-rankings`), and this branch's own diagnostic is what made it tractable:

```
backend (baseURL) /api/data?view=app  -> 200, legacyPlayers=1093, rankStamps=740
bridge  (page origin) /api/dynasty-data -> 200, legacyPlayers=1093, rankStamps=740
=> The payload looks serveable, so the board had data and still did not render it.
console: 404, 404, 502 (Bad Gateway), 503, 503, 404, 404
```

The old diagnostic would have printed "PAYLOAD HAS ROWS BUT NO RANK STAMPS" here and sent the reader at the scrape pipeline. It now probes both origins and counts the underscored legacy-dict encoding, so it correctly reports the contract as healthy.

### The console line, fully accounted for

| status | endpoint | why |
|---|---|---|
| 404 ×4 | `/api/health` + favicon | `StaleDataBanner` (`AppShellWrapper.jsx:94`) polls `/api/health` on every page; **there is no `frontend/app/api/health` route**, so it 404s on the Next origin |
| 502 | `/api/auth/status` | 3000 ms abort (`route.js:10`) → `{status: 502}` (`:30`); `useAuth` probes it on every mount |
| 503 ×2 | `/api/dynasty-data` | by elimination — the only 503-synthesizer on `/rankings`' load path |

The elimination: `rankings/sources` (3 s→503) has **no frontend caller at all**; `/api/status` (3 s→502) is reached only from the admin panel and `/tools/source-health`; everything else is a POST engine `/rankings` never calls on load.

### What changed

A survey of all 15 bridge routes with abort budgets found **no policy** — 3 s, 4 s, 5 s, 10 s, 15 s, 20 s, 30 s, with 502 vs 503 equally arbitrary. Against a backend measured to stall ~10 s, the short end reports *busy* as *dead*:

- `dynasty-data` **4000 → 30000**
- `auth/status` **3000 → 15000**
- `journey.js` now records `msg.location().url` — Chrome puts the status in the console text and the URL only in the location, which is why sourcing the 502 needed a sweep of all 45 routes.

On the data route this is not cosmetic: abort → disk fallback → unstamped → this branch's 503 → `_fetchBaseContractNetwork` throws → `/rankings` renders its error Banner *instead of* the table. That is why the locator says "element(s) not found" rather than showing an empty tbody.

### Two stories I had to kill, one of them mine

I had already written the first of these into a comment before checking it:

- ~~"Playwright runs specs concurrently, so `journey-settings-overrides`' POST stalls the loop during another spec's load"~~ — **false**. `tests/e2e/playwright.config.js:190` sets `workers: 1` in CI and `:135` `fullyParallel: false`. Specs run sequentially. Caught it before committing; the comment now records the refutation instead.
- ~~"/rankings fires the overrides POST itself"~~ — **false**. `tepMultiplier` defaults to `null`, not `1.15` (`useSettings.js:51`, changed by `67caac3b`). **The 1.15 claim in `docs/performance-optimization.md` is stale** and should be corrected separately.

`frontend/__tests__/bridge-timeout-budgets.test.js` pins the budgets *and* the facts those refutations rest on (`workers: 1`, `tepMultiplier: null`), so the comments cannot rot into fiction. The budget parser was verified against all five routes rather than trusted — it reads 30000/15000/45000/3000/3000 correctly.

### Not claimed

**That this fixes the flake.** What is established is only that the backend was unresponsive for >3 s during that load — `auth/status`'s budget blew and produced the 502. *Why* it stalled is still open, and with `workers: 1` and no overrides POST on that path, my previous explanation is gone. This removes a class of false failure; the stall itself is architectural and untouched.

Judge the next E2E run by whether the console block now names its endpoints, not by its colour.

Suite: 120 files / 2008 tests pass.

---

jasonleetucker-code:
Pre-squash diagnostic/reference preserved. The rotating E2E/unstamped-snapshot investigation is useful history, but any residual issue must be reproduced on current main before porting a fix. Closing the stale PR; branch retained.

PR #763, opened 2026-08-06
Move the palette's ownership index into the palette itself

jasonleetucker-code:
Stacked on #761 — it needs that PR's `measure-first-load.mjs` to prove anything, and this is the first cut that instrument justified. Merge #761 first.

## The problem

`AppShell.jsx` — the **root layout** — imported `buildTeamByPlayer` from `lib/waiver-logic.js` (33.9 KB of source) for exactly one purpose: a `useMemo` that built the `teamByPlayer` ownership index and passed it to `<CommandPalette>`.

But CommandPalette is **already lazily loaded** (`useLazyComponent`, `AppShell.jsx:200`) and only reachable after `/` or Cmd+K. So the root layout was pulling the whole waiver module into the chunk every route downloads, in order to prepare a prop for a component that usually never loads.

## The change

CommandPalette takes the raw `sleeperTeams` array and builds the index itself, so `lib/waiver-logic` follows the component into its lazy chunk. Precedent already in the tree: `PlayerPopup.jsx:7,680` imports `buildTeamByPlayer` for the same reason.

## Measured, before and after, same build flags

| | raw | gzipped |
|---|---|---|
| always-loaded before | 641.7 KB | 202.0 KB |
| always-loaded after | **635.1 KB** | **199.7 KB** |
| delta on **every** route | **−6.6 KB** | **−2.3 KB** |

Chunk `2910` fell 12.0 → 5.3 KB raw. waiver-logic relocated to chunk `5145`, now preloaded by exactly two routes — `/rankings` (`page.jsx:56` imports `buildTeamByPlayer` directly) and `/waivers` (its own page) — instead of all 35.

Chunk identity was confirmed by fingerprint, not inferred: waiver-logic's four verdict literals (`"considering"`, `"marginal"`, `"obvious"`, `"reasonable"`) are the only strings in it that survive minification, and they co-occur in exactly one chunk before and after.

## Two things worth stating plainly

**The source size and the shipped size are different numbers, and only one of them reaches users.** "33.9 KB of source leaves the root-layout graph" is true and was the reason to try this. What actually ships, after minification and shared-chunk splitting, is 6.6 KB raw / 2.3 KB gz. Quoting the source figure as the win would overstate it 5×. Same failure mode as the retracted `firstLoadChunks()`, in the optimistic direction.

**The `owner:` grammar was untested.** Every existing CommandPalette case passed `teamByPlayer={null}`, so the prop was never exercised at all. Three tests were added covering the index the palette now builds. Without them this move could have silently disabled owner filtering and nothing would have failed.

## Correction to #761's framing, carried into the ledger

#761's body claims the conditional AppShell import is worth 57.7 KB raw / 19.4 KB gz. **That is an overstatement.** Traced statically from `app/layout.jsx`, only one of those four chunks hangs off AppShell:

| chunk | reached from |
|---|---|
| `2910` | `components/AppShell.jsx` — **the only edge** |
| `8805` | `useDynastyData.js` **and** `components/useLeague.js` |
| `6023` | `useDynastyData.js`, `useTeam.js`, `useTerminal.js` |
| `2063` | `useLeague.js` |

The surviving edges are the **chrome, not the data layer**: `TopBar.jsx:21-22` and `MobileChrome.jsx:21-22` statically import `LeagueSwitcher` → `useLeague` → `lib/dynasty-data`, and `TeamSwitcher` → `useTeam`. Those chunks stay on every route no matter what AppShell does, because the nav bar renders on every route.

So the honest ceiling on "make the player pipeline import conditional" is chunk 2910 alone — and after this PR that is **5.3 KB raw / 2.2 KB gz**, not 57.7 KB. The ledger now documents that alongside the hazard: a lazily-loaded `PrivateAppShell` *component* would remount every page in the app, because both shell branches render at the same position with `{children}` inside (`AppShell.jsx:141-146`), so the element type at that position changes when the chunk lands. ~2 KB gz is not worth that.

## Verification

- `npx vitest run` — **117 files, 1989 tests, all pass** (14 in the CommandPalette spec, including the 3 new `owner:` cases)
- `npm run check:bundles` — all 14 page budgets under; `/more` still has its 1.9 KB headroom, `/rankings` gained 0.4 KB of headroom
- `measure-first-load.mjs` before/after on identical `--webpack` builds; its exclusion control passed both times

## Not claimed

That this is a large win. It is 2.3 KB gz per route. It is here because it is free — no lazy-loading gymnastics, no new abstraction, one prop changed — and because measuring it is what corrected the 57.7 KB figure that was about to justify a much riskier change.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SaMgGfyHGEs38eGJ3kBf6H)_

### DIFF
--- a/docs/performance-optimization.md
+++ b/docs/performance-optimization.md
@@ -1080,14 +1080,84 @@ This also scopes what the `NO_PLAYER_DATA_ROUTE_PREFIXES` work (#759)
 achieved: it removes the multi-MB `/api/data` fetch and the ~1,100-row
 `buildRows` pass from six routes — the dominant cost, and real — but
 **zero bytes of JS**. Removing the bytes needs the import itself to become
-conditional, i.e. splitting `PrivateAppShell` behind the same imperative
-`await import()` idiom `useLazyComponent` (`AppShell.jsx:81`) already uses.
-That is the largest remaining app-owned win and it is bigger than the
-shell-deferral idea.
+conditional.
 
 **Do not reach for `dynamic()` to do it** — `AppShell.jsx:16-72` documents,
 with measurements, that `dynamic()` there puts `{children}` inside a
 Suspense boundary and permanently duplicates every page's DOM.
+
+#### CORRECTION — the 57.7 KB is NOT all AppShell's to remove
+
+The paragraph above (and the PR body it came from) implied that splitting
+the AppShell import would recover all 57.7 KB / 19.4 KB gz. **It would
+not.** Traced statically from `app/layout.jsx`, only **one** of those four
+chunks hangs off AppShell's import of `useDynastyData`:
+
+| chunk | reached from |
+|---|---|
+| `2910` | `components/AppShell.jsx` — **the only edge** |
+| `8805` | `useDynastyData.js` **and** `components/useLeague.js` |
+| `6023` | `useDynastyData.js`, `useTeam.js`, `useTerminal.js` |
+| `2063` | `useLeague.js` |
+
+The surviving edges are the **chrome, not the data layer**:
+`TopBar.jsx:21-22` and `MobileChrome.jsx:21-22` statically import
+`LeagueSwitcher` → `useLeague` → `lib/dynasty-data`, and `TeamSwitcher` →
+`useTeam`. Those three chunks stay on every route no matter what AppShell
+does, because the nav bar renders on every route. Removing them means
+deferring visible chrome and making the switchers pop in after paint.
+
+So the honest ceiling on "make the player pipeline import conditional" is
+chunk 2910 alone — and after the cut below it is **5.3 KB raw / 2.2 KB
+gz**, not 57.7 KB. Weigh that against the remount hazard before attempting
+it: a lazily-loaded `PrivateAppShell` *component* is a trap, because both
+shell branches render at the same position with `{children}` inside
+(`AppShell.jsx:141-146`). When the chunk lands, the element type at that
+position changes and **React remounts every page** — page state discarded,
+effects re-run, fetches fired twice, scroll reset. A different mechanism
+from the `dynamic()` DOM duplication, the same family of damage, and this
+one is plainly visible. A null-rendering leaf bridge that reports upward
+via callback avoids it, at the cost of real complexity for ~2 KB gz.
+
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
## E2E failure is the known pre-existing flake, not this diff

`Boot stack + run journeys` came back **141 passed, 1 failed**:

```
journey-settings-overrides.spec.js:45 › toggling a source fires the overrides request
  Error: overrides endpoint should recompute the blend successfully
  Expected: 200
  Received: 503
```

This PR changes four things: a `CommandPalette` prop (`teamByPlayer` → `sleeperTeams`), the removal of one import from `AppShell.jsx`, a test file, and a ledger doc. `POST /api/rankings/overrides` is a FastAPI handler; nothing in the diff can reach it. `/settings` calls `useDynastyData()` itself (`page.jsx:173`) and never renders the palette unless the user opens it.

It is also a repeat: this exact spec failing this exact way is run 3 of the five documented on #758, a PR that likewise could not have caused it.

**What I narrowed today.** `post_rankings_overrides` has exactly two 503 branches: `not latest_data` (server.py:4013) and a scoring-profile mismatch (`:4045`). I checked `run_scraper` on the theory that a failed scrape nulls the global — it does not; the exception path returns `None` and the partial path returns the existing data, neither reassigns `latest_data`. So a scrape failure after the readiness gate is ruled out as the mechanism.

That leaves both branches open, and neither is currently distinguishable from the outside because neither logs. I am not going to guess which — I have already published two wrong root causes for this flake (the readiness-gate mechanism, and the `contract.health.playerCount` signal), both from reasoning rather than evidence, and a third guess is not warranted. The settling evidence is `backend.log` from a failing run, or a one-line log on each branch so the artifact says which fired.

Tracked on #762. Not fixing it here — it is architectural and unrelated to this change.

**This PR's own verification is unaffected:** `Validate PR` covers what this diff touches (vitest 1989 pass, bundle budgets, ruff, pytest), and the byte delta is from `measure-first-load.mjs` before/after on identical builds.

---

jasonleetucker-code:
Pre-squash stacked PR (based on another old feature branch, not main). Do not merge or bulk-cherry-pick it. Closing as stale; re-derive the palette ownership optimization on current main only if it still measures as useful.

PR #810, opened 2026-08-12
B6 / W18: league-configuration correctness — scoring compatibility becomes a fact, and a cross-league sleeper block stops lying about readiness

jasonleetucker-code:
Closes the B6 scope in `docs/EXECUTION_PLAN.md`: **W18-F001** and **W18-F002**, one league-configuration root-cause family. W18-F003 is untouched and remains B7.

Both defects had RED coverage committed **before** any production edit (`d20abe174`); F002's RED was re-run after F001 went green to prove it still failed for its own reason. The three gaps found in owner review each got their own RED first as well.

---

## W18-F001 — the compatibility answer was a hand-typed string

`leagues_share_scoring()` was `cfg_a.scoring_profile == cfg_b.scoring_profile`. Measured from the two live leagues' own Sleeper objects:

| | dynasty_main | dynasty_new |
|---|---|---|
| `scoringProfile` | `superflex_tep15_ppr1` | `superflex_tep15_ppr1` |
| `rec` | 0.08 | 1.0 |
| `pass_td` | 6 | 4 |
| `pass_yd` | 1/30 | 0.04 |
| `pass_int` | −4 | −1 |
| `bonus_rec_te` | 0.0 | 0.5 |

35 of 48 shared keys differ; dynasty_main carries 93 keys dynasty_new does not. So `/api/data?leagueKey=dynasty_new` served dynasty_main's board.

Three problems, not one: the label is not an identity; `leagues_share_scoring()` had **zero production callers** (six scattered comparisons in `server.py` actually answered it); and four of those six treated a contract with no identity as compatible with everything.

### Repair

- **`scoring_fingerprint()`** — a factual identity over the actual scoring card. Key order, numeric form (`1` vs `1.0`) and absent-vs-explicit-zero normalized away; non-numeric metadata excluded; `None` in → `None` out, because a hash of `{}` is indistinguishable from an answer. A **new** function, not a promotion of `_scoring_hash` — that one gets all three wrong and would manufacture *false incompatibility*.
- **Storage**: a snapshot per league at `data/leagues/scoring_<sleeperLeagueId>.json`, refreshed by the post-scrape warm pass and `scripts/fetch_league_scoring.py`. **Never fetched inside a request** — an 8 s Sleeper round-trip in the `/api/data` gate would trade a correctness bug for a latency one.
- **Contract identity**: `meta.scoringFingerprint`, derived from the contract's *own* `sleeper.scoringSettings`, so it can be recomputed from the artifact it describes.
- **One gate**: `server.py::_scoring_identity_error`, used by all six sites; unproven fails closed in both directions.

### Owner review, gap 1 — staleness is not proof

A snapshot proves *"this league had these rules when the fetch last succeeded"*, not *"it still has them"*. Keeping the old card on a failed refresh is right, but unbounded it becomes this same fail-open reached through time.

`scoring_evidence_state()` now returns **fresh / stale / missing** — the vocabulary `_build_source_timestamps` already uses — and only `fresh` authorizes reuse. The budget is **derived**: the snapshot is a scrape-cadence artifact (`42 */2 * * *`), and this repo already states one staleness rule for those twice identically — `SCRAPE_INTERVAL_HOURS * 3` in `server.py`, and the 6-hour default in `data_contract._SOURCE_MAX_AGE_HOURS`. `SCORING_SNAPSHOT_MAX_AGE_HOURS = 6` is pinned against both precedents so it cannot drift from what derived it.

A second, independent boundary: a card recorded against a **different NFL season** is stale however recently it was fetched — Sleeper leagues chain year to year under new ids, so a registry entry pointing at last season's league would keep fetching a perfectly fresh card for the wrong season. Age cannot catch that.

Stale ≠ deleted: `scoring_settings_for_league` still returns the card for diagnostics; only its authority expires. Same-league requests are structurally unaffected — the gate short-circuits on a matching league key first.

### Owner review, gap 3 — the stamp must agree with the card

Preferring `meta.scoringFingerprint` without checking it against the contract's own card gave up the property that justified having a contract identity at all.

| card | stamp | result |
|---|---|---|
| present | agrees | that fingerprint |
| present | absent | recompute — the migration path |
| present | **disagrees** | **`None`** + a warning naming both |
| present | foreign `sf*` version | `None` — normalizations aren't comparable |
| **absent** | present | **`None`** |

The last row is a decision, not a fallthrough: a stamp with nothing to check it against is unverifiable. The documented migration policy never leaned on that branch — it leaned on the card, which every real contract carries (141 keys live).

---

## W18-F002 — a cross-league sleeper block published as ready

The `sleeper_matches=False` path carried `scoringSettings` / `rosterPositions` / `leagueSettings` forward from the **loaded** league, laid the requested league's teams on top, and stamped `sleeperDataReady = True` unconditionally — while also stamping `sleeperLoadedLeagueKey`, a pair the documented contract says cannot co-occur. `useTeam.js` reads the flag, so the data-not-ready state written for exactly this case never rendered.

The merge was inline in **four** handlers, which is why nothing tested it. It is now one named function with the invariant pinned:

> `sleeperDataReady` is true **iff** every league-specific field represented as ready is complete and belongs to the requested league.

Ready stays *reachable*: `fetch_sleeper_overlay` publishes `leagueConfig` read from the league object `_fetch_league_name` already fetches — no extra round-trip.

### Owner review, gap 2 — ready needs a COMPLETE config

`_fetch_league_config` built its block with `or []` / `or {}`, so a partial Sleeper response gave valid scoring beside `rosterPositions: []` and `leagueSettings: {}` — published as ready.

"Complete" is traced to the consumers, not asserted:

| field | requirement | consumer that requires it |
|---|---|---|
| `scoringSettings` | non-empty | BDVM raises without it |
| `rosterPositions` | **non-empty** | `bdvm/league_config.py` gates on `if roster_positions and scoring:` so `[]` falls through to the registry — it means *missing*; and `starter-slots.js` ranks the live list **above** the registry, so `[]` beats correct settings and yields an empty lineup |
| `leagueSettings` | present, `num_teams > 1` | the same builder raises `LeagueConfigError` at `teams <= 1` |

No empty-but-present value is legitimately complete here — every Sleeper league has slots and a team count. `_fetch_league_config` now returns `None` on a partial answer rather than passing skeletons downstream.

---

## Measured on the shipped configuration

`docs/master-site-audit/evidence/W18/b6_validate.py` — real registry, live board, no fixtures. Pinned in `b6-validation.json`.

`labelsAgree: true`, `fingerprintsAgree: false`, both leagues' evidence `fresh`.

| endpoint | dynasty_main | dynasty_new |
|---|---|---|
| `/api/data` | 200 | **503** `data_not_ready` (was 200, wrong board) |
| `/api/terminal` | 200 | **503** `data_not_ready` |
| `/api/draft-capital` | 200 | 200 |

`/api/draft-capital` stays 200 for a foreign league by design (CLAUDE.md, D-2). What changed there, measured in isolated processes because the route's own cache invalidates a same-process A/B:

| | old gate | new gate |
|---|---|---|
| `rookieSource` | `contract` | `none` |
| rookie rows priced | 40 | 0 |
| picks emitted | 80 | 80 |

dynasty_new was shown 40 rookies priced under 0.08-PPR / 6-point-TD rules on a full-PPR / 4-point-TD board. Withheld now; its real pick board untouched.

**Hot path, on a quiescent process:** evidence state 15.0 µs, league fingerprint 16.6 µs, contract fingerprint 0.6 µs, whole gate 31.0 µs. The pre-review 9.9 µs / 0.3 µs figures were taken without controlling for the `public-league-warmup` thread — under that contention bare `Path.stat()` reads 882 µs — so the two sets are not strictly comparable and these are the controlled ones. The stamp/card comparison would cost ~85 µs per call unmemoized (141 live scoring keys); it is memoized on the card's identity as a pure accelerator.

---

## Operational note for deploy

A league needs `data/leagues/scoring_<sleeperLeagueId>.json` to be *provably* compatible, and now that snapshot must also be **less than 6 hours old and for the current season**. The post-scrape warm pass writes it every cycle; on a cold deploy run `scripts/fetch_league_scoring.py` once. If the warm pass stops succeeding for more than 6 hours, cross-league requests fail closed — deliberately — while every league's own requests keep working.

## Tests

- `tests/api/test_scoring_compatibility.py` — 23 tests + 3 subtests: A/B/C/C2 semantics, the server fail-open shape, seven fingerprint-stability properties, six freshness/season-boundary cases, five stamp↔card consistency cases
- `tests/api/test_cross_league_overlay_coherence.py` — 15 tests + 5 subtests: the forbidden state, fail-closed absence, legitimate ready, NFL-wide reuse, diagnostic-is-not-the-fix, transport-is-not-the-contract, and six completeness cases
- `tests/api/test_league_routing.py` — endpoint-level pins for both findings
- Existing tests that encoded the label semantics are updated rather than worked around; two asserted the fail-open directly and now pin the opposite.

Also carries a docs-only commit subordinating CLAUDE.md to the canonical product hierarchy (`22787f936`), separable via `git show`.

https://claude.ai/code/session_012e7gcLKkmAEz8erFp5c9Qr


jasonleetucker-code:
CI status, both rounds accounted for.

**Round 1 (`ea6b90dde`) — mine, fixed.** `test_valuation_mode_threading` failed with `starter_slots_missing`. Cause was a late-binding bug in my own W18-F001 RED fixture: `addCleanup(lambda: setattr(lr, ..., self._orig_get))` reads the attribute at *teardown* time, so the one test that calls `_install` three times restored a stub instead of the real `get_league_by_key` — leaking a fake registry into every later test in the process. Bisected from 11 failures across three suites, none of which fail in isolation; the B6 baseline is clean on the same command. Fixed in `952bb8fc4` by capturing originals once in `setUp` and restoring in `tearDown`.

**Round 2 (`952bb8fc4`) — not mine; it fails on `main`.**

```
tests/consensus_edge/test_fair_value.py::TestFairValueIndex::test_the_surviving_board_is_on_one_scale
AssertionError: 1.387031408308004 not less than 1.35 : Derius Davis is 1.39x the default board
```

Checked out `origin/main` in a clean worktree with no B6 code and ran that test alone — it fails there at the **identical** value, `1.387031408308004`. It also passed on this branch before I merged main and started failing immediately after, which is the same statement from the other direction: the input is main's data, not my diff.

The trigger is the four automated refresh commits main picked up this morning (`6b7e3303f`, `1092e54d0`, `03a1b324a`, `f6e70fb91` — 47 changed data/CSV files). B6 touches no valuation code: no Hill curve, no blend, no `fair_value`, no `build_api_data_contract`.

Measured on main's current board, in case it's useful for deciding what to do about it:

| leave-one-out / default ratio | rows |
|---|---|
| ≥ 1.35 | 1 |
| ≥ 1.30 | 1 |
| ≥ 1.20 | 1 |

385 priced rows. Derius Davis is the only one over the bound (fair 1369 vs base 987, offense/WR); the next highest is Tyler Badie at 1.104 and the rest of the top six sit between 1.036 and 1.041. So this reads as one row crossing an empirical threshold on ordinary churn rather than the broken denominator the assertion is written to catch — but it is a real bound on someone else's board, and re-deriving it or marking the test `livedata` is a call for that workstream, not something I'll change from inside B6.

I've merged current main into this branch (`c28871a0f`) so the PR evaluates against the same data CI sees. Everything else is green: full Python suite 7225 passed / 25 skipped on the pre-merge tree, frontend 122 files / 2016 tests, `ruff check` + `ruff format --check` clean across 1033 files, product-plan governance clean.

---

jasonleetucker-code:
Correction to the round-2 note above, and it resolves cleanly: **CI is green at `c28871a0f`.**

The `test_the_surviving_board_is_on_one_scale` failure was pre-existing on `main` and caused by data drift, as measured — but I said the next scrape "might or might not" move it back. It did. `fc276a019` (the 11:33 refresh) landed on main a minute before that CI run started, and because the check evaluates the merge with current main, it ran against the newer board and passed. Derius Davis is back under the 1.35 bound.

So no action is needed on that test from this PR, and no threshold or marker change is warranted — it was one row crossing an empirical bound for roughly 90 minutes of board churn. Worth knowing it can happen, since CI runs `-x`.

I've merged that refresh into this branch too (`5d…`, pushed) so local and CI evaluate the same board; `tests/consensus_edge/test_fair_value.py` is 29/29 on it here.

---

jasonleetucker-code:
Re-review against current `main` after the post-incident sequence merged. Head is now `e453889aed0c1f3ca0378c408d7b8b37985e3309`.

**Three figures in the description above no longer describe this head.** Recording them here rather than editing the body, so the original review provenance stays readable.

**1. Hot-path timings — re-measured, and materially faster.** The body's numbers were taken on a different machine. Re-measured on a quiescent process with `socket.socket` / `socket.create_connection` hard-blocked, best-of-7 × 2000 calls:

| | body | this head |
|---|---|---|
| evidence-state lookup | 15.0 µs | **7.84 µs** |
| requested-league fingerprint | 16.6 µs | **7.90 µs** |
| contract fingerprint (memoized) | 0.6 µs | **0.30 µs** |
| whole gate — same league | — | **0.14 µs** |
| whole gate — cross-league (503) | 31.0 µs | **14.42 µs** |

No `NetworkTouched` was raised, so the request-time gate performs zero network I/O — the property the body claims, now verified by blocking rather than by inspection.

**2. Test counts are understated.** `test_scoring_compatibility.py` is **32 tests + 5 subtests** (body says 23 + 3), after a new class added in this pass.

**3. `b6-validation.json` gained a field.** It now carries `fingerprintsComparable` alongside `fingerprintsAgree` — see below.

---

**One defect found and repaired in this pass**, in W18-F001's own evidence validator. `b6_validate.py` computed

```python
fingerprintsAgree = len({fp for each league}) == 1
```

and `scoring_fingerprint_for_league` returns `None` for a league whose snapshot is stale, missing, or recorded against another NFL season — the fail-closed states F001 introduced. Two `None`s collapse to `{None}`, so the artifact published **`fingerprintsAgree: true` for two leagues it simultaneously recorded as unidentifiable**. Reproduced live on a container whose snapshots had aged out: both leagues `evidenceState: "stale"`, `scoringFingerprint: null`, `fingerprintsAgree: true`. The committed artifact was generated while evidence was fresh, so it never displayed the wrong value — the defect is latent and surfaces exactly when the fail-closed states fire.

Comparability is now reported separately, and agreement is `None` — never `False`, which would assert a proven *difference*, a distinct and equally untrue claim. RED committed before the repair.

Also corrected `CLAUDE.md`'s league-scoped field list, which named `sleeper.positions`. That field is the playerId → NFL-position map `buildRows` reads and does not vary by league; the league-scoped field is `sleeper.rosterPositions`. The list now states the exact tuple `sleeper_overlay.LEAGUE_SPECIFIC_SLEEPER_FIELDS` enforces.

---

**Re-verified against current `main`, not assumed:**

- the 6-hour window still follows from today's constants — `SCRAPE_INTERVAL_HOURS = 2`, `data_stale = age > SCRAPE_INTERVAL_HOURS * 3` (stated twice), `_SOURCE_MAX_AGE_HOURS` default 6, cron `42 */2 * * *`;
- #817 has **no semantic overlap**: zero file intersection with B6's 19 files, `tests/conftest.py`'s change is an unrelated `UPTIME_CHECK_ENABLED` default, and main's frontend changes reference no W18 contract field while B6 touches no frontend file;
- live registry: labels agree (`superflex_tep15_ppr1` both), fingerprints differ — `sf1:b7ad1575925091f6` (85 scoring-relevant rules) vs `sf1:82a5f8ef2bfdb098` (41). `/api/data` and `/api/terminal` for `dynasty_new` → 503; `/api/draft-capital` → 200, contract preserved.

Merged current `main` (`f04ee88019cb`) into the branch with zero conflicts; the previous head's only check run was **cancelled**, so this is the first valid CI this PR has had.

---

PR #822, opened 2026-08-13
One canonical value: withdraw the unvalidated league lens, remove the device-local methodology switch

jasonleetucker-code:
Fixes a report that **player values differed between mobile and desktop**.

Two commits, deliberately: `04b8b6d` is the RED provenance (kept — a neutral runner reproducing the defect), `740bbf5` is the repair.

## Decision: C — league-aware valuation is not yet trustworthy as canonical

Evaluated for promotion and **rejected** under the outcome-evidence bar. Full record: [`docs/valuation/LEAGUE_AWARE_METHODOLOGY_REJECTION.md`](docs/valuation/LEAGUE_AWARE_METHODOLOGY_REJECTION.md). This is a verdict on *this implementation* and *today's evidence*, not on league-aware valuation as a goal.

## Correction to the record

**9,991 → 12,489 was never a production measurement.** It came from applying the documented ±25% cap to 9,991, and a CI fixture used `1.25` to reproduce that arithmetic. CI proved the *mechanism*, not a live value.

Measured from the captured live overlay (709 rows, **7 distinct factors**): the QB factor is **1.018366**, so Josh Allen is **9,991 → ~10,175**. Finding W07-F008 independently records **10,171**. The cap has never bound on live data. The defect was real; I overstated its size.

| position | live factor |
|---|---|
| DL | 1.097760 |
| RB | 1.041554 |
| WR | 1.026798 |
| QB | 1.018366 |
| TE | 1.011595 / 0.979705 |

## Why it was rejected

- reads **current roster state**, not durable structure — `top` is the single highest-`rosValue` player rostered *today*, so one trade moves every player at that position
- its input is a 0-100 **logarithmic rank index** ("not points, and not projection-aware") setting a multiplier on a 0-9999 board
- `reference_lineup_scarcity = 0.5` is a bare constant that alone decides the **sign** of every adjustment
- per-position only: 5 of 8 positions get a pure scalar — untestable within position by construction
- multiplies into a scale whose maximum is the Hill **asymptote**, so the board leaves its declared range
- `tePremium` is ABSENT here with a pinned double-count guard because the anchor is already KTC's TE++ board; **`structuralScarcity` has no equivalent guard** while all 18 offense-bearing sources are already Superflex boards
- and the evidence to validate a replacement does not exist

Worth stating plainly: the canonical board was **never "market-only"** — its TE-premium multiplier is already derived from the league's own Sleeper `bonus_rec_te`. Market vs My League was a false dichotomy.

## The repair — two chokepoints, not fifteen call sites

- **`_valuation_scoped_contract`** is the one place the lens reached an engine. It no longer applies it. Every caller keeps its plumbing and keeps stamping what it was *served*, so a future validated methodology re-opens **one** seam instead of nine.
- **`readValuationMode()`** always answers `market`, making persisted `next_settings_v2.valuationMode` inert with **no migration step** — an old phone storing `leagueAdjusted` converges on next load.
- The request is **ignored, not refused**. Refusing would turn an obsolete localStorage value into a broken page for a user who never chose it.
- **`adjusted_rows`** still computes, but writes `experimentalLeagueAdjustedValue`/`Rank`/`Tier` and never `rankDerivedValue` or its aliases.
- The **Market / My league control is removed** from `/rankings` — not disabled, since a greyed-out control implies the choice still exists.

**`publish.py` was repaired alongside**: it read `canonicalConsensusRank` off the adjusted board, so after the split it would have published the *market* ordering under an "adjusted" label. Caught by tests, not by review.

## History reconciliation

An earlier production reading said 8 days / 8,749 rows; a later local reading said the DB does not exist. **Both were true about different machines** — `data/` is gitignored, so a container clone has none. Verified read-only on production:

```
/home/*/trade-calculator/data/board_history.sqlite
8749 rows, 8 days, 2026-08-06..2026-08-13, 1092-1095 rows/day
timer: installed / enabled / active, Result=success
```

The recorder is legitimate and firing daily. **No second recorder was created.** It records the canonical value and methodology version; it does *not* record `league_key` or a scoring fingerprint — a narrow additive forward-only gap, deliberately left for follow-up rather than bundled here.

## Validation

| gate | result |
|---|---|
| backend `pytest tests/` | **7411 passed / 43 skipped / 0 failed** |
| frontend `vitest run` | **2025 passed / 123 files** |
| `test_canonical_value_invariance.py` | **16 passed** (was 4 failed) |
| ruff format + check | clean |

The GREEN contract no longer asserts "9,991 must survive forever". It asserts one methodology, device independence, overlay isolation, no user-selectable basis, and the compact/array transport parity that was already proven.

## Deliberately not in scope

Follow-ups, each named rather than silently skipped: `siteWeights` canonical ownership; `board_history` league/fingerprint columns; the full producer/consumer audit; build/revision diagnostics; the manual `CACHE_VERSION` bump hole.

### DIFF
--- a/docs/valuation/CANONICAL_VALUE_CONSUMER_AUDIT.md
+++ b/docs/valuation/CANONICAL_VALUE_CONSUMER_AUDIT.md
@@ -1,0 +1,114 @@
+# Canonical player value — producer / consumer audit
+
+**As of PR #822** (canonical value uniformity). Purpose: prove the
+mobile/desktop defect class is closed, not just its two known instances.
+
+The defect had one shape — **a device-local setting selected a
+non-canonical methodology, and that methodology wrote the canonical
+field**. Three mechanisms had to line up. This audit checks that no
+fourth path exists.
+
+## Producers
+
+There is exactly one.
+
+| module | role |
+|---|---|
+| `src/api/data_contract.py::_compute_unified_rankings` | **THE canonical producer.** Rank → percentile → Hill, plus value-direct voting for `ktcSfTep` / `idpTradeCalc`. Emits `rankDerivedValue`. |
+| `src/league_intel/overlay.py` | writes `rankDerivedValue` **only on a private scratch list** that never leaves the function, so the existing ranker can sort the experimental board. Returned rows carry canonical untouched. |
+
+Enforced by `tests/api/test_canonical_ownership_protections.py::test_only_approved_modules_assign_the_canonical_value_field`,
+which scans all shipped `.py`/`.js`/`.jsx` and fails on any other writer.
+
+## Canonical field contract
+
+Measured on the live 1,092-row contract:
+
+| field | relationship | status |
+|---|---|---|
+| `rankDerivedValue` | the canonical value | **CANONICAL** |
+| `values.overall` | exact alias, 1092/1092 | **CANONICAL** (pinned) |
+| `values.finalAdjusted` | exact alias, 1092/1092 | **CANONICAL** (pinned) |
+| `values.displayValue` | exact alias, 1092/1092 | **CANONICAL** (pinned) |
+| `values.rawComposite` | differs on 1080/1092 | **RAW SOURCE** — legacy scraper composite, honestly named |
+| `experimentalLeagueAdjustedValue` | `canonical × scarcity factor` | **EXPERIMENTAL** |
+
+The alias equality is **intentional and now pinned**. The server overlay
+used to scale `rankDerivedValue` and leave `values.*` at market —
+publishing a row that disagreed with itself — which is why this is
+asserted rather than assumed.
+
+## Consumers
+
+Every surface below takes its value from the canonical contract
+(`useDynastyData` → `fetchDynastyData` → `buildRows`) or from a
+server-side engine reading `latest_contract_data`. **No surface computes
+or rescales a value of its own** — the no-frontend-ranking-engine rule.
+
+| surface | source | field | was `valuationMode` reachable? | was `siteWeights` reachable? | experimental reachable? | status |
+|---|---|---|---|---|---|---|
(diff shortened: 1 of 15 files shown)


jasonleetucker-code:
## CI is red as intended — and it independently reproduced the defect

`Validate PR` failed on `04b8b6d`. I checked the whole log rather than assuming, because a RED head is only useful if it fails for the reason it claims to.

It failed on **exactly one assertion, and it is the intended one**:

```
AssertionError: Josh Allen: the league-adjusted lens overwrote rankDerivedValue (9991 → 12489)
assert 12489 == 9991
1 failed, 192 passed, 278 deselected
```

This is worth more than a local test run: a neutral runner, from a clean checkout of `main` plus one test file, produced **the exact numbers predicted from the methodology** — canonical 9,991 becoming 12,489 under a `1.25` scarcity factor. The mobile/desktop divergence mechanism is now reproduced in CI rather than argued from code reading.

Two details worth recording:

- CI runs `pytest -x`, so it stopped at the first failure. The other three RED assertions were never reached — the full local result is **4 failed, 7 passed, 1 skipped**.
- **Every other gate is green**: `ruff format` (1,044 files), `ruff check`, the decision-coercion gate (*"no new coercions… in the files this change touches"*), the audit drift gate (*"no drift: every recorded status matches its probe"*), the runtime import gate, and the livedata suite (253 passed, 25 skipped). So the harness breaks nothing — it only asserts the invariant that is currently false.

The 12,489 also breaches the documented 1–9999 canonical scale, from the same unclamped `int(round(base * factor))` at `overlay.py:92`. That was tracked separately as the B9a ceiling defect; it is not a separate bug, it is this one.

**This will not be driven to green by weakening the assertions.** The repair is to stop the contextual value impersonating the canonical one: preserve canonical, give the lens its own field, and make both appliers agree. That lands as a follow-up commit on this branch, and only the GREEN head merges.

---

Issue #839, opened 2026-08-14
Canonical meaningful roster value = ceil(1.5× starter demand), including FLEX


jasonleetucker-code:
## Owner-approved site-wide invariant — 2026-08-14

Replace hard-coded per-position "meaningful roster" counts with one league-config-driven rule wherever the product claims to measure a fantasy team's overall dynasty roster value/strength or a derivative of that meaningful core.

### Canonical rule
For every starting slot class in the actual league configuration:

`meaningful_slots = ceil(1.5 × real_starting_slots)`

Examples:
- 2 QB starters → 3 meaningful QB slots
- 2 RB starters → 3 meaningful RB slots
- 3 WR starters → 5 meaningful WR slots
- 2 TE starters → 3 meaningful TE slots
- 3 DL starters → 5 meaningful DL slots
- 3 LB starters → 5 meaningful LB slots
- etc.

This supersedes treating QB3/RB3/WR5/TE3/DL5/LB5/DB5 as permanently hard-coded numbers. Those counts are simply the current league's result of the universal 1.5× rule where applicable.

### FLEX / Superflex / IDP flex handling — OWNER CLARIFICATION
Regular flex demand is calculated separately from dedicated-position demand.

#### Offensive FLEX
- `meaningful_offensive_flex_slots = ceil(1.5 × real_offensive_flex_slots)`.
- First select the meaningful dedicated-position cores using the same 1.5× rule.
- Then fill the offensive FLEX meaningful slots with the **highest-valued remaining players** who are legally eligible for that league's offensive FLEX slot.
- "Remaining" means the player has not already been counted in a dedicated meaningful-position core.
- Players are chosen by canonical My League value, subject to the slot's real eligibility.

Examples:
- 1 offensive FLEX → 2 meaningful FLEX players from the highest-valued remaining eligible RB/WR/TE/etc. players.
- 2 offensive FLEX → 3 meaningful FLEX players from the highest-valued remaining eligible players.

#### IDP / defensive FLEX
- `meaningful_defensive_flex_slots = ceil(1.5 × real_defensive_flex_slots)`.
- After dedicated DL/EDGE/LB/DB meaningful cores are selected, fill defensive FLEX with the **highest-valued remaining legally eligible IDP players**.
- Preserve the league's actual IDP-flex eligibility rules.

#### Superflex exception
**Superflex is not treated as a generic FLEX bucket for this methodology. It is always treated as QB demand.**

[...]

Issue #829, opened 2026-08-14
Add manual-external-AI Weekly Report Studio workflow


jasonleetucker-code:
## Owner decision — 2026-08-14

Build the weekly pregame/postgame report system so the site does **not** require automatic LLM/API generation by default.

### Default generation mode
**Manual External AI** is the default. The site automatically assembles all deterministic league/report data into one complete generation package; the owner manually copies that package into ChatGPT/Claude (or another chosen AI), generates the full weekly output externally, then pastes/imports the structured result back into the site for validation, preview, rendering, and publication.

This must require **no manual report writing** and should avoid site-side AI/API credit consumption when Manual External AI is used.

### Required architecture
`DATA -> PACKAGE -> EXTERNAL AI -> IMPORT -> VALIDATE -> RENDER -> PUBLISH`

The report system and AI provider must remain decoupled.

### Pregame package
One generation should be able to produce the full weekly pregame package, including the week overview, Game of the Week, all matchup previews, players to watch, upset/watch storylines, standings/playoff/rivalry implications, and graphic headlines/copy.

### Postgame package
One generation should be able to produce the full weekly postgame package, including the weekly recap, Game of the Week recap, all matchup recaps, player/team superlatives, biggest upset/bad beat/miracle result, standings movement, playoff implications, and graphic headlines/copy.

### Deterministic inputs
The site should precompute and package every objective fact it already owns or can canonically derive: teams/logos/records, standings, scores, projections, custom-scoring/best-ball outputs, matchup history, rivalry/division context, transactions, power/odds context, injuries/status where canonically available, key performers, over/underperformance, and other approved public-safe facts. Missing data remains explicit; the AI must not invent unavailable facts.

### Import contract
External AI output should use a strict versioned structured schema (prefer JSON) that maps to canonical report sections and graphic-copy fields. Import must schema-validate, reject/flag malformed or unknown fields, preserve provenance/generation metadata, and provide Preview before Publish. A bad import must not partially publish.

### Graphics
Weekly graphics should be deterministic reusable Premium Sports Intelligence templates (HTML/CSS/SVG/Canvas or the site's canonical share renderer), **not** AI-generated images. AI may supply bounded text fields such as headline/subheadline/storyline; the site renders the actual branded graphic.

### Supported modes
1. **Manual External AI — DEFAULT**: package/copy -> external generation -> import; no site-side LLM call.
[...]

PR #809, opened 2026-08-12
docs: add source intelligence, playoff, Game Day, awards, and trade-history plans

jasonleetucker-code:
## Purpose

Capture owner-approved methodology directions in canonical planning docs without changing production behavior.

## What this adds

- **Multi-format dynasty source archive / normalization spec**: source capability discovery, raw observation provenance, historical paired snapshots, source-family deduplication, source-specific format curves, target-league scarcity/PAR integration, IDP variants, and future validation;
- **KeepTradeCut calibration treatment** for Off / TE+ / TE++ / TE+++ and 1QB/Superflex combinations as same-source calibration anchors rather than independent votes;
- **Hard dynasty-only eligibility gate** for canonical dynasty valuation inputs;
- **Separate Redraft/ROS Intelligence spec** allowing verified seasonal/redraft evidence for ROS Strength, playoff/championship odds, Pick Forecast inputs, contender/rebuilder classification, Game Day and other current-season forecasts while prohibiting leakage into canonical dynasty value;
- **Canonical Playoff Predictor spec** requiring one league-aware seasonal probability engine that derives the requested league's real standings and postseason structure and prominently reports **Make Playoffs %**, **Earn Bye %**, and **Win Championship %**;
- **League-median support as first-class standings logic**: when the league enables an extra weekly game/result against the league median, each simulated week must derive that median from the same simulated league-wide score set, award the extra result, reproduce actual record-to-date without double counting, and carry those standings into playoff/byes/championship/Pick Forecast outputs;
- **Primary-league playoff fixture**: Sleeper/host data remains preferred source of truth; the owner's confirmed fallback/regression fixture is 12 teams, **7 playoff berths, one #1-seed bye, league-median standings enabled**;
- **Game Day probability spec** preserving/auditing `Win Matchup %` and adding **Beat League Median %** from the same joint league-week simulation, with historical calibration and optional 2-0 / 1-1 / 0-2 joint-outcome probabilities;
- **Awards/Honors eligibility spec**: League MVP and Manager of the Year require actual playoff-position/qualification plus a >.500 record under the league's real standings rules; GM of the Year and OPOY/DPOY/ROY/positional awards do not inherit that gate by default;
- **Trade History Aging spec** defining Current Grade vs At-the-Time Grade vs How It Aged, requiring contemporaneous player and pick values, stable identity, provenance, fail-closed missing-history behavior, same-methodology replay, resolved-pick lineage rules, and historical validation of aging-label thresholds;
[...]

PR #875, opened 2026-08-17
C0-R: authorize the continuous C-Series campaign, repair governance drift

jasonleetucker-code:
First unit of the continuous C-Series campaign authorized by the owner directive of 2026-08-17. **Documentation and standing-invariant tests only — no source behaviour changes.**

## Why this had to go first

`docs/EXECUTION_PLAN.md` §0 answered *"what may I build right now?"* with **"Nothing"** — C1A units 1, 2, 3, 4 and 6 closed and the authorization spent. A session reading the canonical front door would correctly refuse to work. Nothing else in the campaign can start until the one record that may carry an authorization carries this one.

## Authorization

`EXECUTION_PLAN.md` §0 rewritten: the campaign is the single authorized scope, and §0.3 names the single current unit — so §7.6's *one record, one answer* rule still holds, satisfied by there being one **Current unit** line rather than one unit in flight.

New §0.2 makes the closure semantics explicit: **`CLOSED-PENDING-PROD` is implementation-complete, not closure.** Full `CLOSED` requires production proof against the **deployed merge SHA**; C10 may not count pending units; the reserved completion phrase stays unusable while any required proof is outstanding; PR-head CI is necessary and is not a substitute.

§2's C1-U6 STOP is marked **superseded in place** rather than deleted — the closure it records is still true, only its forward-looking clause was discharged. §6 (product work that must not preempt foundations) is explicitly still binding: the directive authorized the C-Series *in dependency order*, not starting anywhere in it.

## Owner methodology authority — the substantive half

New `docs/C_SERIES_DIRECTIVE_RECONCILIATION_2026-08-17.md`. A restatement of scope inside a large directive can silently reintroduce a rejected item or replace a validated implementation with an older description of it. Four topics resolved on cited evidence, never on which rule is easier to build:

| topic | resolution |
|---|---|
| **Meaningful roster core** | The fixed `QB3/RB3/WR5/TE3/DL5/LB5/DB5` caps are in `MASTER_PRODUCT_PLAN` §4.1, self-labelled **"initial"** and explicitly *"must prove/consolidate competing existing strength notions before becoming canonical"*. Addendum **#839 (2026-08-14)** is a dated **OWNER DECISION** that names them and states it replaces them. #839 wins — and ships as **V1 champion labelled PRIOR** with the 1.25× / 1.50× / 1.75× / data-derived challenger pass. The 2026-08-15 calibration policy did **not** originate the rule; it says *"Keep"*. |
| **FLEX** | No conflict. T-NEW-19 already states the directive's separate-supplement concept **and** supplies the rounding the directive left open: `ceil(1.5 × real flex slots)` per flex family after dedicated cores, **Superflex folded into QB demand first**, each player counted once. |
| **KTC VA** | The directive quotes the constants of `ktcRawAdjustment` — **V12, a regression-fit approximation**, replaced 2026-04-26 by a verbatim port of KTC's own `processV`, with the V12 route since marked deprecated and off the live path. Implementing the quoted constants would regress from KTC's real algorithm to an old fit of it, so `C3-U2` consolidates onto **the port**. |
| **Future-pick decay** | Measured per-cell vendor year-step, **still classified PRIOR** — the 2-out→3-out extrapolation is untestable on current evidence. |

**Scope: unmapped = 0, and no manifest row added.** Part XVIII (the 1–104 ledger) is the already-traced source family F; Part XIX A–J all resolve to existing rows. **`X-02` and `X-07` stay OWNER-REJECTED**, restated so directive boilerplate cannot revive them.

## Drift repaired — all of it measured

- **execution map §18** listed `C1-U3` as *"not authorized"* after C1-U3, C1-U4 and C1-U6 had all closed. Its bottom line stayed accidentally correct while its reasoning rotted.
- **execution map §12** declared **153** rows and summed to 153 against a manifest of **163** — dropping `C7-BEST-TRADE`, `C8-A11Y-01`, `C9-V3-01` and `X-01`…`X-07`. That is the identical ad-hoc-regex defect `check_execution_map`'s own docstring documents, reproduced one section away from the fix, and the rows it dropped are exactly the out-of-scope dispositions a later session would "rediscover".
- **manifest** rows for the closed `C1-U1` and `C1-U2` still carried pre-work statuses, and the canonical-owner map still said player identity was *"3 independent matchers"*.
- **`PLANNING_DOCUMENT_STATUS` §6** called #809/#816/#835 *"open and untouched"*. **All three are closed unmerged** (verified against GitHub: 02:53:35Z, 02:53:57Z, 02:54:17Z on 2026-08-15). Nothing was lost because their content had been promoted to `main` first — traceability criterion 6 doing precisely its job.
- **#839 provenance** was mis-cited to *"owner decisions 47–49"*, which are #829's Weekly Report Studio decisions; `#839` has no intake-ledger row at all.
- **superseded-rule census** said **5** (manifest) and **3** (traceability), and omitted the meaningful-core supersession — the one a C2 implementer most needs to find. Both now say **6** and enumerate the same six.

## Guarded

`tests/docs/test_governance_census_consistency.py` — **RED→GREEN: 5 failures against the pre-repair docs (`git stash` + run), 22 passing after.** Every count is derived from the rows via the integrity script's own `parse_manifest`, never read from prose.

Two of the assertions were **wrong first** and are fixed here: they tripped on §12's and §18's explanations of their own corrections. Narrowed to the opening claim, the table's total row, and each blocked item's *subject* — a closed unit may legitimately appear in an item's explanatory clause. A test that cannot tell a defect from its own fix is not a guard.

## Verification

- `scripts/check_planning_integrity.py` — OK (11 checks)
- `scripts/check_product_plan_governance.py` — clean
- `tests/docs/` — 22 passed
- `ruff check` / `ruff format --check` on the changed file — clean
- full `pytest -m "not livedata"` — reported in a follow-up comment on this PR

## Production verification checklist (for `CLOSED`)

Documentation-only, so there is no runtime behaviour to verify. This unit is `CLOSED` once the merge lands and `Deploy Production` passes its `--lane full` contract gate on the merged tree — confirming the docs change shipped without disturbing the deploy path.

---


jasonleetucker-code:
## CI is red on the base branch too — and it found a live defect on `main`

**Not this PR.** This diff is documentation plus one Markdown-reading test file; it cannot touch pick arithmetic.

### Proven pre-existing

Hard-gate failure on run 31987737878:

```
FAILED tests/api/test_pick_completeness_red.py::TestRed3UncalibratedCloneDiscountRetired
       ::test_derived_2029_value_uses_measured_year_step
AssertionError: 0.9857058938711573 not less than 0.95 : 2029/2028 ratio 0.986
```

Reproduced in a clean worktree of pristine `origin/main` (`8af0967`), with none of this branch's commits present:

```
AssertionError: 0.9857058938711573 not less than 0.95 : 2029/2028 ratio 0.986
```

Identical to sixteen significant figures. It is the same payload (`exports/latest/dynasty_data_2026-08-17.json`) and the same tree CI evaluated, minus this PR.

### It is not a threshold quibble — the year step has inverted

Measuring all twelve 2028→2029 tier cells on `main`:

| cell | 2028 | 2029 | ratio |
|---|---:|---:|---:|
| Early 4th | 1629 | **1635** | 1.004 |
| Mid 2nd | 2995 | **2997** | 1.001 |
| Mid 4th | 1538 | **1590** | 1.034 |
| Late 1st | 4102 | **4132** | 1.007 |
| Late 3rd | 1991 | **2003** | 1.006 |

**5 of 12 cells price the 2029 pick at or above its 2028 equivalent.** A pick one year further out is worth *more* — which is precisely what `C1-PICK-01`'s year-step model exists to prevent, and it is live on the board right now.

### Cause

```
2028 Early 1st  value=5107  sources=['idpTradeCalc', 'ktc', 'ktcSfTep']
2029 Early 1st  value=5034  sources=['idpTradeCalc', 'ktc']          <- no ktcSfTep
```

**0 of 24 2029 rows carry `ktcSfTep`**, while every 2028 template row does. `ktc` is the scraper's own model composite, retired from voting on 2026-04-28, so the 2029 rows are effectively resting on one voting family where their templates rest on two. The livedata lane says the same thing from the other side — four failures reporting 2029 tier rows as unexplained single-source with `matchedSources: ['idpTradeCalc']`.

Board-wide, `ktcSfTep` covers 501 rows against `idpTradeCalc`'s 911 in this scrape.

### Disposition

This belongs to `C1-U6`'s territory (pick completeness / the measured year step), not to `C0-R`, and it is an always-open-lane production defect rather than something to fix inline in a governance PR. Two things worth deciding separately:

1. **The defect** — 2029 tier rows losing their `ktcSfTep` leg, and the resulting inverted year step reaching the published board.
2. **The gate** — `test_derived_2029_value_uses_measured_year_step` asserts a ratio over the *live* board inside the hard gate, which is the class `docs/ops/STABILIZATION_2026-08-16.md` §3d says must not block there ("assert the invariant, build from `archive_fixtures`, or classify it `livedata`"). Right now one vendor's coverage can turn every open PR red — which is exactly the contamination that pass set out to end, surviving in a test the pass did not reclassify.

I will re-check and re-run when the base recovers.

---

PR #876, opened 2026-08-17
C1-U5 / C1-CONF-01: confidence fields that mean what they are named

jasonleetucker-code:
**Stacked on [#875](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/875) (C0-R)** — review that first; this PR's diff is the two commits after it.

INFRA, and **explicitly not a methodology change**: `CONFIDENCE_LEVELS` stays at exactly four and the five-axis bottleneck is preserved verbatim per `MATH_MODEL_CALIBRATION_POLICY_2026-08-15.md` §7. What was wrong was the *vocabulary* — and vocabulary is what every later consumer reads, which is why `EXECUTION_PLAN.md` §1 requires this before any new confidence consumer is built. C2, C3 and C6 all add several.

## Measured first, on the live payload

All five classes reproduced through the production build path in `test_confidence_naming_red.py` before anything changed:

| RED | defect | measured |
|---|---|---|
| 1 | priced rows wearing the row-builder placeholder | **24 rows** with a finite value and `"None — unranked"` — all 2026 round-5/6 slot picks, **all 24** carrying `pickRookieAnchor` |
| 2 | `"none"` meant four things at once | those 24 **plus 261 genuinely unpriced** rows — 285 rows, one word |
| 3 | `identityConfidence` graded resolution | exactly `{1.0, 0.95, 0.7}` across 1,110 rows |
| 4 | `marketConfidence` couldn't express its name | span `[0.3252, 0.59375]` — never enters the top 40% of a 0–1 scale |
| 5 | the owner imported its consumer | `_compute_pick_confidence` in `data_contract.py`, imported back by `confidence.py` |

### The mechanism behind RED-1

A pick with no voting source never enters `row_normalized`, so Phase 4 (capped at `OVERALL_RANK_LIMIT`) and the off-cap Phase 4b′ pass (which skips `derived <= 0`) both miss it — and then `_anchor_current_year_picks_to_rookies` prices it anyway, **by design and with a comment saying so**, writing no confidence field. The value was right; the label said the row was *unranked and therefore unconfident* when the truth was that nothing had ever looked at it.

## The repair

**`confidenceBasis` is a second axis, not a fifth level.** The overall level is the *weakest* axis, so a fifth would change what the bottleneck `min()` means for every axis at once. What was missing was never a degree of confidence — it was *which owner decided it, from what class of evidence*.

**The hole, not just the rows that fell through it.** `validate_api_data_contract` now **errors** on a priced row whose basis is missing, unknown, or self-contradicting, scanned over the whole array rather than the shape checks' 1000-row prefix — the measured population sat past it. The next pass that prices a row without saying why fails the build.

**Scope discipline.** The anchor pass runs over all **72** current-year slots. A first cut stamped the tether on all of them, which would have downgraded 48 picks' real pick-market confidence to a derivation label — a methodology change wearing a rename's clothes. Scoped to rows nothing assessed. The deeper question it exposes (the tether *overwrites* a value whose confidence came from the pick market — should those 48 still claim `pick_dispersion`?) is recorded as a follow-up, not decided here.

**The renames, dual-written.** `identityResolutionConfidence` / `...Method`, and `marketBreadthAgreementIndex` with the breadth and agreement halves the scraper computed and discarded — publishing those is what keeps the rename from being cosmetic. The scraper change is an identifier rename plus two extra return values; `conf` and `cv` are byte-identical, which matters because `market_conf` is **not** purely diagnostic there — it feeds `_elite_expansion_multiplier`, the single-source discount, the IDP cap headroom and `elite_cap`.

Three lanes with different lifetimes: additive dual-write on the payload (**`CONTRACT_VERSION` deliberately does not bump** — additive is not breaking, and bumping for it trains consumers to ignore bumps); a machine-readable `meta.deprecations` block; and archive readers that stay **bilingual permanently**, because `exports/archive/*.zip` is immutable evidence written under the old spellings. An alias is a temporary promise to writers; bilingual reading is a permanent property of a reader of immutable history.

`test_confidence_rename_aliases.py` pins the set in **three directions** — declared, emitted, and a frozen literal. An undeclared alias is exactly how a "temporary" dual-write becomes permanent.

**Not a consumer, a homonym:** `scripts/crawl_ffpc_sharp.py` reads `identityConfidence` from a `curatedManagers` record feeding `platform_ledger.upsert_manager` — *manager*-identity confidence, a different domain. Renaming it would have been a real defect.

## Blast radius, measured before and after

- **0 values moved.**
- **24 buckets** `none` → `low`, every one a 2026 round-5/6 slot pick — exactly the RED population. The other 48 anchored picks keep `pick_dispersion`.
- **0 rows** where a legacy alias disagrees with its replacement.

Honest classification: the map calls this INFRA and no value moves, but the confidence chip text *does* change on those 24 rows. "No user-visible change" would be false.

## Verification

- `pytest -m "not livedata"` — **7,653 passed, 40 skipped, 0 failed** (baseline 7,615)
- frontend `npm test` — **126 files, 2,051 tests passed**
- `ruff check` / `ruff format --check`, `check_planning_integrity.py`, `check_product_plan_governance.py` — clean
- 33 confidence tests across the three files

Note: `main` currently carries an unrelated pre-existing hard-gate failure (2029 picks missing `ktcSfTep`, inverting the year step) — diagnosed in [#875's thread](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/875#issuecomment-5311097554). It will surface here too until the base recovers.

## Production verification checklist (for full `CLOSED`)

1. `GET /api/data` — every priced row carries a `confidenceBasis` from the closed set; no priced row labelled `"None — unranked"`.
2. The 2026 round-5/6 slot picks read `low` / "tethered to the rookie at this slot".
3. `meta.deprecations` present; each legacy key equals its replacement on a live row.
4. After the first post-deploy scrape, `marketBreadthScore` / `marketAgreementScore` populate (they are `None` on pre-change exports — missing, not zero).
5. `/rankings` and the terminal render confidence unchanged apart from those 24 pick rows.

Record: `docs/confidence/C1_U5_CONFIDENCE_NAMING.md`

---


jasonleetucker-code:
## First CI run failed on the format gate — mine, fixed in `d7b638f`

```
Would reformat: tests/api/test_confidence_naming.py
Would reformat: tests/api/test_confidence_naming_red.py
Would reformat: tests/api/test_confidence_rename_aliases.py
3 files would be reformatted, 1108 files already formatted
```

**Cause: two ruff installations disagreeing.**

```
/root/.local/bin/ruff   0.15.8   <- what bare `ruff format` runs
python -m ruff          0.6.9    <- what requirements-dev.txt pins, and what CI runs
```

I had formatted with the PATH binary and check-passed against it. The pinned version formats parenthesised assertion messages differently, so the files were unformatted by CI's standard. Re-formatted with `python -m ruff`; repo-wide check is now clean at 1111 files, `ruff check .` passes, and the 55 affected tests still pass.

The same mismatch explains an earlier local run reporting *83* files needing reformat while CI reported none — I was reading a different tool's opinion of the whole repo.

## This PR is **not** blocked by the 2029 pick defect

Worth stating plainly, because I said the opposite when I opened it. This PR's base is `claude/c-series-c0-r`, which carries `dynasty_data_2026-08-16.json`; #875's base is `main`, which has moved to the 08-17 scrape where `ktcSfTep` is missing from every 2029 row. So the livedata lane here passed cleanly — **299 passed, 25 skipped** — and `test_derived_2029_value_uses_measured_year_step` is green on this tree.

That is a difference in base data, not evidence the defect is fixed. It is still live on `main` and still needs the disposition described in [#875's thread](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/875#issuecomment-5311097554); it will resurface here the moment this branch takes `main`'s newer data.

One advisory, expected for a stacked branch: `check_release_candidate` reports 62 class-C changes on `main` not in this head. That is the data drift the release-candidate discipline exists to make visible, and it is advisory on PRs by design — strict only at the release moment.

---

PR #877, opened 2026-08-17
C1-U6 D1 (report only): the board is pricing 2029 picks at their 2028 price

jasonleetucker-code:
**Report only. No code change.** This records a live canonical-value defect on `main` and hands the repair a diagnosis, evidence and a named set of wrong answers to avoid.

It came out of independently re-checking why `test_derived_2029_value_uses_measured_year_step` reddens every PR based on `main`. **The test is correct and must not be touched** — it found a real defect.

## What is wrong

**Nine of eighteen `(tier, round)` cells price the 2029 pick at or above its 2028 equivalent**, and every 2029 row is stamped `pickValueProvenance: {"class": "direct_market_blend"}` — a claim that a market priced it.

No market priced it. Every 2029 value is a **verbatim copy of the 2028 value**.

| cell | 2029 | 2028 | ratio |
|---|---:|---:|---:|
| Mid 4th | 1590 | 1536 | **1.0352** |
| Late 1st | 4132 | 4109 | **1.0056** |
| Early 4th | 1635 | 1628 | **1.0043** |
| *(6 more ≥ 1.0)* | | | |
| *(the other 9)* | | | 0.964 – 0.999 |

The measured vendor one-year step is **0.841** (`derivedYearModel`, per-cell range 0.71–0.90). Every cell sits at ~1.00.

One scrape earlier — `dynasty_export_20260816_225946.zip`, the base C1-U5 is stacked on — the same cells derived correctly: provenance `derived_year_step`, ratios **0.704–0.922**, **0 violations in 18 cells**. That is what C1-U6 shipped.

## When

`2029 Early 1st` is absent from the raw payload in all 12 archives before **2026-08-17 02:11 UTC**, and byte-identical to its 2028 twin in every archive since — for **both** pick families, with `pickAnchorsRaw` showing 12 of 12 new anchors equal to their 2028 twin and 0 differing. Both runs either side of the transition have identical scrape health, so this is not a degraded fetch. Two vendors publishing byte-exact clones of their own 2028 board on the same scrape is not a market event.

## Mechanism — two defects in series

**Root.** `_build_site_pick_map` is always asked for `current + 3` (`Dynasty Scraper.py:4934`). When a source has no rows for the requested year, `lookup_tier`/`lookup_slot` substitute the **nearest year's** value (`:4853-4862`, `:4896-4906`) and emit it under the requested year's name — unstepped, unmarked, indistinguishable downstream from an observation. This is the sibling of MISSING IS NEVER ZERO: *missing is not the nearest neighbour's value either.*

**Exposure.** C1-U6 replaced the scraper's hard-coded `_future_pick_years = (2027, 2028)` with a value derived from those same anchors (`:6164`, `:6196`), so the fabrication reads itself back as evidence that a vendor published 2029. The literal had been masking it; making it self-rolling let it roll onto a fabricated year.

**Why nothing caught it.** Each stage behaves correctly given what it was told: the far-future injector only derives years *no vendor publishes*, so it declines; vendor-priced years take no discount by design (T-3/C-2); the discount pass has been stamp-only since C1-U6. Net: a three-years-out pick publishes at its two-years-out price.

This is the **RED-4** class C1-U6 made unrepresentable, re-entering through a door C1-U6 never touched — the retired clone path is still gone, this is a different origin for the same wrong number.

## Blast radius

Every 2029 pick in canonical `rankDerivedValue` — trade calculator, suggestions, arbitrage finder, angles, simulator, draft capital. **Over-valued**: a 2029 first prices ~19% above the measured step (5034 vs 3593). `canonicalSiteValues` too, so market-anchor consumers read the undiscounted anchor. IDP players shift ≲0.1% through the documented IDPTC backbone coupling.

## Recommended repair

Fix at the root: drop the nearest-year substitution so a source contributes nothing for a year it did not publish. `_inject_far_future_pick_sources` then resumes owning far-future derivation — the design that measured 0 violations. Fixing the exposure alone leaves fabricated rows in `pickAnchors` as unmarked evidence.

Named as wrong in the document, with reasons: reclassifying/loosening the test, special-casing 2029, and clamping at the blend (which reinstates audit V-12/C-11).

Verification needs a **full scrape** — the anchors are baked into the archive, so this cannot be measured by rebuilding an existing bundle.

## Effect on the open stack

This is why #875 and #876 are red. Their own diffs are clean; they inherit `main`'s board.

---

### DIFF
--- a/docs/picks/C1_U6_D1_FABRICATED_FUTURE_YEAR_ANCHORS.md
+++ b/docs/picks/C1_U6_D1_FABRICATED_FUTURE_YEAR_ANCHORS.md
@@ -154,6 +154,40 @@ same wrong number.
 
 ---
 
+## 3.4 Four more safety rails are already firing — in the advisory lane
+
+Added after the first CI run on this branch (run 31992373267). The hard gate fails on the
+year-step test above; the **`livedata` lane fails four more times, all on the same nine
+rows**, and it supplies a fact the archive analysis could not:
+
+```
+{'canonicalName': '2029 Early 1st', 'position': 'PICK', 'rank': 70,
+ 'matchedSources': ['idpTradeCalc'],
+ 'reason': 'matching_failure_other_sources_eligible'}
+```
+
+- `tests/api/test_launch_readiness.py::test_no_unexplained_1src_top400` — 9 rows
+- `tests/api/test_picks_end_to_end.py::test_no_unexplained_single_source_with_picks`
+- `tests/api/test_player_identity_regression.py::test_no_unannotated_single_source_in_top_board`
+  — ranks **#70, #88, #98, #167, #185**
+- `tests/api/test_source_monitoring.py::test_launch_readiness_gates_still_pass` (aggregate)
+
+Two things this adds:
+
+1. **Only ONE family votes on these rows.** `ktc` is present in `canonicalSiteValues` but is
+   the scraper's own model composite, retired from voting in 2026-04-28 — so every 2029 row
+   is a single-source row carrying the 30% single-source haircut, at ranks inside the top
+   200.
+2. **The pipeline itself classifies them as `matching_failure_other_sources_eligible`** —
+   it believes other sources should have covered these rows and did not. That is exactly
+   what a fabricated year looks like from the inside: the anchors assert 2029 exists, so
+   every family becomes "eligible", while only the fabrication reaches the blend.
+
+The board's own instrumentation caught this. It did not block because these four tests are
+`livedata` (advisory), which is the correct classification for them — they *are* statements
+about source coverage. The year-step test is the one that belongs in the hard gate, and it
+is the one that fired there.
+
 ## 4. Blast radius
 
 - **Every 2029 pick on the live board**, in the canonical `rankDerivedValue` — so
--- a/docs/picks/C1_U6_D1_FABRICATED_FUTURE_YEAR_ANCHORS.md
+++ b/docs/picks/C1_U6_D1_FABRICATED_FUTURE_YEAR_ANCHORS.md
@@ -1,0 +1,213 @@
+# C1-U6 D1 — the board is pricing 2029 picks at their 2028 price
+
+**Status:** OPEN — live defect on `main`, present in every board built since
+**2026-08-17 02:11 UTC**
+**Class:** canonical value defect (pick valuation) + fabricated market evidence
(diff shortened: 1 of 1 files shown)


jasonleetucker-code:
**CI on this branch is red, and it is the defect this PR documents.** This is a docs-only change; the failure is inherited from `main`.

```
FAILED tests/api/test_pick_completeness_red.py::TestRed3UncalibratedCloneDiscountRetired::test_derived_2029_value_uses_measured_year_step
AssertionError: 0.9839718530101642 not less than 0.95 — 2029/2028 ratio 0.984
```

The run also supplied evidence the archive analysis could not, so I've added §3.4. **Four more safety rails are firing in the `livedata` (advisory) lane, all on the same nine rows:**

- `test_no_unexplained_1src_top400` — 9 rows
- `test_no_unexplained_single_source_with_picks`
- `test_no_unannotated_single_source_in_top_board` — ranks **#70, #88, #98, #167, #185**
- `test_launch_readiness_gates_still_pass` (aggregate)

with this payload:

```
{'canonicalName': '2029 Early 1st', 'position': 'PICK', 'rank': 70,
 'matchedSources': ['idpTradeCalc'],
 'reason': 'matching_failure_other_sources_eligible'}
```

Two facts that adds:

1. **Only one family actually votes on these rows.** `ktc` is present in `canonicalSiteValues` but is the scraper's own model composite, retired from voting in April — so every 2029 row is a single-source row taking the 30% haircut, at ranks inside the top 200.
2. **The pipeline classifies them `matching_failure_other_sources_eligible`** — it believes other sources should have covered them and did not. That is precisely what a fabricated year looks like from the inside: the anchors assert 2029 exists so every family becomes "eligible", while only the fabrication reaches the blend.

The board's own instrumentation caught this. It did not block because those four are correctly classified `livedata` — they *are* statements about source coverage. The year-step test is the one that belongs in the hard gate, and it is the one that fired there.

Also of note in the same run: `idpTradeCalc (picks) raw CSV byte-identical for 33 days (budget 14d)` — one of only two families that price picks has published nothing new for a month, which is why its 2029 clone tracks its 2028 value exactly.

**No fix is pushed here on purpose.** The repair is at the scraper's anchor builder and can only be verified by a full scrape — the fabricated anchors are baked into every archive since 2026-08-17 02:11 UTC, so it cannot be measured by rebuilding an existing bundle. Recommended repair and the three wrong answers to avoid are in the document.

---

PR #893, opened 2026-08-18
Audit batch H: the board diagnostic blamed the wrong subsystem, every time

jasonleetucker-code:
`tests/e2e/helpers/journey.js` exists to explain why an empty rankings board is empty — the
suite's most-repeated and least-informative failure. It was **structurally incapable of giving
the right answer**, and gave a confident wrong one instead:

```
=> PAYLOAD HAS ROWS BUT NO RANK STAMPS: buildRows fail-fasts and returns [] by design.
   The scrape pipeline is not stamping canonicalConsensusRank — investigate upstream,
   do NOT add a client-side blend.
```

## The bug

It fetches `/api/data?view=app`, then counts stamps **only over `body.playersArray`**, using the
un-prefixed `canonicalConsensusRank`.

`view=app` is the runtime view and **deliberately drops `playersArray`** — `server.py` does
`runtime_payload.pop("playersArray", None)`, the documented payload-size optimisation. So that
array is always empty here, `stamped` was always `0`, the "NO RANK STAMPS" branch fired on
**every** board failure, and the "looks serveable, that points at the client" branch was
unreachable.

Everything the diagnostic can actually see lives in the legacy `players` dict, whose rows carry
**`_canonicalConsensusRank`** — with a leading underscore, and it is the key `buildRows` itself
reads on that path (`dynasty-data.js` legacy materializer). The un-prefixed form belongs to the
`playersArray` materializer.

Measured on the real payload it maligned:

| key | populated |
|---|---|
| `_canonicalConsensusRank` | **740 / 1109** |
| `rankDerivedValue` | **849 / 1109** |

The board data was there the whole time.

## Proof

Replaying both expressions against that captured payload:

```
OLD count (arr-only, un-prefixed key): 0   -> fires the FALSE "scrape pipeline" branch
NEW count (key each materializer reads): 740 -> correctly reports the payload serveable
```

## The fix

Count with the same key each materializer reads, and report **both halves** —
`rankStamps=740 (playersArray:0 via canonicalConsensusRank, legacy:740 via
_canonicalConsensusRank)` — so a future mismatch is visible rather than collapsed into one
number.

## Why this is worth its own repair

This message has headlined a week of red E2E runs. It names a subsystem that is fine, and it
rules out the class of fix that isn't needed. **It cost this audit a full investigation cycle** —
I stood up a backend and probed the serving path chasing a scrape-pipeline defect that does not
exist, and provisionally recorded it as a finding before measuring the key names.

The board failure it accompanies is a **stack-readiness race**, not a contract defect: the same
error context carries four 404s, a **502** and a 503, none of which the diagnostic mentions. The
test passes in isolation and failed twice under parallel load.

---


jasonleetucker-code:
Merged with the blocking gate (**Validate PR**) green and the E2E job red. Recording the classification, since merging on a red check should never be silent.

**The E2E failure is not caused by this PR.** It changes one file — `tests/e2e/helpers/journey.js`, the diagnostic itself. Measured on run `32120428479` after rebasing onto batch G (#891):

```
1 failed   journey-settings-overrides.spec.js:45 › toggling a source fires the overrides request
1 flaky    journey-rankings.spec.js:28        › board loads with real rows and clean values
152 passed (was 148), 52 skipped
```

My first hypothesis — that the red was purely a stale base — was **only partly right**. The rebase did take it from 5 failures to 1 + 1 flaky, but it did not clear it. The survivors are now tracked separately in the audit register:

- **F-3a** — `journey-settings-overrides` is a **hard** failure exercising `POST /api/rankings/overrides`, a canonical serving path. Carried as an open audit item rather than absorbed into F-3's "flaky readiness race" framing, which would have understated it.
- **F-3b** — the `journey-rankings` board-readiness race.

The run exits non-zero partly because the repo sets `failOnFlakyTests`, so a retried green is deliberately not accepted. That is correct policy, and the banner this PR repairs is what makes the distinction legible in the first place — which is the argument for landing it rather than holding it behind a failure it was written to diagnose.

Both items remain open in `docs/audit/POST_MERGE_C_SERIES_AUDIT_2026-08-18.md`; neither is closed by this merge.

---

PR #897, opened 2026-08-18
Repair KTC crowd data: one identity owner, retire the dead paths

jasonleetucker-code:
Audit repair. Separate from DraftSharks #894. C2-U2 and forward C-Series work stay frozen.

`KTC_TradeDB` and `KTC_WaiverDB` have reported `partial / valueCount: 0` on every run for months. The objective was not to make two functions return non-empty arrays — it was to leave **one** correct KTC crowd architecture behind. The census says the answer is mostly **subtraction**, plus one repair with a real production consumer.

## The defect

`Dynasty Scraper.py` bound `content` in exactly one place — inside Strategy 3, itself guarded by `if not name_map:`. The identity-map block then tested `if "content" in dir() and content:` — a runtime probe for *"did Strategy 3 run?"*. Its own comment said **"Always build"**.

It could only build when the primary scrape had **failed**. So on every healthy run the crowd databases were skipped with `skipReason: missing_ktc_id_map`, `KTC_CROWD_DATA` stayed empty, and the export gate omitted `ktcCrowd` entirely.

Corroborated independently: **`ktcCrowd` appears in 0 of 173 committed export archives** (decompressed, not a raw grep on the zips). It has never once shipped.

Behind it, `_resolve_ktc_player` returned a **truthy** `f"Player#{pid}"` on a miss — a value that passed every downstream emptiness check and then joined nothing. A missing identity wearing the costume of a present one.

## One owner

New `src/sources/ktc_identity.py`. Both crowd pages carry three inline arrays, and the feeds reference the **search index**, not the value board:

| identity source | entries | waiver refs unresolved |
|---|---|---|
| `playersArray` | 500 | 47 |
| **`allPlayerSearchValues`** | **1,997** | **0** |

`playersArray` is not a smaller-but-correct map — it is the *wrong population*, and every reference it missed was a real claim silently dropped.

The owner prefers the search index, falls back to the value board and **records which it used**, fails closed on id collisions, and classifies every reference as `player` / `pick` / `faab_amount` / `unresolved`. Picks are **classified, never player-resolved** — they arrive both as ids whose row is vendor-flagged `position: RDP` and as literal labels in three observed shapes. **No `Player#<id>` is constructed anywhere**; an unresolved vendor id stays unresolved, with a reason.

Measured on identical live HTML, the live producer (`scripts/fetch_crowd_faab.py`, the 3-hourly `dynasty-crowd-faab` timer) now keeps **200/200 claims (was 150/200)** and names **124/200 drops (was 94/200)**. Across both feeds every reference is named — trades: 508 players + 299 picks + 1 FAAB amount + **0 unresolved**; waivers: **0 genuine misses**, the 76 residuals being the `-1` no-drop sentinel, named as such.

## Retired rather than repaired

- **WaiverDB — SUPERSEDED.** `fetch_crowd_faab.py` is the live owner. The scraper path was a *second* parser of the identical inline array feeding a *second* recommender input, whose only output was a display factor row the live engine already emits — with a claim count, so the survivor is strictly better.
- **TradeDB — FUTURE, NOT PRODUCTION.** Zero consumers: the trades list reached only `len()` for a status count. Deferred to C4-U3 / `C4-MTL-02`; its measured live shape is recorded on the manifest row so that unit need not rediscover it.
- **`ktcIdMap`** — also never shipped (0 of 173) and had no consumer in any language. Removing it *and* the crowd path leaves `KTC_ID_TO_NAME` with no readers at all, so the global is **deleted** rather than repaired for nobody. This is the one place the work diverged from the approved plan, which had said to populate it unconditionally.

`TOLERABLE_PARTIAL_SOURCES` had exactly these two members and is now **empty**. Being allowlisted is precisely why a permanent failure never once had to be looked at; the mechanism stays, with a test pinning that it forgives nothing today.

## Named, and deliberately not repaired here

`scrape_ktc` Strategy 2 still derives its own id→name map to join KTC's value-history API. Same concept, and the owner covers ~4× as many players — but that map feeds KTC's **ranking** values, so migrating it would move the canonical board, which this change promises not to do. Pinned exactly by `KNOWN_DEFERRED_DERIVATIONS`, so neither adding a new derivation nor removing this one can happen silently. It needs its own measured unit.

## Verification

- 22 owner/regression tests + 3 seam integration tests over sanitized fixtures (real structure, synthetic identities).
- **Seven-mutation proof** — restoring the `dir()` probe, the failure-only binding, the `Player#` fabrication, the wrong-population preference, collision-picks-a-winner, unresolved-gets-a-name, and a re-derived producer each fail a named test. Clean restore green.
- **Board inertness on identical input: 0 values moved · 0 ranks changed · 0 newly priced/unpriced · 0 confidence buckets changed.**
- Live seam proof via `fetch_crowd_faab.py --dry-run`: 200 claims fetched, 95/200 comparable for `dynasty_main`, 80/200 for `dynasty_new`, zero identity exclusions.
- ruff · coercion gate · `check_planning_integrity.py` all clean. 673 lines deleted from the scraper.

Two guards were themselves corrected after they misfired: the owner-uniqueness guard first flagged three legitimate *value* readers of `playersArray` (a different concept with its own owner), and a docstring-stripping helper silently stopped matching and reported a clean tree that was not clean.

https://claude.ai/code/session_01Kve8hqgo4v4PeagCHi92ds

---


jasonleetucker-code:
CI red on **Validate PR**, and it is failing on the base branch too — I measured it rather than assumed it.

The unit tests passed (`299 passed, 25 skipped, 396 subtests`). The job failed at the advisory **Source-health check**, on two errors that have nothing to do with this diff:

```
##[error]draftSharks    last fetched 2026-08-05T18:13:02+00:00 (301.8h > 24.0h)
##[error]draftSharksIdp last fetched 2026-08-05T18:13:02+00:00 (301.8h > 24.0h)
```

Run on a pristine `origin/main` worktree, the same script exits **1** with the identical two errors. That is the DraftSharks ingestion staleness #894 exists to fix, reading a committed export payload this branch does not touch (`git diff origin/main -- data/ exports/ CSVs/ config/` is empty).

The only source-health line this PR changes is the label on two warnings:

```
- [source-health][warn] partial_run_tolerable:KTC_TradeDB
+ [source-health][warn] partial_run_unknown:KTC_TradeDB
```

That is the intended effect of emptying `TOLERABLE_PARTIAL_SOURCES`, and both stay warnings — `contract source-health errors: 0` on both branches. They still appear at all only because the committed payload was produced by the pre-change scraper; the next scrape drops them from `sourceRunSummary` entirely.

Two things did change on the pushed head, both mine:

1. **A second `MISSING IS NEVER ZERO` defect on the same live path**, found while writing the format-filtering test the plan called for. The producer derived a crowd league's format with `(settings.get("qBs") or 0) >= 2`, so an absent key became the positive claim *"this league is 1QB"*. Measured against the live feed: renaming `qBs` takes the comparable count from **95/200 to 0/200**, and the run reports that as a healthy empty market rather than a broken parse. Format is now `bool | None`; an unstated format cannot be proven comparable and is excluded, and a feed where no row is classifiable raises. `tep: 0` stays a real statement, distinct from an absent `tep`. Live behaviour is unchanged (95/200 before and after).

2. **Reverted two data files I committed by accident.** `data/scrape_state/sleeper_last_success` and `data/sleeper_last_good.json` were refreshed as a side effect of local import smoke-tests and swept in by `git add -A` — class-C build/test-consumed data, unrelated to this change. Gone from the head.

I will re-check when the run on `e3790eb7d` lands.

---

jasonleetucker-code:
**Correction to my previous comment.** I said the failure was purely the pre-existing draftSharks source-health staleness. That was wrong, and I should not have concluded it from a 45-line tail of the log.

The actual blocking failure was **`ruff format --check .`**, several steps earlier, on two files this PR edits:

```
Would reformat: Dynasty Scraper.py
Would reformat: tests/api/test_source_monitoring.py
2 files would be reformatted, 1142 files already formatted
```

That is mine. I ran `ruff format --check` and `ruff check` against a **subset of paths** rather than the repo — the same class of mistake as running the coercion gate without `GITHUB_BASE_REF`, and neither subset happened to include `Dynasty Scraper.py`. The source-health step is `continue-on-error: true` and never fails the job; I mistook a red annotation near the end of the tail for the cause.

Fixed on `e203644d1`, and both changes are confined to regions this PR already touched — no collateral reformatting:

- two line-collapses that only became short *because* of my deletions;
- `import ast` removed from `Dynasty Scraper.py` — it became unused when `_parse_ktc_literal` (its only caller, via `ast.literal_eval`) was deleted with the crowd path. `ruff check` had flagged it and my subset run never saw it.

Now verified **repo-wide** rather than by subset: `ruff format --check .` → `1144 files already formatted`; `ruff check .` → `All checks passed!`; 633 tests pass across `tests/sources`, `tests/trade`, `tests/api/test_source_monitoring.py`, `tests/api/test_contract_health_lanes.py`, `tests/utils/test_name_clean.py`; the scraper still imports and `scrape_ktc` is intact.

What I said about draftSharks still holds and is unchanged — `##[error]draftSharks last fetched … (301.9h > 24.0h)` reproduces on a pristine `origin/main` worktree with the same exit code — but it was **not** what was failing this job, and I should have said so only after reading the whole log.

I also caught the two Sleeper state files (`data/scrape_state/sleeper_last_success`, `data/sleeper_last_good.json`) drifting a second time from local import smoke-tests. `git diff origin/main..HEAD -- data/ exports/ CSVs/ config/` is empty, so the commit carries none of it.

---

PR #902, opened 2026-08-18
Source-network census, and F-3 re-measured on CI

jasonleetucker-code:
Documentation only. Directive §12–§18 (the source census) plus a correction to the audit register's F-3 figures.

## The §14 defect is worse than "a stale surface"

There are **four** source populations across four surfaces, and **none of them is the real one**:

| surface | population | vocabulary |
|---|---|---|
| `raw.sites` | **2** | source keys (legacy `ktc`, not `ktcSfTep`) |
| `coverageAudit.expectedSites` | **2** — `{offense:['ktc'], idp:['idpTradeCalc']}` | a **hardcoded literal**, not derived |
| `settings.sourceRunSummary.sources` | **14** | **scraper-run names** — `DLF_LocalCSV`, `Flock`, `DraftSharks_IDP` |
| `_RANKING_SOURCES` (canonical) | **21** | source keys |
| **actually voting in the contract** | **21** | source keys |

Two consequences worth separating:

- **Coverage is measured against 2 of 21.** Nineteen sources could stop contributing without moving a coverage number.
- **The run summary speaks a different language from the registry.** `DLF_LocalCSV` covers four registry keys, `Flock` two; `DraftSharks` and `DraftSharks_IDP` are separate run names but one provider family. Nothing maps between the vocabularies, so a per-source disposition cannot round-trip.

## Independence is correct — recorded so it stays that way

**21 sources are 13 families.** DraftSharks offense+IDP is one family; the four DLF boards are one; `fantasyNavigatorSf` is correctly grouped under `ktc` as KTC-derived. No repair needed — recorded because the count is easy to get wrong by reading the source list instead of the family list.

## Dispositions

All 21 registry sources are `ACTIVE — HEALTHY` except the two DraftSharks boards (`ACTIVE — DEGRADED`, 303h). Worth noting: `draftSharksRos` is **1.8h** — the ROS fetch works and only the *dynasty* board is broken, which is the evidence #894's diagnosis rests on.

Non-registry paths dispositioned explicitly, including the KTC retirements from #897.

## A new finding

`footballGuys` / `footballGuysSf` / `footballGuysIdp` carry `_last_success` stamps **2049h (85 days)** old, with **no registry entry, no CSV and no consumer**.

The stamps are inert — `check_source_health` derives its population from the registry (measured: 20 fresh + 2 stale = 21 registry + sleeper), so FootballGuys is not counted and raises no false alarm. What is **not** inert: `data_contract.py` still explains certain players as *"only ranked by FootballGuys IDP"*, so those coverage-gap explanations name a source that no longer exists.

## Freshness stays an owner decision

No approved staleness rule exists in the authority hierarchy, and §15 forbids inventing decay or weighting during an audit. Recorded as **OWNER DECISION REQUIRED** rather than answered.

## F-3 re-measured

The register's E2E figures came from a **local** run on a stale base. Re-run on CI after batch G merged and #893 was rebased onto it:

```
1 failed   journey-settings-overrides.spec.js:45
1 flaky    journey-rankings.spec.js:28
152 passed (was 148), 52 skipped
```

Two corrections: "three flaky under parallel load" becomes **one** at a current base; and **`journey-settings-overrides` is a hard failure, not flakiness** — it exercises `POST /api/rankings/overrides`, a canonical serving path, so it is now carried as open item **F-3a** rather than absorbed into the readiness-race framing (**F-3b**). The non-zero exit is partly `failOnFlakyTests`, which is deliberate policy, not a defect.

https://claude.ai/code/session_01Kve8hqgo4v4PeagCHi92ds

---

### DIFF
--- a/docs/audit/POST_MERGE_C_SERIES_AUDIT_2026-08-18.md
+++ b/docs/audit/POST_MERGE_C_SERIES_AUDIT_2026-08-18.md
@@ -171,6 +171,32 @@ next question.
 treat the readiness race as its own unit. A signal nobody reads is not a signal — and one that
 lies when read is worse.
 
+**RE-MEASURED 2026-08-18, on CI rather than locally.** Batch G merged and #893 was rebased onto
+it, so the suite was re-run on a current base. Measured on run `32120428479`:
+
+```
+1 failed   journey-settings-overrides.spec.js:45 › toggling a source fires the overrides request
+1 flaky    journey-rankings.spec.js:28        › board loads with real rows and clean values
+152 passed (was 148), 52 skipped, 5.2m
+```
+
+So the count improved from **5 failures to 1 failure + 1 flaky**, which is batch G's two repairs
+landing plus the rebase. Two things this corrects in the entry above:
+
+* the earlier "three flaky under parallel load" figure was taken from a **local** run on a stale
+  base; on CI at a current base it is **one** flaky (`journey-rankings`, the board-readiness
+  race) — the same class, a smaller population;
+* **`journey-settings-overrides` is a hard failure, not flakiness.** It exercises the rankings
+  **override** path (`POST /api/rankings/overrides`), which is a canonical serving path, so it is
+  carried as an open audit item rather than absorbed into F-3's "readiness race" framing.
+
+The run exits non-zero partly *because* the repo sets `failOnFlakyTests` — a retried green is
+deliberately not accepted, and the banner explains why (without it, `e2e.yml`'s close step would
+drain every open `e2e-failures` issue). That is correct policy and is not itself a defect.
+
+**Open:** `F-3a` — diagnose `journey-settings-overrides`; **`F-3b`** — the `journey-rankings`
+board-readiness race. Neither is caused by #893, which changes only the diagnostic helper.
+
 ### F-4 · Production proof outstanding for five units · BLOCKED-EXTERNAL
 
 `C0-R`, `C1-U5`, `C1-U8`, `C1-U9` and `C2-U1` are all `CLOSED-PENDING-PROD`. Their code is in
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
**Correction pushed — the PR description above overstates the `expectedSites` finding.**

I wrote that *"coverage is measured against 2 of 21 — nineteen sources could vanish without moving a coverage number."* That frames `coverageAudit.expectedSites` as broken. It is not.

`expectedSites` is `TOP_OFF_EXPECTED_SITE_KEYS` / `TOP_IDP_EXPECTED_SITE_KEYS` = `("ktc",)` / `("idpTradeCalc",)` — the two **backbone anchors**. Its consumer, `server.py::_missing_expected_sites` (audit O-3), asks *"did we lose an anchor?"*, not *"are all sources healthy?"*, and its docstring is explicit that having the payload declare its own anchors is what lets the guard avoid hardcoding a source name. **Two is the correct population for that surface.**

I caught this while starting the S-1 repair — I went to make `expectedSites` derived and found it was already right for its purpose.

**What survives, restated precisely:**

1. **No *published* surface answers "how are the 21 doing?"** `check_source_health` does know — measured 20 fresh + 2 stale = 21 registry + `sleeper` — but it is a CI script writing annotations, not `/api/status`, `/tools/source-health` or the contract. So a reader of the product's own health surfaces cannot see the production population. That is an **absence**, not a wrong number.
2. **Two vocabularies, no mapping.** `sourceRunSummary` names scraper runs (`DLF_LocalCSV`, `Flock`); the registry names source keys. One run name covers up to four keys, so a run-level "complete" cannot be decomposed into which boards actually arrived.
3. **The `ktc` / `ktcSfTep` spelling split** across `raw.sites`, `expectedSites` and `canonicalSiteValues`.

S-1 is restated as *publish a 21-source surface*, with an explicit **do not widen `expectedSites`** so the next reader doesn't "fix" a working detector.

Everything else in this PR is unchanged: the 13-family independence measurement, the dispositions, the FootballGuys finding, and the F-3 re-measure.

---

Issue #804, opened 2026-08-11
Correlated-source / shared-lineage anomaly protection (uncovered class, pre-existing)


jasonleetucker-code:
## The gap

Measured during the corridor-removal pass (#799): when **several sources are wrong in the same direction at once**, the blend moves by up to **~48%** and *neither* the retired market corridor *nor* the blend-integrity hull invariant detects it.

This is a **pre-existing** gap. It is not created by removing the corridor — the corridor did not catch it either.

## Why neither mechanism catches it

Sources agreeing on something wrong is **indistinguishable from agreement** at the blend:

- The **hull invariant** asks whether the blended value left the range of its own contributions. If every contribution moves together, the blend moves with them and stays comfortably inside the hull. Nothing structurally impossible has happened — the inputs were simply wrong.
- The **retired corridor** anchored on `idpTradeCalc`, which is itself one of the correlated voters. When the market moves together, the anchor moves too, so the band never opens.

And there is no independent arbiter available: **every IDP-covering source votes** in the blend. There is no held-out IDP reference to check the consensus against.

## Measured evidence

From the upstream-defense audit (anomalies injected at the **source CSVs**, whole pipeline rebuilt — `docs/master-site-audit/evidence/W02/cd_upstream_defense.py`, results in `cd_upstream_defense.json`):

| injected failure | Hampel | blend moved | old corridor | hull |
|---|---|---|---|---|
| one source ×5 | absorbed | ≤1.7% | 0/6 | 0/6 |
| one source ×20 | dropped on 3/6 | ≤1.7% | 0/6 | 0/6 |
| anchor ×5 | caught by declared-range check | 0.0% | 0/6 | 0/6 |
| **correlated ×3** | **partial (2/6)** | **median 5.7%, max 48%** | **0/6** | **0/6** |
| **correlated ×4** | **partial (3/6)** | **median 5.7%, max 48%** | **0/6** | **0/6** |

Single-source anomalies are well covered by existing upstream defenses. Correlated ones are not.

## What a future investigation should consider

Evidence-driven only:

- **Source independence** — which sources are genuinely independent observations vs. re-publications of one another. Several ranking sources plausibly derive from shared upstream data or shared methodology; that is a measurable question, not an assumption.
- **Shared lineage / methodology families** — grouping sources that fail together, and treating a family as closer to one vote than to N.
[...]

PR #911, opened 2026-08-18
Crowd FAAB: exclude incomparable leagues, refuse stale and offense-only evidence

jasonleetucker-code:
**Lane 4 — Market / FAAB / Sharp / Analyst.** First unit: the *external* FAAB market layer.

> **Branch note for the integration lane:** the lane brief names `v1/market-analyst-intelligence`, but this session is pinned by the harness to `claude/market-faab-analyst-v1-bbieie` and instructed never to push elsewhere. Same content; rename on integration if the branch name matters.

---

## Methodology

Owner spec: [`docs/FAAB_MARKET_SIGNAL_NORMALIZATION_2026-08-14.md`](../blob/main/docs/FAAB_MARKET_SIGNAL_NORMALIZATION_2026-08-14.md) §3, §5, §7, §10 — approved, partly implemented.

**Own-league history was already correct and is untouched.** `faab_history.fetch_bid_history` normalizes to percent-of-original-budget across the $1,000 / $200 / $100 eras, keeps $0 bids, and refuses a season with no usable `waiver_budget` rather than fabricating $100.

The gap was the **external** lane: the KTC crowd waiver feed → `data/faab/crowd_history_<leagueKey>.json` → `crowd_bid_index` → `faab_engine.rival_bid_cdf`, where it is blended into the modelled rival share at **weight 0.6**. What that pool admits moves real recommended bids.

Full record: **`docs/faab-external-market-comparability.md`** (new).

## What was measured

Live feed, 2026-08-18, 200 rows / 86 leagues, gated to `dynasty_main` (12 teams, superflex, TEP, 2 TE starters, IDP):

| population | n | leagues | median bidPct | zero-bid share |
|---|---|---|---|---|
| the previous gate (superflex + TEP-as-bool + teams ±2) | 106 | 50 | 0.65% | 29% |
| …of which `rostersPerPlayer > 1` (multi-copy) | **39** | 12 | **0.20%** | 41% |
| …of which total budget < $10 | 7 | 2 | 0.00% | **100%** |
| single-copy only | 67 | 38 | **1.00%** | 22% |

1. **Multi-copy leagues were 37% of the admitted sample.** MyFantasyLeague's `rostersPerPlayer` (2/3/4) lets the same player sit on several rosters at once — no waiver scarcity, so claims clear near $0, a 5× drag at the median. The field was parsed by nobody.
2. **$1-total-budget leagues** contributed 7 rows, all 0.00%: a league that cannot express a price, only "claimed / did not".
3. **`tep` is a 0/1/2/3 level** (Off / TE+ / TE++ / TE+++) flattened to a bool, and **`is2TE`** is a separate flag the spec names explicitly — both unread.
4. **The feed contains no IDP league.** Across all 200 rows the starting slots are exactly `QB/RB/WR/TE/PK/Def`; `Def` is a team D/ST and **zero** rows carry a DL/LB/DB slot.
5. **`dynastyPlatformType` is `1` on every row** — the platform (MFL), not per-league dynasty proof.
6. **Freshness was never read** — a dead timer would silently price today's market off a months-old ledger.

After the change the same feed yields 66–67 comparable rows for `dynasty_main` (A=16, B=41, C=10) and 58 for `dynasty_new`, with the exclusion census printed per run.

## Source-family assumptions

- **Own-league Sleeper history** and the **cross-league crowd feed** stay separate populations — one league's culture vs. many leagues' prices. Unchanged.
- **Dynasty is a source-level claim**, not per-league verification (§8 of the doc). Stamped as `source_level_claim:ktc_dynasty_waiver_database` so a consumer sees which kind of evidence it holds.
- The crowd population is **offense-only**, derived from what the retained rows contain rather than hardcoded — it self-corrects if KTC ever carries an IDP league.

## Normalization

`normalized_bid_share(bid, budget)` returns **`None`** — never `0.0`, never an assumed $100 denominator — when the budget is missing, unparseable or ≤ 0. The budget is the *denominator*; fabricating one is wrong by up to 10×, not slightly off. A **$0 bid is a real observation** and returns `0.0`. `equivalent_on_budget()` expresses it on the target league's scale ($40 of $200 → $20 of $100); `None` propagates.

## Freshness rules

| state | meaning |
|---|---|
| `fresh` | updated inside `maxFileAgeDays` (7) — the only state that authorises pricing today's market |
| `stale` | not refreshed inside the budget, **or has no `updatedAt` at all** — unmeasurable freshness is not freshness |
| `missing` | no ledger, or nothing survived classification |

Stale evidence is **retained and readable**; only its authority expires — the same posture `league_registry.scoring_evidence_state` takes. The crowd timer runs 3-hourly, so a 7-day budget is 56 missed runs: it detects a dead timer without flapping.

## Recommendation semantics

**Nothing here can touch canonical value.** The objective FAAB ceiling is computed before any external observation is read and is structurally unable to move; this is the market layer only, pricing how *contested* a claim will be. Re-pinned through the new accessor by `TestTheInvariantSurvivesTheRewire`.

**Tiers are reported, never weighted.** The owner spec is explicit that tier weights are evidence-gated and no outcome data exists to fit them, so a demoted row still votes in full (`test_a_tier_changes_no_number`).

**Refusals are visible.** `/api/faab/recommend` gains `crowdMarket` — state, `asOf`, `ageDays`, tier composition, exclusion census, `pricesIdp`, `dynastyProvenance`, `refusalReason`, `playerHasEvidence` — so "no crowd price for this player" and "we declined to quote one" cannot read the same.

## Tests

`tests/trade/test_faab_comparability.py` (new, 58 cases) — $200→$100 and $1,000→$100 invariance; already-$100 identity; zero bids kept; missing budget → `None`; multi-copy and degenerate-budget exclusion; SF/non-SF (symmetric about the *target*, so a 1QB target excludes superflex evidence); TEP on/off and severity gap; 2-TE tiering; IDP refusal and unlock; renamed-key parse failures reading as unknown rather than a default; stored-shape ≡ raw-shape classification.

`tests/trade/test_faab_crowd.py` — freshness states, legacy-row retention, on-read re-classification, the dollar equivalent, dynasty provenance, the IDP gate, and the headline invariant through the new accessor.

`tests/api/test_faab_recommend_endpoint.py` — the `crowdMarket` block across fresh / stale / no-ledger / IDP-refused / IDP-unlocked, plus on-read exclusion.

`tests/trade/test_faab_config_parity.py` — now scans `faab_comparability.py` too, with a guard test that fails if the new module stops being scanned. A third defaults source (the dataclass fields) is pinned separately.

All deterministic, no network — they belong in the hard `-m "not livedata"` gate per the CI-lanes rule.

```
tests/trade  512 passed
tests/api/test_faab_recommend_endpoint.py + test_faab_analytics.py  22 + 15 passed
tests/news, tests/lineup  255 passed
ruff  clean
```

## Consumer list

| consumer | change |
|---|---|
| `scripts/fetch_crowd_faab.py` | delegates to the owner; stores format evidence + normalized fields; prints the exclusion census; refuses when *every* row is unclassifiable |
| `src/trade/faab_history.py` | `build_crowd_market` / `crowd_evidence_for` / `crowd_refusal_reason`; `crowd_bid_index` kept as a compat wrapper with its contract intact |
| `server.py::/api/faab/recommend` | reads through the fail-closed accessor; stamps `crowdMarket` |
| `frontend/components/waivers/FaabRecommendation.jsx` | **untouched** — the UI lane's call whether to render the new block |

## Known limitations

1. **No IDP population exists** — external clearing-price evidence for defenders is *unavailable*, not zero. The gate reports the refusal; it substitutes nothing.
2. **Tier weights are unvalidated** and change no number. Validating them needs Brisket's realized clearing prices joined to contemporaneous crowd observations (`scripts/faab_backtest.py` is the natural home).
3. **Dynasty is a source-level claim**, not per-league verification.
4. **Legacy rows stay `unverified`** until the 120-day retention window clears them.
5. **TEP severity is carried but dormant for Brisket** — our registry records TEP via the scoring-profile label, not a KTC-style 0–3 level, so the gap check only fires for a target whose level is known.
6. **Sleeper most-added/dropped market heat is not implemented** — spec §4 makes the production transform evidence-gated.

## Cross-lane dependencies

- **Integration lane:** the branch-name note above; nothing else in this PR overlaps the V1 contract sweep.
- **UI lane:** `crowdMarket` is additive and unconsumed; rendering it is optional.
- **Trade lane:** no trade math touched. `faab_engine.recommend`'s signature is unchanged.
- Also included, documentation only: `scratchpad/LANE4_SIGNAL_EMITTER_INVENTORY.md` — the Buy/Sell emitter landscape for a future Central Reconciliation unit, recording that `src/news/unified_signal_engine.py` documents itself as "single entry point for every BUY/SELL/HOLD decision" and is **imported by nothing in production** (only its own test and a prose mention). Consequence: `src/news/usage_signals.py` reaches no user surface. No reconciler is built here — that needs an owner decision.

## Production checks

After the first real run on prod:

```bash
python scripts/fetch_crowd_faab.py --league dynasty_main --dry-run   # exclusion census
```

- rows in `data/faab/crowd_history_<leagueKey>.json` carry a `comparability` block and a `rostersPerPlayer` setting;
- `/api/faab/recommend` reports `crowdMarket.state: "fresh"`.

If it reports `stale`, the `dynasty-crowd-faab` timer has stopped — that is the condition this change makes visible rather than silently absorbing. Existing accumulated rows will report as `unverified` tier until they age out; that is expected, not a regression.

---


jasonleetucker-code:
## Added scope: `351b6dc` — stale trending evidence (Lane 4, unit 2)

The PR body above covers the **crowd** market signal. This commit closes the same class of hole in the **other** market-demand input to the same engine, plus one small owner-stated gap in the sharp windows. Same lane, same domain; flagging it here because it materially extends what the PR body describes.

### The defect

`src/adapters/sleeper_trending.py` serves the **previous** snapshot when a fetch fails — deliberately, and with no absolute cap, so an upstream outage keeps one snapshot in front of every request indefinitely. The adapter is right; the consumer was not. `server.py` stamped the age into `inputsAsOf.trending` but never passed it to the recommender, so `_demand_evidence_factors`:

1. asserted `"N adds in the last 24h"` however old the snapshot was — a claim about a day that may long since have passed;
2. counted as **realised** evidence, so `compute_confidence` reported a dead upstream as present input.

`stale_inputs` flagged the identical input in a sibling field, so the payload contradicted itself.

**So did the test suite, visibly.** The fixture stamped `2026-07-25`, `staleInputs` therefore always contained `trending`, and a separate test asserted the factor row was present. Both are fixed: the fixture now stamps a genuinely current observation, and the stale case is tested explicitly rather than being the accidental default.

### Semantics

| state | row |
|---|---|
| fresh | `"12,000 adds in the last 24h"`, realised |
| stale, or age unmeasurable | `missing`, and it names *when* the observation was made |
| absent | `missing`, `"missing"` |

The confidence drop falls out of the existing mechanism — `compute_confidence` already excludes `missing` rows — so there is no second staleness rule.

**One owner for the ceiling.** `faab_contention` grows `input_age_seconds()` / `input_is_stale()`, `stale_inputs` is re-expressed in terms of them, and the recommender asks that owner. A literal copied into the recommender is exactly how the two came to disagree, so its absence is asserted structurally by an AST test.

**Zero adds stays a real observation** — not-trending is signal. A *malformed* count is a different thing and now becomes a missing input rather than a fabricated zero, which retires a real entry from the decision-path coercion baseline (677 → 676) instead of re-blessing it under changed text.

### Deliberately not implemented

Owner spec §4 asks for a velocity view of Most Added behind a COLD…SURGING label and states the production transform is **evidence-gated**. Two things block it, neither of them effort:

1. no backtest exists to justify a transform, and §4 caps Sleeper heat's net upward effect at ~10% of the pre-heat bid absent one;
2. the persisted series that would supply velocity is `src/retention/evidence_store.py::trending_series`, and that package's docstring forbids decision-path reads outright — *"a value that fed back into the thing it records would make every measurement taken from it circular."*

Moving that boundary is a separate, owned decision. Recorded in `docs/faab-model.md` rather than quietly skipped.

### Sharp windows (owner intent §9)

Sharp Roster Percentage published `sevenDay / thirtyDay / seasonToDate`; the **14-day** baseline was simply absent while the Buy/Sell tracker already spoke `14d`. Added — one `holdings_as_of` read and nothing else, because the population-overlap guard is per-baseline and the new window inherits the protection against a grown cohort reading as ownership gain. Pinned by a test that it sees a move the 7-day window cannot; a window that duplicates a neighbour does not earn its place.

**Cross-lane:** the frontend renders trend columns from an explicit key list (`frontend/app/market/sharp-roster-percentage/page.jsx`), so the backend key is additive and surfacing it is the UI lane's call. No UI touched here.

### Verification

```
tests/ -m "not livedata"   8,204 passed / 51 skipped
ruff check + ruff format --check (pinned ~=0.6.0)   clean
check_decision_coercions / product_plan_governance / planning_integrity / audit_status   clean
import gate   ok
```

**Production check:** on a box where the trending fetch has been failing, `/api/waiver/faab-recommend` should now show the trending factor as `missing` with an observation time, `trending` in `staleInputs`, and a confidence no higher than the same request with fresh data — where before the three disagreed.

---

jasonleetucker-code:
## CI is red on a **base-branch** failure, not this PR — and the cause is measured

`Validate PR` failed on `351b6dc`. The failure is in the pick-completeness census, which this PR touches nothing near. **It reproduces on `origin/main` with none of my commits**, so I am not pushing a speculative fix; this needs the lane that owns pick valuation / audit classification.

```
FAILED tests/api/test_pick_completeness.py::TestPickCompletenessCensus::test_contract_validation_census_is_green
  20 × pick_completeness_census:2029 {Early,Mid,Late} {2nd..6th}:missing_or_unpriced
FAILED tests/api/test_picks_end_to_end.py::TestPicksPresentInContract::test_every_pick_has_rank_and_value   (livedata lane)
  15 × 2029 pick rows with no finite canonical value
```

### Bisected to a single data refresh — same code, three snapshots

I checked out `origin/main` in a scratch worktree and swapped only `exports/latest/` + `CSVs/`:

| main commit | refresh | census |
|---|---|---|
| `68b42ea` | 13:34 UTC | **pass** |
| `8ec1978` | 15:11 UTC | **pass** |
| `f9444f7` | 17:14 UTC | **fail** |

Code identical across all three. This is data-induced, and it predates and is independent of this PR.

### What changed in that refresh

`pickAnchors.idpTradeCalc` collapsed from **rounds 1–4 to round 1 only**, for every year:

| source | `8ec1978` (pass) | `f9444f7` (fail) |
|---|---|---|
| `ktc` | 2026/2027/2028 → rounds 1,2,3,4 | unchanged |
| `idpTradeCalc` | 2026/2027/2028 → rounds 1,2,3,4 | **2026/2027/2028 → round 1** |

2029 is the C1-U6 far-future horizon derived from the nearest published year, so losing 2028 rounds 2–4 on one of the two pick markets takes 2029 rounds 2–6 with it.

### The part I'd flag hardest

**`CSVs/site_raw/idpTradeCalc.csv` was not modified in that refresh, and still contains all 36 tier pick rows in both snapshots.** The vendor input is byte-identical — the same file the source-health step separately reports as unchanged for 34 days — yet the derived anchors dropped three quarters of their cells. So the collapse happened between a complete CSV and the written anchors, not at the source.

That bears on **F-22**, currently classified P2 "reporting inconsistency" on the reasoning that *the contract independently reads the CSV and receives all 36 vendor pick rows, therefore no canonical pick value was shown to be wrong*. The first half still holds. The second half is what this evidence touches: the discarded `idpTradeCalc` anchors turn out to be **load-bearing for the 2029 derivation**, and when they collapse, 15 pick rows lose their canonical value.

I am explicitly **not** claiming F-22 is P0 and I have not touched pick math — the brief is clear that classification and closure belong to the integration lane. This is new measured evidence for that decision, offered rather than acted on.

### Secondary observation, same owner

`test_contract_validation_census_is_green` sits in the **blocking** (`-m "not livedata"`) lane while asserting a completeness status over the *live* board. That is the shape `CLAUDE.md` §"Rule for new tests" warns against — *"a test in the hard gate must not assert an absolute count, a floor or a health status over the LIVE board — those are functions of which sources answered the last scrape."* Today a single upstream pick-board change turned every open PR red, which is the same failure mode the structural/source-health lane split was introduced to prevent. Worth deciding whether this assertion belongs in the source-health lane instead.

### Status of this PR

Green locally on the tree I control:

```
tests/ -m "not livedata"   8,204 passed / 51 skipped
ruff check + format (pinned ~=0.6.0), coercion, governance, planning, audit-status, import gates   clean
```

Blocked only by the base condition above, since CI validates the merge. I'll re-run once `main` recovers and will not push anything speculative in the meantime.

---

jasonleetucker-code:
**Integration gate — lane 4.** Reviewed the diff, the ownership boundary and the shared-file collisions. No canonical-ownership violation: the FAAB engine, crowd path and sharp roster percentage are all lane 4's, and the `server.py` changes stay inside the crowd/trending path. The trending-count treatment is right — "absent from the board is a real zero, a malformed count is a missing input" is the distinction the coercion gate exists to keep visible.

**One change requested before merge, and one thing to know about a shared file.**

**1. Move the two `scratchpad/*.md` files into `docs/`.** `scratchpad/` is not gitignored and has **0 tracked files on `main`** — this PR would introduce it as a tracked directory, and in this environment "scratchpad" is also the name of the session-local temp directory, so the two get confused. These are not scratch: `LANE4_SIGNAL_EMITTER_INVENTORY.md` describes itself as *"the evidence base for a future Central Buy/Sell Reconciliation unit"*, which is a durable evidentiary claim that needs a discoverable home. Suggest `docs/faab/` or `docs/market/` alongside `faab-external-market-comparability.md`.

**2. `config/coercion_baseline.json` is contended by three lanes** (#911, #913, #914) and each recomputed `count` from a different base — yours goes 677→676 while `main` is already at 676 with a different set. Do **not** hand-merge it. On rebase, run `python scripts/check_decision_coercions.py --write-baseline` and commit the result; the gate scopes to changed files, so a stale count from another lane won't wrongly fail you.

**3. CI is red for a reason that is not yours.** The unit-test hard gate is failing repo-wide on audit F-30 (unpriced 2029 picks after a pick-market truncation) — it also skipped a production deploy. The fix is in #910, merging shortly. Rebase after that lands and re-run before I re-verify.

---

jasonleetucker-code:
**Integration order — measured, and you are the second lane that needs a pre-resolution.**

I trial-merged all six branches into a scratch worktree off `main` rather than predicting from file overlaps. `server.py` (four lanes) and `CLAUDE.md` (two lanes) merge **cleanly** — including your FAAB-crowd section against #913's new trade-comparison section. The only conflict anywhere is `config/coercion_baseline.json`, hit by **#911 and #913**.

```
#910 integration/stability  →  #914 canonical roster  →  #913 trade
                            →  #912 contract consumers
   #915 and #911 (you) integrate independently — no shared-owner collision
```

You are not gated behind the roster/trade chain. Rebase onto `main` once #910 lands and re-run; I verify at your exact head.

**Do not hand-resolve the baseline.** Yours recomputes `count` 677 → 676 from a base that had 677, while `main` is already 676 with a *different* set — a hand-merge yields a number matching nobody's tree. Run `python scripts/check_decision_coercions.py --write-baseline` and commit. Verified: with that, the full six-lane set integrates and the baseline lands at **670**.

**Your finding is now registered as `F-33`, and tracked as `V1-129`.** It is recorded as a *defect against an already-required capability* (FAAB core), not as new product scope — the owner's rule permits adding those without redefining the denominator, and the denominator change log says so explicitly. The entry keeps two things visible because they are easy to lose: that **own-league history was already correct and is untouched** (so a future reader does not "fix" the half that works), and that admitting evidence which cannot be compared to the question is not weak evidence but *not evidence* — weighting it at 0.6 was the defect.

Still requested from my earlier review: move the two `scratchpad/*.md` files into `docs/`. `scratchpad/` has 0 tracked files on `main` and is the session-local temp directory by convention here; `LANE4_SIGNAL_EMITTER_INVENTORY.md` describes itself as the evidence base for a future reconciliation unit, which needs a durable home.

---

jasonleetucker-code:
## Synchronization pass complete + V1-57 / V1-60 (`a00749f` → `5030791`)

`origin/main` merged through `52541f5` (20 commits: #909 observability F-19/F-20/F-21, plus automated refreshes). Re-validated at exact head — the previous green run is superseded.

### One silent mis-merge, caught

`config/coercion_baseline.json` **auto-merged cleanly and was wrong.** Main and this branch each retired one entry and each wrote `count: 676`. Git took both entry removals (correct) and the identical count line (incorrect) — after two removals the truth is **675**. Corrected, both removals verified present, and `check_decision_coercions` now agrees: *675 present, 675 accepted as debt*.

Everything else resolved semantically rather than trusted:

| file | resolution |
|---|---|
| `server.py` | this branch's diff vs main is 4 hunks, all inside `post_waiver_faab_recommend`; #909's `_board_age_hours` intact |
| `src/api/ops_alerts.py`, `data_contract.py`, `Dynasty Scraper.py`, `health-check.yml` | **byte-identical to main** — nothing of #909 lost or re-implemented |
| `CLAUDE.md` | one additive hunk (35 lines), no main-side edit to reconcile |
| `config/trade/faab.json` | mine only, additive |

**The 2029 pick-completeness failure I reported earlier has cleared on its own** — main's 19:21 refresh restored `pickAnchors.idpTradeCalc` to rounds 1–4, exactly as diagnosed. The census passes on the merged tree. The transient nature is itself the evidence for the lane-split point in my earlier comment.

### V1-57 — reconciled before implementing

The deployed reality and the requirement only partly overlap, and conflating them would have produced the duplicate scheduler you warned about:

| | script | scope | state |
|---|---|---|---|
| `dynasty-crowd-faab` | `fetch_crowd_faab.py` | what **other** leagues pay | **already deployed**, 3-hourly — not duplicated |
| *(none)* | `fetch_faab_history.py` | what **this** league pays | **no timer, no workflow, no cron** anywhere |

The engine fits its market priors — median/p75/p90, zero-bid share, per-manager aggression — from the *second*, and without the file it falls back to configured priors and says so in `contention.notes`. So the gap has always been reported and never closed. The script's own docstring says "run this on a cadence"; there was none.

Added the service/timer pair + one `install_simple_timer` line. **Daily 07:40 UTC**, clear of the 2-hourly scrape, crowd-faab (`00/3:25`) and the sharp chain. Daily rather than 3-hourly because the asymmetry matters: the crowd feed is a five-day rolling window where a missed run loses observations forever, whereas Sleeper retains transactions indefinitely — so here a missed run costs freshness and no data.

`tests/deploy/test_all_timers_are_wired.py` failed on the templates before the installer line existed. That guardrail working is the point: shipping both templates while nothing installs them is the documented failure that hid crowd-faab, sharp-activity and board-snapshot until 2026-08-05.

### V1-60 — the silent zero was real, and there were three

Every counter on `CollectResult` is `0` both when the lane ran and found nothing and when it never ran; a bare `CollectResult()` serialises identically to a real empty pass — `"0 rosters, 0 errors"` — which reads as a healthy platform with an empty wire. Three paths produced one: `skip_sleeper`, `skip_ffpc`, and `collect_ffpc_rosters` with no `ffpc:` key in the cohort.

Now `status` + `unavailableReason`: `skipped_by_caller`, or `no_cohort_managers_on_platform` — the latter a fact about the **cohort**, not FFPC's health, said rather than left to be inferred from a zero. An empty pass that really ran still reports `ok`; the distinction only works if both directions hold.

### V1-58 / V1-59 — blocked on production access, and the code is already right

`data/ops/sharp-production-smoke.json` has **never** recorded a verified population — the last nine runs are `unverifiable_unauthenticated` ×8 and `deploy_failed` ×1, all `cohort={}`. Locally `cohort_members()` returns 0 with a coverage block naming each pool as 0.

**That is the correct behaviour, not the defect.** The repo already treats `unverifiable_unauthenticated` as *insufficient evidence, not bad news* (`test_sharp_smoke_commit_order.py:94`), and the coverage block explains a zero instead of asserting one — which is the semantic V1-59 asks for. What is missing is **production authentication**, and no change in this lane can prove population without it. A bootstrap that assumed a cohort would defeat the reporting that is working. → **Claude 5, ops dependency.**

### Contract not found — flagging rather than guessing

`VERSION_1_COMPLETION_CONTRACT.md` is **not on `main` or any non-archive branch** visible to this session, and no file carries the `V1-55`…`V1-65` identifiers (nearest owner record: `docs/C_SERIES_REPLAN_AND_COMPLETION_CONTRACT.md`). I worked from the enumerated list as given; the numeric mapping in `scratchpad/LANE4_V1_RECONCILIATION.md` is **inferred from the given order** and should be confirmed before being quoted as contract ids.

### Handed off, not built

- **#804** — analysis preserved. Its own evidence says a defensible correlation grouping needs longitudinal rank digests that do not exist (24 sources live, 3 archived, 0 of 176 bundles carry more). The retention proposal (**+31 KB/bundle** digest vs +155 KB full CSVs) is a perishable-evidence/ops decision → **Claude 5**. No weights or families invented.
- **Central Buy/Sell reconciler** — inventory preserved, not built; C6-SIG-01/02 stay post-V1 unless reclassified.
- **UI** — `crowdMarket`, `status`/`unavailableReason`, and the 14-day sharp window are published as truthful API state; rendering → **Claude 6**.

### Exact-head validation

```
tests/ -m "not livedata"        8,320 passed / 51 skipped
ruff check + format (~=0.6.0)   clean
coercion gate                   675 present, 675 accepted — clean
governance / planning / audit-status / import   clean
release-candidate               "the validated tree IS the tree that will merge"
                                0 class-B, 0 class-C outstanding
```

---

jasonleetucker-code:

Mark it **READY FOR INTEGRATION** when it genuinely is, then **immediately take your next dependency-ready unit**. Your side: implement, stay synchronised with `origin/main`, fix lane-owned defects, run the deterministic/integration/exact-head checks, document limitations and required post-deploy verification honestly.

You integrate **independently** of the roster→trade→frontend chain, so you are not queued behind it.

**Approved on ownership.** `faab_comparability` bounds its own authority correctly, and the line that made me stop worrying is *"a hype cycle can say a claim is expensive. It can never say the player is worth more."* — the objective ceiling is computed before any external observation is read and is structurally unable to move. Deciding comparability in a module rather than in the fetch script, so a tightened policy re-applies to an accumulated ledger instead of requiring months to rebuild it, is the right shape for the same reason the scoring-fingerprint evidence states are.

**One condition before I merge**, and it is the only thing outstanding: the two `scratchpad/*.md` files. That directory has **0 tracked files** on `main` and is the session-local temp convention here, while `LANE4_SIGNAL_EMITTER_INVENTORY.md` describes itself as the evidence base for a future reconciliation unit. Move them under `docs/` or drop them — a durable record in a temp directory is a record that quietly disappears.

Two things I will **not** hold you for, so you can move on: the crowd finding is registered as `F-33` and tracked as `V1-129`, filed as a defect against the already-required FAAB core rather than as new product scope; and `V1-55` stays `VERIFIED` because the engine's identity held — it was the admission policy of an input that was wrong, not the ceiling/recommendation split.

---

jasonleetucker-code:
# ✅ READY FOR INTEGRATION — @Claude 5

**Head:** `5c375b8` · **Branch:** `claude/market-faab-analyst-v1-bbieie` · **I am not merging this.**

## Exact-head evidence (at `5c375b8`)

```
tests/ -m "not livedata"        8,329 passed / 51 skipped / 324 deselected
ruff check + ruff format        clean (pinned ~=0.6.0 — a newer ruff reformats 184 unrelated files)
check_decision_coercions        674 present, 674 accepted — clean
product-plan governance         clean
planning integrity              clean
audit-status drift              no drift
import gate                     ok
release-candidate               "the validated tree IS the tree that will merge"
                                0 class-A / class-B / class-C outstanding
```

**Sync state:** merged `origin/main` twice during this pass. Three commits landed on main *after* the final gate started — `4b18ba5`, `9f3f517` (Sharp production smoke), `1055710` (idpshow refresh) — all ops/data records with **zero file overlap** with this diff, verified by set-intersection. I stopped chasing there deliberately: per `CLAUDE.md`'s HEAD-FREEZE section, blocking on every base change chases the 2-hourly refresh forever, and the strict once-only check is yours at the release moment.

## What's in it

| unit | |
|---|---|
| **Crowd FAAB comparability** | multi-copy leagues (37% of the admitted sample, 5× cheaper median) and degenerate budgets excluded; offense-only population refuses IDP claims; stale ledger refused; A/B/C tiers reported and **weighting nothing** |
| **Stale trending evidence** | no longer claims "last 24h" at any age, no longer holds confidence up; one owner for the staleness ceiling, pinned by an AST test |
| **Analyst claim schema** | owner T-NEW-10 gating artifact; STASH structurally cannot merge into conviction BUY |
| **V1-57** | own-league FAAB bid-history timer (crowd-faab already covers the *other* half — not duplicated) |
| **V1-60** | three silent zeros in the roster lanes → `status` + `unavailableReason` |
| **V1-63** | zero-quality manager was voting at **full** weight; `networkConcentration` published `0.0` for an undefined ratio |
| **V1-65** | assessed as already consolidated; the sharp-vs-insider **vocabulary** boundary now pinned |

## Two things for your judgement

1. **`VERSION_1_COMPLETION_CONTRACT.md` does not exist** on `main` or any non-archive branch, and no file carries the `V1-55`…`V1-65` identifiers. The mapping in `scratchpad/LANE4_V1_RECONCILIATION.md` is **inferred from the instruction's ordering** — please confirm before it is quoted as contract ids.

2. **A clean auto-merge was wrong once.** `config/coercion_baseline.json`: main and I each retired an entry and each wrote `count: 676`; git took both removals and the identical count, when after two removals the truth was 675. Worth knowing if other lanes touch that file — it will not conflict, and it will be wrong.

## Handed to you, not built

- **#804** — the rank-digest retention proposal (**+31 KB/bundle** vs +155 KB for full CSVs) is a perishable-evidence/ops decision. Evidence is perishable in the literal sense: 24 sources live, 3 archived, 0 of 176 bundles carry more, so every day without it is a day of correlation evidence that cannot be reconstructed. No weights or families invented.
- **Central Buy/Sell reconciler** — inventory preserved, not built; C6-SIG-01/02 stay post-V1 unless you reclassify.
- **V1-58 / V1-59** — blocked on **production authentication**, not code. The smoke has never recorded a verified population (`unverifiable_unauthenticated` ×8, `deploy_failed` ×1), which is the *correct* reporting; a bootstrap that assumed a cohort would defeat it.
- **UI** — `crowdMarket`, `status`/`unavailableReason`, the 14-day sharp window are published as truthful API state → Claude 6.

Moving on to the next dependency-ready item now; ping me here if integration turns anything up.

---

jasonleetucker-code:
# READY FOR INTEGRATION — re-evidenced at `51ac3ad`

Integrated the review push (`27ea5fb`), synced main, added V1-56, re-ran everything. **0 behind main.** I am not merging.

## On the review's two findings

**The `scratchpad/` → `docs/lane4/` move: taken, and thank you.** Sensible placement. One correction for the record: the PR has **zero review threads** (`get_review_comments` → `totalCount: 0`), so I don't think the move was requested twice and missed — I think this is its first appearance. Either way it's in, the three files kept their history through the rename, and the `CLAIM_SCHEMA.md` cross-reference now resolves. I verified no stale `scratchpad/LANE4` references remain anywhere.

**The `CLAUDE.md` finding is a misread, and I want to correct it precisely rather than let it stand.** That uncommitted edit was mine, in flight, and it did **not** delete the `faab_comparability` block. It replaced a different, adjacent paragraph — the stale claim that `faab_analytics.py` "gates its median on `bid > 0`". Both are present at `51ac3ad`:

```
$ grep -c 'faab_comparability.py`` (2026-08-18)' CLAUDE.md   → 1   # block intact
$ grep -c 'CORRECTED 2026-08-18' CLAUDE.md                   → 1   # my V1-56 edit
```

The revert landed in the reviewer's own workspace, not mine, so nothing was lost — but "deleted the documentation block this PR added" would have been a serious thing had it been true, and it wasn't. The lesson I'll take is on me: a long-running uncommitted edit is indistinguishable from stray state to anyone else looking, so I'll commit before triggering review.

## V1-56 (new since the last handoff)

The `/waivers` FAAB context strip's headline field is **correct** — `faab_analytics.py` keeps $0 bids and computes mean and median over the full set, pinned by `test_zero_bid_free_agents_are_counted`. **CLAUDE.md said it wasn't**, telling readers to distrust a right number and prefer another module on obsolete grounds. Corrected under CLAUDE.md's own precedence rule (*live code can prove a status claim in any document, including this one, stale*).

It matters because of the size: 41–77% of adds cost nothing per season, so a nonzero-only median reads ~2% of budget where the truth is 0% — on a representative bid set, `0.0` vs `8.5`.

⚠️ **This edits `CLAUDE.md`, a shared governance file** — flagged for coordination, not treated as lane-local.

## Exact-head evidence (`51ac3ad`)

```
tests/ -m "not livedata"     8,329 passed / 51 skipped / 324 deselected
ruff check + format          clean (pinned ~=0.6.0)
coercion gate                674 present, 674 accepted — clean
governance / planning / audit-status / import   clean
release-candidate            "the validated tree IS the tree that will merge"
                             0 class-A / class-B / class-C outstanding
sync                         0 behind origin/main
```

On the merge point: agreed, and no argument from me — I have never merged this and won't. I flagged it ready and moved on to the next queue item, which is the sequencing I was asked to follow. The merge decision is the human's.

---

jasonleetucker-code:
**Three things: #911 is reconciled by me, your baseline invariant is now enforced, and #804 capture is AUTHORIZED. Take the capture unit now.**

### 1 · #911 — I reconciled it onto the new `main`; do not re-do it

`#910` merged (`ff1d9dd`), which put your `51ac3ad` behind. Rather than hand that back, I merged `main` into your branch and pushed `0a5ef8f`. Your CI is re-running on the true merge tree; nothing is needed from you.

**Your invariant earned its keep on that very merge**, and not in the way I expected. The merge was clean *and* the declared count already agreed (674/674) — so by the naive reading there was nothing to fix. Regenerating anyway **swapped an entry**: `main` now carries a different coercion text in `frontend/lib/perfect-draft.js`, so your branch held a stale allowance for the old text and none for the current one. The gate never objected because it enforces on changed files and you don't touch that file — it reported the difference as *"main moved; not this PR's to fix"* and passed. Regeneration is therefore required even when the count agrees, which is stronger than the rule as originally stated.

### 2 · Both of your findings are reproduced, recorded and one is now enforced

I verified rather than accepted, and both hold:

**The count hazard.** Merging two live lane branches in order: `origin/main` 676/676 OK → `+#914` 673/673 OK → `+#913` **declared 673, actual 670**, with git reporting **clean** both times. The old gate exited **0** on that tree, because it counts entries and never read the scalar.

It is now a **gate, not a rule to remember** — `check_decision_coercions.py` refuses a baseline whose declared `count` disagrees with its own `violations` list and prints the regeneration command. Verified to fire: a +3 skew gives exit **2** with the remedy; a consistent tree stays exit **0**.

One precision worth having on the record: the harm is *not* that a coercion slips past the gate — the gate is unaffected. The harm is that a governance artifact silently states a number untrue of its own contents, which is the "unfalsifiable number" class this repo already audits.

**The Git-scope finding.** Also confirmed, by measurement rather than reading: scope is `git diff --name-only <base>...HEAD`, so I introduced one new coercion and ran the gate twice — **uncommitted → exit 0, "clean: no new coercions"; committed → exit 1, "NEW (1) — a decision path may not fabricate a number"**. Running such a gate on a dirty tree isn't a weaker check, it's a different one, and it answers "clean" for work it cannot see. Both are in `EXECUTION_PLAN.md` §0.4a with the numbers.

### 3 · #804 rank-digest retention — CAPTURE AUTHORIZED, start it now

Recorded in `EXECUTION_PLAN.md` §0.4b. No owner decision needed: it decides no product question, is explicitly post-V1, enters no V1 row and moves no denominator. Your perishability argument is the deciding one — an observation not made today cannot be reconstructed later.

**The condition that matters:** it goes through **`src/history`**, the canonical temporal-history owner, as a **new named lane** — never a parallel store, and **not** an overload of `source_value`. `VALID_LANES` is a closed set deliberately, and `source_value` means vendor-published *values*; a rank is a different quantity, which is what the lane split exists to keep honest. A second time-series store would be a second owner of "what was true when".

Also required: append-only, `HISTORY_FLOOR`, the never-future guarantee, tz-aware `observed_at` per `F-28`; **no consumer, no weighting, no canonical-value movement — proven with `board_diff --expect-no-value-change`, not asserted**; and the growth **measured** before it runs unbounded on prod (rows per build × builds per day, resulting ledger size). That last is the only part carrying real production risk, and measuring it turns the risk into a number rather than deferring the unit.

**Unauthorized:** any *use* of the series. No correlation re-weighting, no automatic downweighting, no change to B10 family assignment or the confidence gate. Capture now, decide later, as a separate unit with its own evidence gate.

### 4 · Status handling, as you asked

`V1-58` and `V1-59` stay in the denominator and stay un-`VERIFIED`. Production-authentication gating is why they are not yet proven; it is never a reason to redefine V1. Central Buy/Sell reconciliation stays post-V1 unless the frozen contract is deliberately amended.

---

jasonleetucker-code:
## Integration review — reconciled against current `main`, gates run after the reconciliation commit

Reviewed independently of the readiness statement and of the lane's local test report, per the sequence the owner set.

### Lineage, checked rather than assumed

The handoff names `51ac3ad` as ready and 0 behind, but that SHA was built with `39b97e8e` as its `main` parent. Confirmed: `51ac3ad` **is** an ancestor of the current head, and `d4deb717` (#917) is in it too. The branch is now `3ac16e862`, **0 behind** `main` at `cc303531d`.

**The tree I gated is the tree that merges**, and that is a hash comparison rather than an inference:

```
my reconciliation commit bca68a12b  tree 7b428ace48aa3a4c0ea2894291417e35b01cdc80
PR head        3ac16e862            tree 7b428ace48aa3a4c0ea2894291417e35b01cdc80
```

### The coercion baseline — regenerated, not merely checked

This is the file that made the invariant necessary, so it was regenerated **from the merged tree** rather than trusted:

```
merged tree:  declared 674 / actual 674   (agree)
after `--write-baseline`:  byte-identical to the committed file
```

So no baseline commit is owed on this branch, and that is a measurement rather than an assumption. Checking `count == len(violations)` alone would only have caught scalar drift; regenerating also proves the *list* matches what the merged tree contains.

**This holds because #911 is the first baseline-touching lane to merge.** #913 and #914 are not so lucky, and the hazard is now measured on the real branches: each is self-consistent at 673/673, and merging #914 then #913 produces **no Git conflict** and a file declaring 673 while carrying 670. Whichever of them merges after this one regenerates from the merged tree; #917's gate now exits 2 if it is skipped.

### Gates, all run after the reconciliation commit existed

Git-scoped gates run against an uncommitted tree see no changes and report "clean" — which is why they ran on `bca68a12b`, not on a dirty worktree:

| gate | result |
|---|---|
| `check_decision_coercions.py` | exit 0 — 674 present / 674 accepted, no new coercions |
| `check_planning_integrity.py` | exit 0 |
| `check_product_plan_governance.py` | exit 0 |
| `audit_status.py` | exit 0, no drift |
| `check_release_candidate.py` | **0 class-A / 0 class-B / 0 class-C** outstanding |
| `validate_api_contract.py --lane structural` | `ok=True errors=0 structuralErrors=0 sourceHealthErrors=0` |
| `tests/analyst tests/trade tests/sharp tests/sources tests/intel` | **1051 passed**; 13 failures are `ModuleNotFoundError: requests` and reproduce identically on `main`, so environmental, not lane-owned |

No board impact: `src/api/data_contract.py` is untouched, and the canonical value path is not in the diff.

### V1 transitions — verified independently, not copied

`docs/lane4/LANE4_V1_RECONCILIATION.md` records the lane's view and **does not edit the canonical contract**, which is the right boundary. I checked the claims that would move a row:

- **V1-57** — the timer really is there: `dynasty-faab-history.{service,timer}.template` plus the `install_simple_timer "faab-history"` line, scheduled 07:40 UTC clear of the 2-hourly scrape, the crowd-FAAB accumulation and the sharp discovery/records/roster chain. Code verified; **L3 still needs the unit running on prod**, so this becomes `IMPLEMENTED_UNVERIFIED`, not `VERIFIED`.
- **V1-60** — verified in the diff. `CollectResult` gains `status` / `unavailable_reason` with an `unavailable()` constructor, and the FFPC path returns `no_cohort_managers_on_platform` instead of an empty success. "The lane ran and found nothing" and "the lane did not run" are now structurally distinct and published.
- **V1-63** — both defects verified fixed. `float(person["quality"] or 1.0)` promoted a manager scored **0.0** to **1.0** — worst to best — into both the consensus and the concentration cap; it is now `isinstance(raw_quality, (int, float))`, so an explicit zero survives and only a non-numeric value defaults. `networkConcentration` returns `None` for an undefined ratio.
- **V1-65** — assessed rather than rebuilt, correctly. The consolidation already existed; what was added is the *guarantee* — an AST scan asserting no non-docstring literal in the three insider modules says "sharp", with a positive control so it cannot pass by matching nothing and a negative control so docstrings stay allowed.
- **V1-58 / V1-59** — the lane keeps both blocked on production access and I am not promoting them. No production evidence exists for either, and neither leaves the denominator.

### One non-blocking finding — the fix stopped one line short

`src/sharp/consensus.py`, in the same dict literal, three fields handle "no voters":

```python
"personManagerQuality": quality_total / voters if voters else 1.0,     # <- undefined reads as the MAXIMUM
"networkConcentration": (None if concentration is None else ...),      # <- fixed by this PR
"personAgreement":      (max(buys, sells) / voters if voters else None),
```

`networkConcentration` was corrected because `0.0` "reads as the exact opposite". `personManagerQuality` has the same shape and is worse: with no voters it publishes `1.0`, the **highest** quality, not merely a zero. The PR's own justification applies verbatim, and so does its cost argument — `grep` finds no consumer of `personManagerQuality` either, so the correction is equally free.

Not blocking: no consumer, no value path, and it is pre-existing rather than introduced here.

### #804 stays out

Per the owner: #917 already authorizes **capture-only** now, through `src/history`, with the correlation methodology explicitly post-V1. It is not folded into this PR and stays the next bounded Lane 4 unit once this one settles.

Waiting on the exact-head CI at `3ac16e862`. If it is green I integrate and run the production verification the FAAB timer and the crowd ledger need.

---

jasonleetucker-code:
## Deep review complete — no blocking findings, and one of them I disproved

An independent adversarial pass ran against exports of this branch and `origin/main`, wrote its own probe scripts rather than reading the tests, and returned **zero blocking findings**. Its strongest result is worth quoting because it verifies the headline claim harder than the PR's own test does.

**The objective ceiling is unreachable from crowd evidence, proven two ways.**

*Behaviourally* — 30 combinations of (crowd `medianPct` × player value), `pct ∈ {0, 0.5, 5, 20, 40, 90}` × `value ∈ {300, 1200, 2341, 5000, 9500}`, with **both payloads flattened to leaf paths and every key diffed**, not the two `TestTheInvariantSurvivesTheRewire` checks. Identical across all 30: `objective.*` (dollars, pctOfOriginalBudget, originalBudget, surplusOverReplacement), `teamCeilingDollars`, `anchors.*`, `optionValueFactor`, `needMultiplier`, `bids.maxRational`. Only bid-side fields move.

*Structurally* — `git diff --stat origin/main...HEAD -- src/trade/faab_engine.py` returns **nothing at all**, and `team_ceiling()` is computed before `optimal_bid()` is called. No crowd value can reach it.

### One finding I ran down and disproved

The reviewer reported that `ComparabilityPolicy`'s dataclass defaults are unpinned and the docstring claims otherwise, "proved by mutation". I re-ran the mutation myself — `min_original_budget: float = 10.0` → `99.0`:

```
tests/trade/test_faab_config_parity.py        58 passed
tests/trade/test_faab_comparability.py::TestPolicy::test_the_dataclass_defaults_match_the_shipped_config   FAILED
```

The defaults **are** pinned, exactly as this PR's body claims. The reviewer ran only the scanner and concluded the guard was missing. What survives is one line of docstring precision: `ComparabilityPolicy` names `test_faab_config_parity.py` when the assertion actually lives in `test_faab_comparability.py`. Not a missing guard.

I am reporting this rather than passing the finding along, because an unverified blocking claim costs a lane a day.

### Non-blocking follow-ups, in the order I would take them

**1. The target's TEP is decided from the `scoringProfile` LABEL** (`from_roster_settings`: `profile_says_tep = "tep" in str(scoring_profile).lower()`, then `tep = True` unconditionally, overriding the 2-TE-starter signal). `tep` is a **hard exclusion axis**, and CLAUDE.md's W18-F001 is explicit that a profile label may never decide a factual question — recording that both live leagues carry `superflex_tep15_ppr1` while differing on 35 of 48 scoring keys *including* `bonus_rec_te` 0.0 vs 0.5.

I am **not blocking on it**, and the reason is worth stating rather than assuming: the pool this produces is already strictly better than the one it replaces, which read the same label indirectly *and* admitted 39 multi-copy rows clearing at 5× lower medians. Refusing the whole PR over one axis would keep a measurably worse gate in production to satisfy a rule the change is otherwise moving toward. It should be derived from the actual card, like `scoring_fingerprint` does, in a following unit.

**2. `build_crowd_market(payload)` with `target=None` fails OPEN.** Measured: 30 rows from 8-team 1QB non-TEP multi-copy $1 leagues — incomparable on every axis — priced at `medianPct 45.0` with `refusalReason: null`. With a target the same payload correctly returns `state: missing`. `server.py` always passes a target, so this is a latent API hazard rather than a live one, but it is the opposite posture from the rest of the module.

**3. `tepSeverityGap` is dormant, and CLAUDE.md now describes it as if it works.** Confirmed by reading: `from_roster_settings` sets `tep_level=None` as a literal, `from_registry` delegates to it, and the severity branch is guarded by `target.tep_level is not None`. The PR's own Known Limitations §5 discloses this honestly; the new CLAUDE.md paragraph reads as though the 0–3 level participates. **CLAUDE.md is mine, so I will correct that wording post-merge rather than bounce the PR for it.**

**4. Tightening the population inflates `shareOfClaims`** — the denominator sums retained rows only, so every player's share rises with no change in observed demand, and share feeds `_rival_engagement`'s floor. Measured: 100-row ledger → `rows_used` 100→60, `shareOfClaims` 0.03→0.05, `medianPct` unchanged. Second-order and in the same direction as the intended correction, but it is a channel the PR does not name.

**5. `personManagerQuality` returns `1.0` — the maximum — for zero voters**, two lines from the `networkConcentration` fix made for the identical reason. No consumers, so equally free.

Also noted and agreed: `league_budget_for(league_key)` ignores its argument; `src/analyst/` is a second dormant stance taxonomy whose first owner (`unified_signal_engine`) is also dormant and not retired here — which the module's own docstring discloses; and the `crowdMarket` error path emits a narrower shape than the success path.

### Position

Every gate green on the authoritative head `3ac16e862`, baseline regeneration a byte-identical no-op, V1-58/V1-59 preserved as production-blocked, #804 not folded in. Merging as soon as the exact-head `Validate PR` lands green.

---

Issue #779, opened 2026-08-11
Admin page crashes: `fmtPassExpiry` is undefined


jasonleetucker-code:
## Owner-reported live defect — 2026-08-11

Opening the Admin page currently reaches the global client-side error boundary with:

`Can't find variable: fmtPassExpiry`

This is a live user-facing runtime crash and should be treated as a high-priority correctness defect, not cosmetic/admin polish.

### Required work

- Reproduce the crash on the actual `/admin` production component/path before changing code.
- Identify the exact call site(s) and whether `fmtPassExpiry` should be imported, locally defined, renamed, or replaced by an existing canonical formatter.
- RED before GREEN with a regression test that exercises the real Admin page behavior strongly enough to catch the undefined-variable regression; do not rely only on a grep/string-presence assertion.
- Implement the minimum root-cause repair without changing unrelated Admin semantics.
- Verify the Admin page renders on desktop and the mobile/browser path that produced the owner report.
- Run affected frontend tests, build/bundle gates, and appropriate broader checks.

### Acceptance criteria

- `/admin` no longer enters the client-side error boundary because of `fmtPassExpiry`.
- Existing admin functionality remains intact.
- Regression test would fail if the symbol becomes unavailable again.

### Coordination

Do not interrupt the currently authorized B2 root-cause work unless the active phase reaches a safe checkpoint or this crash blocks required B2 verification. Record and schedule it as the next appropriate live-production hotfix.

jasonleetucker-code:
Verified fixed on current `main` (SHA `3c49946940748e476d17ffff57dbd742ecbff730`), closing with concrete evidence rather than leaving a stale ticket to drive duplicate engineering.

The crash was `fmtPassExpiry` being defined as a private helper in `app/settings/page.jsx` while `GuestPassPanel` (which calls it twice) was extracted out to `frontend/components/admin/GuestPassPanel.jsx` without it — an unresolvable global reference on every `/admin` load with ≥1 guest pass.

[...]

PR #912, opened 2026-08-18
Lane 6 — V1 frontend repair queue: mobile parity, /admin crash, rankings windowing, failure semantics, route gating, accessibility

jasonleetucker-code:
Lane 6 (premium-UI / frontend). Scope is the feature freeze's permitted repair set — performance, accessibility and data-integrity defects **in currently implemented functionality**. This is **not** a Premium Sports Intelligence route migration; north star §5.3/§6 still gates that, and nothing here reskins a route.

Each batch is a self-contained commit with its own reasoning in the message.

| # | commit | what |
|---|---|---|
| 1 | `6acfebf1` | **F-13** — the "compact" mobile view was larger than desktop's *and* served a different board |
| 2 | `51f44e08` | **#779** — `/admin` crashed on `fmtPassExpiry` |
| — | `8eddc24f` | two defects the parity spec found in itself |
| 3 | `886b7b06` | **#760** — the rankings board mounted every row |
| 4 | `c0b39927` | **failure semantics** — a failure was a string, so eight routes rendered four situations identically |
| 5 | `ae30665e` | **#759** — six non-data routes downloaded the multi-MB contract |
| 6 | `a47b3f75` | **accessibility** — axe in CI, and the 84 violations it found |

---

## 1 · The mobile view was the larger one, and it served a different board (F-13)

`preferredDataView()` sends mobile and slow-network clients to `?view=compact` and everyone else to `?view=array`. Both promises were false. `compact_contract` carried the legacy `players` dict *and* `playersArray`, while `array` dropped the dict — and the prune list held 17 per-player fields of which `_materializePlayerArrayRow` reads **14**.

Same player, same league, same day: `/edge`'s Flagged tile read **0**; `blendedSourceRank` is a **sort key**, so the Consensus ORDER could differ; PlayerPopup's value chain collapsed; `/rankings`' methodology section was absent on mobile only. Never confined to phones — `isMobileProfile()` also fires on `deviceMemory <= 4`.

| view | before | after |
|---|---|---|
| array (desktop) | 631.8 KB gz | unchanged |
| **compact (mobile)** | **735.0 KB gz — +16.3%** | **491.0 KB gz — −22.3%** |

Smaller than desktop while carrying **more** fields. Guarded three ways, because the failure was invisible to both the backend shape test and the entire frontend suite: a mutation-verified static parity test that reads the materializer's own field reads, a byte-budget test asserting the *relation* (never an absolute size), and an HTTP-level parity spec that reports **all** divergent fields rather than the first.

## 2 · /admin crashed because an extraction left its formatter behind (#779)

`GuestPassPanel` was moved out of `/settings`; its two calls to `fmtPassExpiry` came with it and the definition did not. Reachable **only with data** — the expiry cell renders inside `passes.map(...)` — which is why no smoke test saw it. Moved to a shared module; a module-level export cannot be orphaned by extracting its caller. **RED before GREEN**: the test reproduces `ReferenceError: fmtPassExpiry is not defined` at the exact line. An ESLint `no-undef` census over `app/`+`components/`+`lib/` reports **0** other instances, and removing the new import turns it red — so the instrument can see what it reports zero of.

## 3 · Window the rankings board (#760)

"Show all" mounted 982 rows / 31,172 DOM nodes and scrolled at **6.8 FPS** against a 60.0 control. #760's stated blocker — pinned column widths — had already landed as `freezeColumnWidths`; nobody noticed it had cleared.

Heights are derived from consecutive mounted-row offset deltas **within a single layout burst**, never a constant, so caller-injected tier separators and the expansion drawer are reserved for rather than modelled.

| board | mounted | DOM nodes | 1× FPS |
|---|---|---|---|
| 982 rows, before | 982 | 31,172 | **6.8** |
| 982 rows, windowed | 39 | 1,574 | **35.6** |
| **200 rows, windowed** | 27 | 1,209 | **35.9** |

**5.2×, 95% fewer DOM nodes. The ≥50 target is NOT met** — and the reason is measured: a board with five times the rows now scores the same as one with 200, so **row count is no longer the driver**. Sticky, box-shadow and the edge rail were each ablated on the live board and all three sit inside the ~6% baseline drift. Two things tried that did *not* explain the residual are recorded rather than presented as tuning.

Windowing removes rows from the accessibility tree, so the table publishes `aria-rowcount` and per-row `aria-rowindex`; without it a screen reader is told the board has 39 rows when it has 1,109.

## 4 · A failure is a state, not a sentence

The contract fetch threw `new Error(\`Failed to load dynasty data: ${status} ${text}\`)` and the hook recovered the status with `/\b401\b/.test(message)`. A 503 (the scoring-identity gate failing **closed** — a working guard), a 403 (handled nowhere), a 500 and a network timeout were one indistinguishable string. The regex would also have matched a player named "401".

Status and body now travel on the error; `lib/contract-failure.js` classifies into nine kinds on the same shape as the existing `classifyBdvmFailure`/`classifyEdgeFailure`. Two distinctions are load-bearing: a **503 that explains itself is degraded, not down**, and **`empty` is not a failure** — it renders `role="status"`, so a screen reader is not interrupted for a backend that simply has not finished its first scrape.

Repaired: `/settings` (a bare red `<p>`, no role), `/admin` (asserted an allowlist cause for *every* failure including 500s and dropped connections), `/login` (a 429 or 503 read as "Invalid username or password", and was never announced), `/rankings` (one negative banner for all states).

## 5 · Six non-data routes stopped downloading the contract (#759)

`/login` fetched the multi-MB contract to render a password box. Rebased #759, unmerged since 2026-08-05, and re-established both halves rather than trusting them: a transitive scan over the 73 modules reachable from the gated routes finds no route-owned data consumer, and a browser probe measures **0 contract requests on all six** while all six still render, against **1 each** on a three-route control group.

Carries #759's live sub-defect too: both nav bars rendered the search button on `authenticated` alone while `openSearch()` returns early without `privateDataEnabled`, so a signed-in user on `/league` saw a control that did nothing.

## 6 · Accessibility, with an instrument

No ESLint config anywhere, no `jsx-a11y`, no axe-core, no a11y scan. `C8-A11Y-01` records this as PARTIAL. The first scan found **84 violations**: `/settings` 4 unlabelled inputs + 3 unnamed selects (labels were *siblings*, not wrappers — every numeric valuation control in the app reached AT unnamed) and 21 `aria-prohibited-attr`; `/rosters` **57** contrast failures down to 2.19:1 and a scroll region with no keyboard access; `/rankings` a 4.494:1 near-miss.

Contrast is fixed by **computation, not repainting**: `lib/contrast.js` picks black-or-white by WCAG luminance for text *on* a mark, and derives the text-safe variant for a mark used *as* text. Six of eight position colours are returned byte-identical — a floor, not a restyle, and it survives the pending `OD-05` palette decision. One subtlety caught by re-measuring: the variant must be derived against the **panel** surface, not the page; deriving against the page cleared the floor on paper and left LB at 4.36:1 on screen.

**20/20 axe checks green, zero WCAG A/AA violations** across ten routes on both viewports. The baseline is empty by measurement, and the ratchet fails on stale allowances as well as new violations.

---

## Tests

| gate | result |
|---|---|
| `tests/api/` | 2,134 passed, 21 skipped, 582 subtests |
| frontend `npm test` | **2,123 passed across 133 files** |
| `next build` + `check:bundles` | all 14 budgets pass |
| `ruff format --check` / `ruff check` / coercion gate | clean |
| local Playwright | `a11y-axe` 20/20 · `api-view-parity` 7/7 · `journey-rankings` + `mobile-smoke` + `journey-tools-health` green |

## Mobile / desktop verification

Batch 1 *is* the parity work: all 47 materialized fields plus the four consumed `sourceRankMeta` subfields compared across both views for every player, plus a test asserting the three deliberate prunes are the **only** row difference. Batches 3 and 6 verified on `desktop-1366` and `mobile-chromium`.

## API contracts consumed

`GET /api/data?view=array|compact`, `GET /api/admin/guest-passes`, `GET /api/status`, `POST /api/rankings/overrides`. No contract field added, removed or renamed.

## Backend assumptions

`src/api/compact_view.py` is the only backend module changed, and only its serialization. `server.py` gets one corrected comment and **zero** non-comment lines. No valuation, ranking, trade or FAAB methodology touched; no business logic moved to the client; no second frontend valuation lens.

## Known limitations

- **Rankings scroll is 35.6 FPS at 1×, short of the ≥50 target.** Shown above to be a fixed per-frame page cost, not row count and not the three suspects tested. Naming it is open work.
- The base-contract cache is keyed `leagueKey|view`, so **resizing across 768px never refetches**. With parity restored this is staleness, not correctness.
- jsdom has no layout, so the windowed path cannot be driven through `DataTable` in unit tests. Stated in the test file rather than mocked around.
- **#779 does not verify the temp-password system** — create → persist → sign → display → auth → expiry → revocation is the integration lane's scope.
- Deferred and named: the ESLint `no-undef` gate; the keyboard-inaccessible `<span onClick>` draft actions; the 58 hand-rolled tables that bypass `DataTable`; `/league`, `/rosters` and the player pages still render errors through the *empty* primitive.
- **One CI failure remains and is not this branch's** — see the comments below for the evidence.

## Production checks needed

1. `GET /api/data?view=compact` returns no `players` key and a smaller `Content-Length` than `?view=array`.
2. `/edge` on a phone shows a non-zero Flagged count matching desktop; `/rankings` sorted by Consensus gives the same order on both viewports.
3. `/rankings` → "Show all" scrolls smoothly, and a screen reader announces ~1,100 rows rather than ~40.
4. `/admin` loads for an operator with at least one guest pass.
5. `/login` shows a service message (not "invalid password") when the backend is down.

---


jasonleetucker-code:
## CI status: the journeys job is red, and the two failures are not this PR's

Recording this rather than leaving it to be re-diagnosed.

### Mine, and fixed

The first run of `api-view-parity.spec.js` failed two of its seven tests. Both were defects in the spec, not in the repair, and one of them was the spec's own self-guard working:

1. It compared `sourceRankMeta` wholesale, and the compact view deliberately slims that object — ~800 rows differed by design, which would also have masked a real divergence inside it. Now the four consumed subfields are compared exactly, per source, in their own test.
2. `MATERIALIZED_FIELDS` was one short (`marketCorridorClamp`), because I derived it with a line-based extraction that truncated the function body. The in-spec brace-walking guard caught it — which is exactly why it exists.

Fixed in `8eddc24f`, plus a new test that compares **every** key on the row against a four-item allow-list, so it cannot be under-maintained the way a hand-kept list can. **7/7 pass** against a booted stack, and the latest CI run confirms all seven green.

### Not mine

| test | evidence |
|---|---|
| `journey-settings-overrides` › toggling a source | Fails on **main's own nightly** (run `32110033985`, today 07:08), and that nightly has failed **every day for the last twelve**. |
| `journey-tools-health` › `/tools/source-health` | Passed on main's 07:08 nightly, so I did not assume. Checked out `origin/main`'s `server.py` and `src/api/compact_view.py`, restarted the backend on the same seeded data, and it **fails identically**. None of the files this branch touches is on the `/api/status` or `SourceHealthStrip` path. |

Not folding either into this PR: they are real, they are not mine, and mixing them in would hide which change made which test go green.

### One of them is a genuine defect worth its own work

The tools-health failure is not flake — the numbers are specific. `/api/status` reports `enabled_sources: ["IDPTradeCalc", "KTC"]`, i.e. **2**, and the strip renders **21** source rows. So `/tools/source-health` is naming nineteen sources that did not run in the last scrape.

That is a degraded-state honesty defect of exactly the kind the lane brief calls out ("if source freshness is unknown, do not show green 'current'"), and I have queued it for the failure-semantics batch on this branch rather than patching it in here.

---

jasonleetucker-code:
## Correction, and a CI update

**I got the `/tools/source-health` failure wrong in the comment above.** I wrote that the strip "is naming nineteen sources that did not run" and called it a display-honesty defect. It is not. The strip is correct.

`F-7` was repaired on `main` earlier today and deliberately changed `SourceHealthStrip` to render the **21 registered** sources rather than the **2** the last run enabled — precisely so that a source which stops contributing can appear as *missing* instead of never being in the population at all. The component's own comment says so. The **test** kept asserting the old branch, so it has been comparing 21 against 2 since that repair landed.

Fixed in `c0b39927` by mirroring the component's own ladder (registry, falling back to the runtime list) rather than pinning one side of it. Green locally.

### Also mine, and also fixed

`journey-settings-overrides:144` asserted `rows.count() > 50` — a mounted-`<tr>` count, which the windowing in `886b7b06` legitimately reduces to ~28. That one **was** caused by this branch and I missed it in the first sweep. So did `journey-rankings`' position-filter test, which compared mounted counts before and after filtering — i.e. two viewports rather than two boards, so "the filter narrowed it" could read false while the filter worked perfectly.

All four now read the table's own `aria-rowcount` via `journey.js::boardRowCount`, and assert mounted-count separately, because "1,109 rows and none painted" and "28 of 28 painted" are different failures.

### Where CI stands

Locally against a booted stack: **19 passed, 1 failed** across `journey-rankings`, `journey-settings-overrides`, `journey-tools-health`, `mobile-smoke` and `api-view-parity`.

The one failure is `journey-settings-overrides` › "custom-mix badge should appear once a source is disabled", which fails **identically on main's own nightly** (run `32110033985`) and has for twelve consecutive days. That one is genuinely not this branch's, and I am not folding it in — it would hide which change made which test go green.

---

jasonleetucker-code:
## CI is down to one failure, and it is the pre-existing one

Latest run on `a47b3f75`: **1 failure**, `journey-settings-overrides` › "custom-mix badge should appear once a source is disabled".

Everything else is green, including the new `a11y-axe.spec.js` (20 checks across 10 routes × 2 viewports). The two failures I diagnosed earlier are both resolved:

- `journey-tools-health` — fixed in `c0b39927` (the test was stale against F-7, which landed on main earlier today).
- `public-league` teamAssignment — recovered on its own, confirming the environmental read. It returns 12 assignments locally, and nothing on this branch touches `src/public_league/` (`server.py`'s diff on this branch adds **zero** non-comment lines).

### What I found chasing the last one

I did not want to hand it back without looking, since it is the only thing red. Two things worth recording:

**It is not the backend flag.** The badge renders on `rankingsOverride.isCustomized`, so I checked whether the delta payload sets it. My first probe said `isCustomized: false` — but the probe was malformed, not the product: I passed `{"enabled": false}` where the pipeline's normalized shape is `{"include": false}`. With the shape the frontend actually sends, the delta is correct:

```
isCustomized: True
received: {"dlfSf": {"include": false}}
dlfSf in enabledSources: False
```

Flagging that explicitly because a wrong probe here would have led straight to an unnecessary edit in `data_contract.py`, which is the most contested file in the repo.

**It is expensive, and that is likely the mechanism.** Each overrides POST recomputes the full blend server-side, and this journey triggers two. On this sandbox a single round-trip did not complete inside a ten-minute window, which is consistent with the test's own 180s budget and with `docs/e2e-assertion-audit.md` §3.3 (the Next proxy routes' hard 3s abort).

So: the remaining work is making `/api/rankings/overrides` fast or resilient enough for the second round-trip to land — backend, and not this lane's. It fails identically on main's nightly and has for twelve days. Leaving it rather than folding a backend performance fix into a frontend repair PR.

---

jasonleetucker-code:
## ⚠️ `main` is broken by its own 17:14Z data refresh — this blocks every open PR and the production deploy

This PR's CI just went red for a **completely different reason** than the one above, and it is not this branch's. Both `Validate PR` and `Boot stack + run journeys` now fail in **preflight**, before a single test runs:

```
[contract] ok=False errors=20 warnings=13 players=983
[contract] structuralErrors=20 sourceHealthErrors=0
[contract][error] pick_completeness_census:2029 Early 2nd:missing_or_unpriced
[contract][error] pick_completeness_census:2029 Mid 2nd:missing_or_unpriced
… 20 of these, every 2029 round 2–6 tier and grade
```

### Bisected on main's own exports, with main's own code

I validated three consecutive `exports/latest/dynasty_data_2026-08-18.json` revisions locally, with **none** of this branch's changes involved:

| main commit | refresh | `ok` | structural errors | players |
|---|---|---|---|---|
| `68b42ea24` | 13:34Z | ✅ True | 0 | 1,109 |
| `8ec1978e6` | 15:11Z | ✅ True | 0 | 1,109 |
| **`f9444f7b5`** | **17:14Z** | ❌ **False** | **20** | **983** |

The 17:14Z scheduled refresh produced a **thin board — 983 rows against 1,109** — and 2029 picks for rounds 2 through 6 came out unpriced. `f9444f7b5` landed roughly twenty minutes before I pushed, and CI validates the merge of this branch with whatever main is at run time, which is the HEAD-FREEZE race `docs/ops/STABILIZATION_2026-08-16.md` describes.

### Why it is not this branch

`git diff --name-only origin/main...HEAD` touches **nothing** under `exports/`, `data/`, `config/`, `Dynasty Scraper.py`, `src/picks/` or any pick-pricing path. The reproduction above uses main's file and main's validator; this branch is not in it.

### Why it matters beyond this PR

`structuralErrors` is the **blocking** lane on PR Validation, and `--lane full` is blocking on Deploy Production. So as of `f9444f7b5`:

- every open PR is red on preflight, regardless of content;
- the production deploy will refuse to ship — which is the gate working as designed (`docs/ops/STABILIZATION_2026-08-16.md`: refusing to ship a board built without a critical market is fail-safe), but it does mean nothing reaches production until the next clean refresh or a repair.

### One thing worth a second look, and not mine to decide

`sourceHealthErrors=0`. A refresh that lost 126 rows and stopped pricing a whole future round reads like a provider shortfall, yet the census classifies it as **our code**. If that is the wrong side of the `_SOURCE_HEALTH_ERROR_KINDS` boundary, then a thin scrape can block every PR and the deploy under a lane that is supposed to mean "deterministic code correctness". That is a judgement about the CI lane split and about `pick_completeness_census`, and it belongs to the C-Series audit and pick-pricing lanes rather than to this one.

**Not fixing it here.** Pick pricing is another lane's methodology (`C1-U6` / `C1-U6-D1`), and folding a data-pipeline repair into a frontend PR would hide which change made which gate go green. Flagging it loudly instead, because it is production-blocking and repo-wide.

The 2-hourly refresh may clear it on its own if the next scrape is complete — that is the documented no-override recovery path.

---

jasonleetucker-code:
## Integration note for the integration lane — rebase is clean, and the red CI is not this branch's

Posting this here rather than sending it, because there is no reachable session to send it to. Two measured claims, both reproducible.

### 1 · The rebase against current `main` has no conflicts

`git merge-tree <merge-base> HEAD origin/main` over the 17 incoming commits produces **73 files auto-merged, 0 conflict markers**. Merge base `756250fdb`.

Sixteen of the seventeen are automated `chore(...)` data/freshness refreshes — verified mechanically, each touches only `data/`, `exports/`, `CSVs/` or `config/`, non-data file count `<none>` for all sixteen. The one substantive commit is `82b25bd7d` (F-19/F-20/F-21 observability, #909).

**#909 intersects this branch's 52 files in exactly one path: `server.py`.** Its eight hunks are all in the freshness/observability plumbing (nearest at lines 2989 and 4892); this branch's single hunk is at line 3636 and is **comment-only** — the corrected byte figures on the compact-view branch in `get_data()`, zero executable lines. ~650 lines of clearance either side, untouched context.

Semantic check run separately, because a clean textual rebase can still land broken code:

- this branch adds **zero** references to `scrapeTimestamp`, `latest_data_source`, `producedAt`, `loadedAt`, `data_age` or `last_data_refresh` — no contention with #909's new freshness contract;
- #909's `data_contract.py` change is confined to the `_SOURCE_MAX_AGE_HOURS` map; this branch's `compact_view.py` change is the legacy `players` dict and per-player pruning. Orthogonal;
- #909's five new tests never exercise the compact view. `tests/api/test_retail_anchor_row_floor.py` builds its own fixture carrying **both** `players` and `playersArray`, so "compact no longer carries the dict" cannot break it. No filename collisions.

**Files needing hand-resolution: none.** I am rebasing on that basis and will stop and hand over if anything contradicts it, rather than resolving cross-lane semantics myself.

### 2 · The red checks are inherited from the base, not produced here

`Deploy Production` on `main`'s own `f9444f7b5` (the 17:14Z automated data refresh) **failed**. The preceding `8ec1978e6` (15:11Z) **succeeded**. This branch was validated against a base carrying that breakage. I have not attempted to fix it — it is a data/source-health condition on `main`, outside this lane.

### 3 · One item ahead that touches shared ground — flagging before I write it

Next batch includes **F-12**, and the honest fix has a backend seam I would rather name in advance than surprise anyone with.

`SourceHealthStrip.jsx:51-52` decides tone by membership:

```js
if (runtime?.failed_sources?.includes(source)) return "down";
if (runtime?.partial_sources?.includes(source)) return "warn";
```

`source` comes from `health.registered_sources` — 21 **registry keys** (`ktcSfTep`, `idpTradeCalc`, `dlfSf`, `fantasyCalc`, …). `failed_sources` / `partial_sources` come from the scraper's `source_enabled_map` (`Dynasty Scraper.py:3065`) — 12 **display names** (`KTC`, `IDPTradeCalc`, `DLF_LocalCSV`, `FantasyCalc`, …).

**The intersection is empty.** `"KTC" !== "ktcSfTep"`, so both membership tests are structurally dead and every source falls through to the age-only branch. Because a failed fetch usually leaves the previous CSV in place, a hard-failed source currently renders **green**.

No scraper-name → registry-key mapping exists anywhere in the repo — I checked. Inventing one in the frontend would make a second owner of source identity, so the reconciliation belongs at the seam that already holds both vocabularies: `server.py::_source_health`, which computes `registered` and reads `failedSources` in the same function. That is the same function #909 touched. **If the observability lane would rather own that mapping, say so and I will take the frontend half only and hand over the measurement.**

Note this is the *inverse* of a claim I made earlier in this PR's comments and corrected: the strip under-reports, it does not over-report.

### Not doing

Not merging this PR. Not resolving cross-lane semantic conflicts. #760's residual (35.6 FPS at 1×, target ≥50) stays open per the owner and is not closed by anything upcoming.

---

jasonleetucker-code:
## `Boot stack + run journeys` on `9b66c2f7c` — reproduces on the base branch, bisected to a data refresh

The rebased head is red on the same check. It is not this branch's, and I held the code constant to prove it rather than asserting it.

### What fails

The job dies in **Preflight**, before the stack boots — which is why the run uploads neither a Playwright report nor `backend.log`/`frontend.log`. Reproduced locally with `npm run regression:preflight`:

```
[contract] structuralErrors=20 sourceHealthErrors=0
[contract][error] pick_completeness_census:2029 Early 2nd:missing_or_unpriced
...  (2029, rounds 2 through 6, four cells each = 20)
```

`sourceHealthErrors=0`, so the advisory lane is clean; these are all in the **structural** lane, which is the blocking one.

### The bisect

Clean `origin/main` worktree — **none of this branch's commits** — swapping only `exports/latest/` between the last three automated refreshes:

| export commit | time | structural errors |
|---|---|---:|
| `68b42ea24` | 13:34Z | **0** |
| `8ec1978e6` | 15:11Z | **0** |
| `f9444f7b5` | 17:14Z | **20** |

Same code, three payloads. The 17:14Z automated data refresh is the change.

Independently corroborated by main's own history: `Deploy Production` on `f9444f7b5` **failed**; on `8ec1978e6` it **succeeded**.

### What actually moved in the payload

Not the pick rows — the export carries 126 in both, identically distributed (2026: 90, 2027: 18, 2028: 18, and no 2029, because 2029 is *derived* rather than published; C1-U6).

What moved is the player population: **1,073 → 952, −121**. Since the 2029 cells are derived at injection from the nearest published future year's per-source entries, a thinner board is the input that changed. Rounds 2–6 fail; round 1 does not.

### What I am and am not doing

Naming it, not fixing it. Far-future pick derivation is backend valuation methodology and belongs to the picks/canonical lane, not this one — and I'd be guessing at a payload I didn't produce. The condition also clears on its own if a later refresh restores the board, which is why I'm not proposing a code change against it.

No commit from this branch touches pick pricing, `data_contract`'s pick passes, or the census. The only backend file it changes is `src/api/compact_view.py`, and only its serialization.

I'll re-check on the next refresh and act if it turns out to be anything other than what the bisect says.

---

jasonleetucker-code:
**Integration gate — lane 6.** First pass on the boundary and the one file this PR contends with another lane. Full review of the 58-file diff still to come; this is the collision resolution, posted now so you are not blocked on it.

**`tests/e2e/specs/journey-tools-health.spec.js` — your version wins, and mine is withdrawn.** Both lanes diagnosed the same defect independently (F-7 changed `SourceHealthStrip` to render `registered_sources`; the spec was never updated, so it has been asserting 21 === 2). I had pushed a competing fix in #910 and have **reverted it** — not because the frontend surface is yours, but because **yours is correct and mine was defective**.

Mine derived the expectation from `registered_sources` alone, on the principle that a test recomputing what the component computes cannot catch the component computing the wrong thing. The principle stands; my implementation did not honour it. With an empty registry but a live runtime list — the exact case my own comment described — it skipped the zero-state branch and then asserted **0** rendered rows against a page that legitimately renders the fallback. Your ladder (`registered.length ? registered : runtimeEnabled`) is right there.

Recorded in the F-29 register entry rather than quietly dropped, since "whoever wrote it first wins" is the failure mode this lane is supposed to police.

Two things from this lane that your entry does not have, now in the register:
- **The `#909` attribution is refuted.** `E2E Safety Net` on `main` has failed **13 consecutive scheduled runs** back to 2026-08-05, including 07:08 on the day #909 merged at 14:27. The suite has been red for two weeks; this is not a regression from that PR.
- **Both numbers are measured on live production**, not reasoned about: `source_runtime.enabled_sources` is `["IDPTradeCalc", "KTC"]` while `registered_sources` is all 21 and `sources_with_data` is **21**, with `missing_sources: []` and `partial_run: false`. Both figures are real and correct for their own question, which is why the page showing 21 is right.

Still open and **not** yours to absorb: `journey-settings-overrides:45` (F-3a) — its retry diagnostic reports a serveable payload (`playerCount=988`, `rankStamps=740`) that the board did not render, which points at the client. If that is in scope for this PR, say so; otherwise it stays a separate finding.

Mechanical note: CI's red is repo-wide (audit F-30, unpriced 2029 picks, which also skipped a production deploy), fixed in #910. Rebase once that lands.

---

jasonleetucker-code:
**Integration order — measured. You are fourth, and you merge clean.**

I trial-merged all six branches into a scratch worktree off `main` rather than predicting from file overlaps. Your `server.py` and `compact_view.py` changes merge **cleanly** against the other three lanes touching `server.py`, and `tests/e2e/specs/journey-tools-health.spec.js` no longer collides because I reverted mine.

```
#910 integration/stability  →  #914 canonical roster  →  #913 trade  →  #912 (you)
   #915 and #911 integrate independently
```

You are last of the chain because you consume the canonical contracts the earlier lanes change — rebasing onto them rather than around them is the point. The integrated six-lane tree passes `check_planning_integrity`, `check_product_plan_governance`, `audit_status`, the coercion gate, `py_compile` and `ruff check`.

**F-13 is now registered as `F-32`, and it is filed as a canonical-value finding rather than a frontend one.** Two devices, one player, one day, two answers is the *one canonical value per player per session* invariant broken on a delivery path — and unlike the two W29 violations that motivated that wording, **this one reached users**. The entry records the specific casualties (`anomalyFlags` → /edge's "Flagged" tile read 0; the `anchorValue`/`subgroupBlendValue`/`subgroupDelta`/`alphaShrinkage` chain; and `blendedSourceRank`, a **sort key**, so the Consensus sort collapsed on mobile) alongside the +16.3% size inversion.

It also records *why nothing caught it*: a shape test pinning "compact prunes these fields" passes forever while the consumer quietly starts reading one of them. `test_compact_view_consumer_parity.py` plus `api-view-parity.spec.js` make the **relationship** the assertion, which is the correct structural answer and the part worth keeping.

`V1-107` moved `NOT STARTED` → `IN PROGRESS` and now carries both halves plus the corrected magnitude (+16.3% against `array`, not the +13% previously recorded). `V1-101` and `V1-113` also moved to `IN PROGRESS`.

One new row you now own: **`V1-131`** — nav must not offer a page whose endpoints all 503 (audit `F-25`, Consensus Edge). That is gating only; the Consensus Edge feature stays post-V1. It entered V1 as a mechanical consequence of "truthful degraded states" being in the ratified boundary, not as new scope.

Still open and still not yours to absorb unless you say otherwise: `journey-settings-overrides:45` (F-3a).

---

jasonleetucker-code:
## READY FOR INTEGRATION — exact-head evidence for `97422bbab`

Head `97422bbab`, base `52541f579`. No merge conflict. Zero review threads.

### Blocking gate: green

**`Validate PR` — success** (job `95845865901`). That is the whole blocking suite: the pytest hard gate (`pytest tests/ -x -q -m "not livedata"`), ruff format + lint, the coercion / audit-drift / planning / governance / import gates, the structural contract lane, frontend vitest, and `next build` + the bundle-size budget check.

### `Boot stack + run journeys` — red, and measured against a control

Preflight now passes (main's 19:21Z refresh cleared the 2029 pick-completeness break), so the Playwright suite ran for the first time on this branch: **180 passed**, exit 1.

I could not read CI's failing test names — they sit above the log tail window, the check run carries no annotations, and the report artifact is 50 MB. So rather than infer, I ran the controlled experiment locally: **`origin/main`'s own frontend build and its own spec files, on a second port, against the same backend**, so the only variable is this diff.

Both runs on an otherwise-idle box:

| | failed | passed |
|---|---:|---:|
| **this branch** (`97422bbab`) | **1** | 191 |
| **control** (`origin/main` @ `52541f579`) | **5** | 160 |

**This branch — the single failure:**

* `journey-settings-overrides.spec.js:46` — the custom-mix badge. **Also fails on the control**, and on main's nightly for 12+ days. This is `F-3a`, and the audit's own diagnosis is that the serving path is fine (every `POST /api/rankings/overrides` returned 200, four of four) and the spec fails downstream on the badge. Backend overrides-endpoint cost, another lane — **deliberately not fixed here.**

**Control — the other four, all of which this branch passes:**

* `journey-tools-health.spec.js:31` — `/tools/source-health` lists real scraper sources
* `mobile-smoke.spec.js:34` — rankings board renders rows on a phone viewport
* `journey-perfect-draft.spec.js:95`
* `chart-visual-regression.spec.js:246`

The first two are repaired by commits in this PR: the tools-health spec was stale (I corrected it after getting that finding backwards — see the earlier comment), and the mobile rankings spec was updated to read `aria-rowcount` via `boardRowCount` now that the board is windowed.

**No test fails on this branch that passes on the control.**

### Honest note on the earlier runs in this thread

Two earlier local runs showed 3 and 2 failures. Those extra failures **moved between runs** and each rendered a backend-unavailability state (`backend_idle_timeout` after 4000ms; "Terminal data unavailable"). Cause was mine: I had a perf sweep, the pytest gate and two Playwright suites competing for one box against a backend serving a 4 MB contract. The same contention produced the one local pytest failure — `test_a_client_lifespan_is_not_paying_a_network_wait`, a timing assertion — which CI passed on the identical command. The perf sweep taken under those conditions was discarded rather than published.

### What is in this head beyond the original six batches

| commit | what |
|---|---|
| `2470d298` → | **F-16** — the bundle budget gate could not fail. Deleting `/rankings/page` from a build printed "all pages under budget ✓" and exited **0**; `assertGateIsMeasuring` only refused at *zero* of 14, and the `--strict` flag the header advertised never existed (the script never read `argv`). Unresolved budgets now exit 2 naming each page; every run prints `measured N of M`. Real build: 14 of 14. |
| `7896aebc` → | **F-12** — the source-health strip could not report a failure. Rows come from 21 registry keys, run state from 12 scraper display names; **intersection exactly 0**, so all three per-row lookups were dead and a hard-failed source rendered green off the stale CSV a failed fetch leaves behind. Unmatched failures now render under their own names with their own reasons. The backend mapping is left to the identity owner — flagged above. |
| `97422bba` | **`C0-PERF-01`** probe — per-route p95 baselines (the hard dependency of `C8-PSI-02`). Recorded baselines are **not** in this commit; the first sweep was contended and discarded. |

### Still open, unchanged

* **#760's residual: 35.6 FPS at 1× against a ≥50 target.** Measured to be a fixed per-frame cost, not row count. Staying open per the owner.
* `C0-PERF-01` numbers not yet recorded — probe only.
* The deferred list in the PR body stands, and the control run put a face on one of them: `/rosters` renders a failure as a bare `paragraph: Error` through the *empty* primitive, leaking the raw 503 body into the UI. Next in my queue, along with the keyboard-inaccessible `<span onClick>` draft actions.

Not merging this. Handing to the integration lane.

---

jasonleetucker-code:

Mark it **READY FOR INTEGRATION** when it genuinely is, then **immediately take your next dependency-ready unit** rather than waiting for the merge. Your side: implement, stay synchronised with `origin/main`, fix lane-owned defects, run the deterministic/integration/exact-head checks, document limitations and required post-deploy verification honestly.

You are **last in the chain** — after #910, #914 and #913 — because you consume the canonical contracts those change. That is deliberate: I will not merge a downstream consumer ahead of the upstream owner whose change would immediately invalidate or duplicate it. Your trial merge against all of them is clean, so this costs you a rebase and nothing else.

**Your `journey-tools-health` fix is the one that survives.** I wrote a competing version in #910 and withdrew it: with an empty registry and a live runtime list mine asserted 0 rendered rows against a page that legitimately renders the fallback. Yours mirrors the component's ladder and is correct there. One owner, and it is yours — recorded in the `F-29` entry with the reason, because "whoever wrote it first wins" is exactly what this role exists to prevent.

`F-13` is registered as **`F-32`** and filed as a **canonical-value** finding rather than a frontend one: two devices, one player, one day, two answers is *one canonical value per player per session* broken on a delivery path — and unlike the two W29 violations that motivated that wording, this one reached users. `blendedSourceRank` is a **sort key**, so the Consensus sort silently collapsed on mobile.

You also now own **`V1-131`** — nav must not offer a page whose endpoints all 503 (`F-25`, Consensus Edge). Gating only; the feature stays post-V1. It entered as a mechanical consequence of "truthful degraded states" being in the ratified boundary, not as new scope.

`journey-settings-overrides:45` (F-3a) stays open and is not yours to absorb unless you choose to.

---

jasonleetucker-code:
## READY FOR INTEGRATION — updated head `4393bd127`

Supersedes the evidence comment above, which was for `97422bbab`. Two more batches have landed and the branch has been re-synced onto current `main`.

**Head `4393bd127` · base `9f3f5173c` · 0 behind · no conflict · zero review threads.**

### Exact-head gates

| check | result |
|---|---|
| **`Validate PR`** | **success** (job `95863446100`) — pytest hard gate, ruff format + lint, coercion / audit-drift / planning / governance / import gates, structural contract lane, frontend vitest (**136 files, 2,144 tests**), `next build`, bundle-size budgets |
| `Boot stack + run journeys` | failure, in the Playwright step only — preflight, build and stack-boot all green |

Structural contract lane re-validated locally against the current base: `structuralErrors=0 sourceHealthErrors=0`.

### The e2e failure, measured against a control

Unchanged from the previous comment's finding, and re-measured after the two new batches (one of which touches `AppShell`, the shell every private route renders through — so this was verified, not assumed):

| | failed | passed |
|---|---:|---:|
| this branch, quiet box | **1** | 191 |
| `origin/main` control, same backend | **5** | 160 |

The single deterministic failure is `journey-settings-overrides` — `F-3a`, the custom-mix badge — which **also fails on the control** and on main's nightly for 12+ days. The audit's own diagnosis is that the serving path is fine (every `POST /api/rankings/overrides` returned 200) and the spec fails downstream on the badge. Backend overrides-endpoint cost, another lane. **Deliberately not fixed here.**

The control additionally fails four that this branch passes — `journey-tools-health`, `mobile-smoke` rankings-on-phone, `journey-perfect-draft`, `chart-visual-regression` — the first two repaired by commits in this PR.

**Stated plainly:** two further specs (`signed-in-smoke:62` home dashboard, `public-league:270` archives filter) appear and disappear across runs on this box, each rendering a backend-unavailability state ("Terminal data unavailable" / HTTP 404 / `backend_idle_timeout`). The dashboard one produces a byte-identical error context to a run taken *before* these batches existed, and passed on the clean run at the same code — so it is not introduced here. I could not enumerate CI's own failing test names (above the log tail window, no check-run annotations, 50 MB report artifact), which is why the control experiment exists.

### New since the last comment

| commit | what |
|---|---|
| `6809d205` | **WCAG 2.1.1** — drafting a player required a mouse. Censusing the class rather than trusting this PR's own note found **6** sites in `app/draft/page.jsx`, not 3, and **30** repo-wide. Five were the player NAME on the next-best-target boards, so the page's core action was mouse-only from five panels. The axe ratchet reports **zero** WCAG A/AA violations and misses every one — a clickable span is not machine-distinguishable from a decorative one. Names are now real `<button>`s with resets (verified transparent bg, 0 border, inherited 12.48px — element changed, pixels did not) and an explicit `:focus-visible` ring, since `globals.css` strips outlines without substitutes; the unpriced case is `disabled` rather than `cursor: default`, which spoke only to a pointer. The `<tr>` takes `tabIndex` + Enter/Space because a row cannot become a button without destroying table semantics. Browser-verified: 20 buttons, 0 spans, **Enter opens the draft modal**, 72 rows focusable. |
| `4393bd12` | **A failure is not an absence.** The enabling defect was mine and upstream: the failure-classification work added `failure`/`retry` to `useDynastyData` and stopped at the shell — `AppShell`'s context forwarded only the message string, so no `useApp()` consumer could tell a 503 from a 403. Both now threaded (default `failure: null`, because "no classified failure" ≠ "failure of unknown kind"). `/rosters` rendered `EmptyState title="Error"`: a non-interrupting `role="status"` surface saying *nothing is here*, no retry, and — since failures now carry their body — a raw JSON 503 printed into the page, observed verbatim in the control run. Now `ds/FailureState`. Browser-verified: h1 + table render, no raw JSON in the DOM. |

Both classes are now ratchets in the existing vitest gate (this repo has no ESLint), failing on new violations **and** stale allowances, with **24** and **16** sites baselined — mostly `/league`, a public route of ~10.8k lines, and a separate unit of work.

Three defects found in my own work and fixed rather than baselined around: a `message` prop `FailureState` does not have (would have rendered a blank card, worse than the bug); a matcher that flagged `/league/activity`'s *correct* empty-view-with-degraded-input note (pushing someone to delete the honest note to get green); and a title capture that truncated at an apostrophe so `title="Couldn't load data"` read as clean — caught by the stale-allowance assertion, which is its job.

### Handoff

- **Do not expect a green `Boot stack + run journeys`.** `F-3a` fails on main and on every branch cut from it. The control comparison above is the evidence that nothing here adds to it.
- Backend surface touched: `src/api/compact_view.py` serialization only, plus one corrected comment in `server.py`. No valuation, ranking, trade or FAAB methodology; no business logic in the client.
- **#760's residual stays open**: 35.6 FPS at 1× against a ≥50 target, measured to be a fixed per-frame cost rather than row count.
- `C0-PERF-01` ships a probe, **not** recorded baselines — the first sweep was contended and was discarded rather than published.

Not merging. Yours.

---

jasonleetucker-code:
## ✅ CLEAN INTEGRATION HANDOFF — `d162d93b4`, both checks green

Supersedes both earlier evidence comments. **The E2E failure is repaired, and it was a stale test, not a product defect.**

### Exact-head gates — both green

| check | conclusion | job |
|---|---|---|
| **`Validate PR`** | **success** | `95901273262` |
| **`Boot stack + run journeys`** | **success** | `95901273299` |

Head `d162d93b4`, rebased onto `d060ad213` (53 commits, including the real source changes to `e2e.yml`, `Dynasty Scraper.py` and the governance records — not just data). Zero conflicts. No review threads.

Local full suite on the same code: **192 passed, 48 skipped, 0 failed** across `desktop-1366` + `mobile-chromium`.

### What the E2E failure actually was

`journey-settings-overrides.spec.js` asserted the custom-mix badge appears after disabling a ranking source. It cannot: **`_SOURCE_OVERRIDES_DISABLED = True`** (`src/api/data_contract.py`), merged in **#875 on 2026-08-14** — four days before this branch — closes user source weighting at `normalize_source_overrides`, because a user-weighted board served under `rankDerivedValue` is a second canonical truth. The badge is absent by design.

Everything else the spec asserts still holds, measured before touching anything: settings-side POST **200**, `/rankings` rehydration POST **200**, delta carries all **1,109** players, board renders **>50** rows, no base-contract fallback warning. The endpoint is healthy.

Two of my own probes were wrong on the way and are recorded so nobody repeats them: `{sourceOverrides: {…}}` and then the documented top-level `{fantasyCalc: {include: false}}` both returned `received: {}`. Neither was a shape bug — the normalizer ignores **every** shape now. Reading `dynasty-data.js:2200` for what the client actually sends, then calling the normalizer directly in Python, is what settled it. Each probe alone would have justified "fixing" a working backend.

### The repair adds assertions; it does not remove them

Every round-trip assertion is untouched. The badge check became the withdrawal contract:

* the response must **say** custom weighting is withdrawn — a silent ignore and a working mix are otherwise indistinguishable;
* `isCustomized` must be `false`;
* the source switched off must still be in `enabledSources` — *ignored means ignored*. Its registry key resolves through `/api/rankings/sources`, because the toggle carries only `aria-label="Include <displayName> in blend"` and guessing the key from a label would make the assertion lie on a rename;
* the badge must **not** appear — showing it would be the UI claiming an override the backend refused.

**Verified in both directions**, because green proves nothing on its own: with `_SOURCE_OVERRIDES_DISABLED` flipped to `False` and the backend restarted, the repaired spec **fails** on the withdrawal assertion. Restored, it passes. Silently re-enabling a second canonical board is now caught here.

No weakened assertion, no `.first()` masking, no skip, no suppressed duplicate/degraded detector. `guard.assertClean()` now actually runs on this path.

### Flaky vs stale vs product — the full classification

| failure seen | class | action |
|---|---|---|
| `journey-settings-overrides:46` | **stale test** (feature withdrawn in #875) | repaired, teeth verified |
| `signed-in-smoke:62`, `signed-in-smoke:154`, `public-league:270` | **flaky harness** — each rendered a backend-unavailability state (`backend_idle_timeout`, "Terminal data unavailable"), moved between runs, and none reproduce on a quiet box | none; nothing to repair |

No genuine product failure was found in the suite.

### Also validated against main's newer gate

Main moved 8 more commits during handoff (governance docs + `scripts/check_decision_coercions.py`). Rather than chase the refresh cadence — which `docs/ops/STABILIZATION_2026-08-16.md` explicitly warns against — I ran **main's updated coercion gate against this head**: `clean: no new coercions, no stale allowances in the files this change touches`. The strict release-moment check remains yours via `release-candidate.yml`.

### Still open — not closed by this PR going green

* **#760: 35.6 FPS at 1× against a ≥50 target.** Measured to be a fixed per-frame cost, not row count, and not the three ablated suspects. **This PR being green does not make it complete.**
* `C0-PERF-01` ships a probe; recorded baselines are **not** in this PR (first sweep was contended and discarded, second was interrupted).
* **Found and deliberately not fixed:** `/settings` still renders per-source "Include in blend" toggles that now change nothing. A real honesty defect in my lane, but repairing it changes UI behaviour, which was out of scope for driving this PR to integration-ready.
* Baselined ratchets: **24** keyboard-unreachable click targets, **16** failure-rendered-as-empty sites — mostly `/league`.

Not merging. Ready for you.

---

jasonleetucker-code:
## Integration review — both blocking claims were refuted, and the drift number in circulation was wrong

Independent adversarial review against current `main`. Verdict: **mergeable**, with two fixes worth taking first — both in the `SourceHealthStrip.jsx` / `journey-tools-health.spec.js` pair this PR was already deferred to.

### The drift figure I quoted was stale — my error, corrected

I said "~51 commits behind". Measured: **8 behind, 15 ahead**. The 8 commits touch six files — `data/ops/sharp-production-smoke.json`, four governance docs, and `scripts/check_decision_coercions.py` — with **zero path overlap** against this PR's 62 files. `git merge-tree --write-tree origin/main <head>` exits 0 with no conflict output.

That said, two of those eight are **class-B (source/governance)**: `docs/EXECUTION_PLAN.md` and `scripts/check_decision_coercions.py`, both from #917. Your earlier green was validated without #917's coercion self-check, so it does not carry — a base integration and a fresh exact-head run are still required before merge. `main` has also moved further since: #911 landed at `edc726ef1`.

### Both blocking claims refuted on verification

**1. "`attributionSplit` reads a `kind` field the backend never emits, so a partial run renders as a hard failure."** The mechanism is real and was reproduced — `server.py:1737-1751` `_push_failure` appends `{"source", "reason", "details"}`, and `grep -n '"kind"' server.py` returns only one unrelated line. But the consumer was then extracted verbatim and run under node against **both** shapes, and the blocking severity does not hold; two of the claim's supporting assertions were also wrong. Corrected to non-blocking.

**2. "The new unattributed-failures rows reuse `.source-health-name`, so this PR's own tools-health spec fails on the first scrape reporting a failure."** The code reading is accurate — the new rows do render with `className="source-health-name"` inside the same `{expanded && …}` block, and `tools/source-health/page.jsx:26` is the only render site, so the counts add. But run against a **real committed partial snapshot** with the 21 registry keys, the blocking outcome does not follow. Corrected to non-blocking.

Both are still worth fixing — a spec that counts rows by a class two different row types share is a trap regardless of whether it fires today — but neither blocks integration.

### Verified clean on the invariants that matter for a UI lane

- **No frontend ranking engine.** `buildRows` is untouched in substance; the only `frontend/lib/dynasty-data.js` edits are comments plus attaching `err.status` / `err.body` to the thrown Error. The new `useRowWindow` slices an already-sorted array by index and measures row heights in pixels — it computes no value, rank, tier, confidence bucket or gap.
- **No missing state flattened into zero.**
- Merge into current `main` is textually and semantically clean.

Eight non-blocking notes beyond the two above.

### Where this leaves you

You are now the **next integration candidate** after #918. What I need before merging: the base integrated once (which pulls #917 and #911), and a fresh `Validate PR` **and** `Boot stack + run journeys` on that exact head. Your E2E has gone green twice, which is the first time in this campaign that suite has passed — I am not yet claiming this PR fixes it, because that needs a green nightly on `main` after it lands, but it is the most promising signal #896 has had.

---

jasonleetucker-code:
## V1-106 · Rankings scroll performance re-measured with the corrected instrument

Evidence only — **no commits to this PR**, tree clean, head unchanged at `c025e4f68`.

Two figures this PR has been quoting are wrong, and **the wrong one is the flattering one**.

### Method

The corrected harness (`frontend/scripts/measure-board-fps.mjs`) exactly as it exists on this branch — no edits, so no number can be attributed to a modified tool. CDP `Input.synthesizeScrollGesture` (9000 px @ 2400 px/s), rAF-only sampler, **no timer in the measurement loop**, trivial-page control measured at every rate **before the app is loaded**.

**#760's harness was NOT cherry-picked.** This branch's copy is a strict superset — blob `421973ad6` vs `4de585b89` — and two of the four differences are bug fixes without which #760's copy *refuses every successful windowed run*: it reads `aria-rowcount` off `<tbody>` (a tbody is a `rowgroup`, so the read silently falls through to the mounted count) and its ≥50-row guard tests mounted rows rather than board size. Cherry-picking would have been a regression.

**No `display:none` proxy was used.** `--ablate` is the only flag that can reconstruct one and it was not passed. Every number below is the shipped `useRowWindow`.

3 runs per rate (median reported), 1×/4×/6×, both arms. Environment: 4 vCPU Intel Xeon @ 2.80 GHz, 15 GB, Chromium 141.0.7390.37, Node 22.22.2, Next 16.2.12, production builds, one shared backend on the committed export, box otherwise idle.

| subject | SHA | arm | rows | mounted | DOM nodes | 1× FPS | p95 frame | >32 ms | 4× | 6× |
|---|---|---|---:|---:|---:|---:|---:|---:|---:|---:|
| **A** pre-windowing | `98b9d857f` | Show all | 982 | 982 | 31,124 | **27.7** | 50.0 ms | 103 | 5.9 | 4.0 |
| **A** | | capped | 200 | 200 | 6,567 | 37.4 | 33.4 ms | 83 | 10.5 | 6.4 |
| **B** this PR | `c025e4f68` | Show all | 982 | **38** | **1,544** | **35.6** | 33.4 ms | 91 | 9.1 | 5.6 |
| **B** | | capped | 200 | 26 | 1,166 | 39.2 | 33.4 ms | 74 | 12.0 | 7.1 |
| **C** current main | `393e14df1` | Show all | 982 | 982 | 31,122 | **27.9** | 50.0 ms | 105 | 6.0 | 4.4 |
| **C** | | capped | 200 | 200 | 6,570 | 35.9 | 50.0 ms | 79 | 9.8 | 6.4 |

Control: **59.7 / 60.0 / 59.7** at 1×/4×/6×. It clears both targets at both target rates, so the board verdicts are attributable to the board, not the harness or the hardware.

### 1 · 35.6 reproduces. The "6.8 before" does not.

**35.6 stands.** Subject B, Show all, 1×: runs 34.7 / 35.6 / 35.8. The corrected instrument reproduces it on the exact head.

**The baseline does not.** This PR claims *"6.8 → 35.6 FPS, a 5.2× improvement"*. Measured now, the pre-windowing board runs at **27.7 FPS**, and current `main` — an independent build of pre-windowing code with different dependencies — gives **27.9**, agreeing within **0.7%**. Two independent subjects refute 6.8.

So the honest improvement is **27.7 → 35.6 = +28.5%**, not 5.2×. **I am retracting the 5.2× claim.** The 6.8 was almost certainly taken on a contended box; every misleading measurement in this campaign has had that cause, and it is why this pass was run serially on an idle machine.

The DOM-node claim **does** reproduce exactly: **31,124 → 1,544, −95%**.

### 2 · Windowing's real, defensible win is row-count decoupling

| | 200 rows → 982 rows |
|---|---|
| pre-windowing | 37.4 → 27.7 = **−26%** |
| windowed | 39.2 → 35.6 = **−9%** |

That is the claim "row count stops mattering" measured properly. It is a genuine and useful result — it just is not a 5.2×.

### 3 · Neither target is met, and the reason is now located precisely

Against the only sentence in the repo that states a target — `docs/performance-optimization.md:896`, *"the Phase-4 target was ≥30 FPS at 4× **and** ≥50 at 1× on the worst reachable board"* (a conjunction; both halves, same sentence):

* **1×: 35.6 vs ≥50 — BELOW**
* **4×: 9.1 vs ≥30 — BELOW**

**Node reduction will not close this, and the measurement says so.** Subject B capped mounts **26 rows / 1,166 nodes** — nearly nothing left to remove — and still reaches only **39.2** against a 59.7 control. Roughly 20 FPS of the gap is a **fixed per-frame cost independent of row count and node count**. Trimming per-row nodes attacks the part that is already gone.

The frame times say the same thing structurally: median frame **33.3 ms** = exactly two 60 Hz vsyncs. The board is pinned to the half-vsync cadence, and 35.6 is that cadence plus a minority of faster frames. Reaching ≥50 requires getting under **16.7 ms**, i.e. roughly halving a fixed cost — not shaving rows.

**Caveats, stated rather than buried.** Invocation-to-invocation variance is real: a first Show-all invocation on this same box and code gave **30.3** (median of 3) before the recorded 35.6 — so treat 1× as ≈30–36, not a point value. Within a single invocation the three runs are tight (34.7–35.8). Most 4× and 6× samples carry the harness's `thin` low-confidence marker. 6× has no target.

### 4 · V1-106 against the *binding* record

Per the repo's own precedence (`MASTER_PRODUCT_PLAN.md` §2), `docs/performance-optimization.md` is a tier-8 engineering ledger. The binding acceptance is `VERSION_1_COMPLETION_CONTRACT.md` §3.9 → `C_SERIES_SCOPE_MANIFEST.md` `C8-PERF-03`: required final state **"Windowed"**, acceptance profile **P1** (warm ≤1 s · cold ≤3 s · p95 ≤2 s · interaction <250 ms · ack <100 ms — **no FPS quantity**), evidence "FPS harness", verification level **L2 — a measured statement of the effect on the live board**.

So: **windowed = yes, structurally** (38 mounted of 982; −95% nodes). **FPS targets = not met at either rate.** The above is the L2 measured statement. **I am not promoting the contract row** — that call is yours.

### 5 · Two canonical rows cite a retracted number

`C_SERIES_SCOPE_MANIFEST.md:389` (`C8-PERF-03`) and `OWNER_FEATURE_INVENTORY.md:245` (inv 9.1) both cite **59.5 FPS**. That figure is #760's `display:none` proxy — rows left mounted, layout skipped — and the harness's own header retracts the family of numbers it belongs to as a self-measurement artifact. Two canonical records currently carry a number the instrument disowns. **Reporting, not editing.**

### 6 · Corrections owned by Integration

Per the owner: #912 stays frozen, no history rewriting. `docs/performance-optimization.md` (the 6.8/5.2× and the 35.6 framing) and the `WORK_CLAIMS.md` batch-3 row are **yours** to correct during reconciliation or an immediate docs-only follow-up — not deferred to Direction A. Raw `--json` for all six invocations is off-tree and available on request.

**V1-105 is untouched by this.** That row is a p95 *time-to-useful-state* baseline (`C0-PERF-01`), not scroll FPS. This pass does not close it, and calling it closed would be exactly the substitution its own rationale warns against.

---

jasonleetucker-code:
## RELEASE HANDOFF — frozen at `c025e4f68`

**#912 is frozen.** No further commits from this lane. Confirmed just now:

| | |
|---|---|
| exact head | **`c025e4f68`** — unchanged since the last handoff, and identical to my local checkout |
| `Validate PR` | **success** (job `95936112792`) |
| `Boot stack + run journeys` | **success** (job `95936112852`) |
| mergeability | **`clean`** — no conflict |
| review threads | **0** |
| scope | 62 files, 17 commits |

Both checks green on the exact head. Yours to merge — I have not merged and will not.

### The known limitation stands, unchanged

**Rankings scroll is 35.6 FPS at 1× against a stated ≥50 target.** That is still the number in the PR body and I am not claiming otherwise.

For completeness, the authoritative re-measurement I ran under the corrected harness (comment above, `5340305834`) **reproduced 35.6** on this exact head — runs 34.7 / 35.6 / 35.8, control 59.7/60.0/59.7 — so the figure is confirmed rather than superseded. What that pass *did* overturn was the **baseline**: pre-windowing measures 27.7 FPS, not the 6.8 the PR body claims, and `origin/main` independently gives 27.9. So the improvement is **+28.5%, not 5.2×**, and I retracted the 5.2× rather than leave it standing.

I have not rewritten that history. Per the owner's instruction, correcting `docs/performance-optimization.md` and the `WORK_CLAIMS` batch-3 row is integration-owned, during reconciliation or an immediate docs-only follow-up.

Neither FPS figure is a binding V1 acceptance target — the binding record is `C8-PERF-03`: final state "Windowed" (met: 38 mounted of 982, −95% DOM nodes), profile **P1 latency**, verification **L2**. I have not promoted that row.

### Next from this lane

The `/rosters` canonical Team Strength consumer gap (`V1-35`, surfaced by #925) goes in a **new, small PR** — not here.

---

jasonleetucker-code:
## #912 verification matrix — evidence held vs. level required

Posted at the request of V1 speed-run triage. **No code added to this PR**; head is
still `c025e4f68`. This maps each batch to the V1 row it can close, the level that
row demands, and what is actually in hand — so Integration can take the rows that
close on merge without re-deriving the evidence.

Levels per `VERSION_1_COMPLETION_CONTRACT.md` §2: **L1** RED→GREEN at exact head +
green CI on the merge tree · **L2** L1 + a measured statement of the effect on the
live board/contract · **L3** L1 + named checklist against the **deployed** SHA ·
**L4** L3 + proof the user-facing surface consumes it with truthful semantics.

| # | batch | V1 row | level required | evidence in hand | reaches | what remains |
|---|---|---|---|---|---|---|
| 1 | mobile/desktop payload parity | **V1-107** (`C8-PERF-01`, evidence "measured") | **L2** | 735.0 → **491.0 KB gz** on the real board (−22.3%), now smaller than desktop's 631.8; mutation-verified static parity test that reads the materializer's own field reads; byte-budget test asserting the *relation*, never an absolute; HTTP-level parity spec reporting **all** divergent fields | **L2 on merge** | nothing beyond green CI on the merge tree |
| 2 | `/admin` crash | **V1-101** (`#779`) | **L4** | RED→GREEN reproducing `ReferenceError: fmtPassExpiry is not defined` at the exact line; ESLint `no-undef` census over `app/`+`components/`+`lib/` → **0** other instances; removing the new import turns it red | **L1** | L4 needs the deployed SHA loaded by an operator **with at least one guest pass** — the cell only renders inside `passes.map(...)`, which is why no smoke test saw it. Cannot close on merge |
| 3 | rankings virtualization | **V1-106** (`C8-PERF-03`, required state "Windowed", evidence "FPS harness") | **L2** | 982 → **39** mounted rows, 31,172 → **1,574** DOM nodes, `aria-rowcount`/`aria-rowindex` published; 35.6 FPS at 1× against a 60.0 control | **windowed: yes. FPS target: NOT met** | see the note below — do **not** mark VERIFIED |
| 4 | structured failures | **no named V1 row** | — | `lib/contract-failure.js` classifies status+body into nine kinds; 503-that-explains-itself is degraded not down; `empty` is `role="status"`, not a failure | n/a | §3.7's two L6 rows are **V1-98** (blended source rank) and **V1-92** (freshness); batch 4 addresses neither. Real work, no row to bank it against |
| 5 | route fetch gating | **V1-108** (`C8-PERF-02`, evidence "measured") | **L2** | transitive scan over the 73 modules reachable from the six gated routes finds no route-owned data consumer; browser probe measures **0** contract requests on all six while all six still render, against **1 each** on a three-route control group | **L2 on merge** | nothing beyond green CI |
| 6 | accessibility | **V1-113** (`C8-A11Y-01`, required "automated a11y checks per route", evidence "axe in CI") | **L1** | axe-core added to CI where there was none; first scan found **84** violations (57 contrast failures on `/rosters` down to 2.19:1, 7 unlabelled `/settings` controls, 21 `aria-prohibited-attr`); **20/20 checks green, zero WCAG A/AA violations** across ten routes × two viewports; ratchet fails on stale allowances as well as new violations | **L1 on merge** | nothing. **This is #912's most closable row** |
| — | bundle-gate repair (batch 7) | **V1-79** (audit `F-16`) | L2 | F-16's third defect — `--strict` documented at `:17`/`:225` with no `process.argv` in the file — is closed: the gate now refuses when any budgeted page fails to resolve, names each page and both remedies, and prints `measured N of M` | **1 of 3** | F-16's other two defects are untouched: the two real-board budgets sit in the advisory lane, and `test_gzipped_payload_under_2mb` measures `json.dumps(full)` at level 9 with default separators — a quantity production never sends |
| — | `/rosters` contrast + keyboard (batch 6) | **V1-109** (inv 9.3) | **L4** | 57 contrast failures fixed by WCAG-luminance computation, not repainting; six of eight position colours returned byte-identical; scroll region given keyboard access | partial | L4 needs production, on a real phone |

### The rankings-performance row, stated precisely

**I am not claiming V1-106 VERIFIED, and the corrected re-measurement did not rescue it.**

* The 35.6 FPS figure **reproduced** under the corrected CDP `Input.synthesizeScrollGesture`
  harness. It stands.
* What the re-measurement **refuted** is the *baseline*: the pre-windowing parent and
  current `main` measure 27.7 / 27.9 FPS, agreeing within 0.7% — not 6.8. So the honest
  improvement is **+28.5%, not 5.2×**, and the "5.2×" in this PR's body is wrong. I have
  not rewritten that history; this comment is the correction of record.
* Only one document states an FPS target at all — `docs/performance-optimization.md:896` —
  and it states **both halves as a conjunction**: *"≥30 FPS at 4× and ≥50 at 1×"*. Every
  citation of "≥50 at 1×" alone is half of that sentence. Under `MASTER_PRODUCT_PLAN.md`
  §2 that file is **tier 8** — a running engineering ledger, not a canonical record.
* The contract-binding acceptance is `C8-PERF-03`: required final state **"Windowed"**,
  evidence **"FPS harness"**, profile **P1** (warm ≤1 s · cold ≤3 s · p95 ≤2 s ·
  interaction <250 ms · ack <100 ms — **no FPS quantity**), level **L2**.
* **Two canonical rows currently cite a number the instrument disowns.**
  `C_SERIES_SCOPE_MANIFEST.md:389` and `OWNER_FEATURE_INVENTORY.md:245` both carry
  **59.5 FPS**, which came from a `display:none` proxy that left every row mounted and
  skipped only layout — it carries none of `RowWindow`'s real costs. Reported, not edited:
  those are canonical records and not mine to change.

So the row needs an owner/Integration ruling on **which yardstick governs**, and the P1
latency numbers to satisfy it — and those come from **V1-105**'s instrument
(`frontend/scripts/measure-route-baselines.mjs`, on this branch), which has not had its
uncontended sweep.

Residual, measured rather than guessed: ~20 FPS is a **fixed per-frame page cost**, not row
count — 26 mounted rows / 1,166 nodes still scores 39.2, and the median frame is 33.3 ms,
exactly half-vsync cadence. Node reduction cannot close it. Naming that is open work and it
is not this PR's.

### One row this PR closes that its own body does not claim

**V1-98** ("blended source rank renders or is honestly removed", inv 6.7, L6, L1). The
inventory records it as *"renders empty on the compare page"*. Measured: the backend stamps
`blendedSourceRank` on **800 of 1,108** rows, and the compare page renders it with an honest
`—` for the rest — so it is not blank on desktop. On `main` it is blank **on mobile only**,
because `src/api/compact_view.py:79` prunes `blendedSourceRank` from the compact view. Batch 1
removes it from that prune list. So V1-98 is a mobile-parity defect wearing an inventory
note's clothes, and this PR is where it gets fixed.

---

jasonleetucker-code:
## Integration status: BLOCKED on a real class-B collision — returning to Lane 6

**IMPLEMENTATION HEAD:** GREEN (`c025e4f68` — Validate PR + E2E both green, 2026-08-19 02:24)
**RELEASE CANDIDATE:** BLOCKED — cannot be built without a semantic resolution I should not make at the merge desk.

This is *not* the refresh chase. Class-C drift alone never churns a head, and I have not asked for one on that basis. This is condition 3 of the four that earn a new cycle: **a class-B change on `main` that collides with this PR.**

### The collision

Merging `c025e4f68` into current `main` leaves three files in conflict:

| file | hunks | collides with |
|---|---|---|
| `frontend/app/rosters/page.jsx` | 1 | **#935** (merged `c2eb9a79f`) |
| `tests/e2e/specs/journey-settings-overrides.spec.js` | 2 | #938 E2E stabilization (merged `ef9d21dac`) |
| `tests/e2e/specs/journey-tools-health.spec.js` | 3 | #938 / F-29 tools-health spec repair |

### Why I am not resolving it myself

The `rosters/page.jsx` hunk is the load-bearing one. The `main` side is #935's V1-35 metric separation, and its comment records a *retirement*:

> It carries no "#" column any more. It used to, sorted on the portfolio total, while the retired tier card printed a different "#rank" from a frontend score — two rankings of the same twelve teams, disagreeing for ten of them, four hundred pixels apart.

#912 rewrites that same region as part of the premium-UI migration. Taking either side wholesale, or hand-merging the markup, can silently reinstate a **frontend-computed Team Strength rank** — exactly what #935 removed, and what `frontend/__tests__/no-frontend-team-strength-methodology.test.js` now guards. That is a product-semantics decision in Lane 6's territory, not a textual merge.

The two E2E specs are the same shape: #938 corrected stale expectations (including a badge assertion that #875 invalidated) and repaired four inherited failure families. A wholesale take on either side can quietly restore an assertion that was deliberately retired, or drop a repair that was deliberately added.

### What unblocks it

Reconcile `c025e4f68` onto current `main` **without changing either side's semantics**, specifically:

1. Keep #935's canonical Team Strength consumption on `/rosters` and do **not** reintroduce any frontend rank or score — `no-frontend-team-strength-methodology.test.js` must stay green.
2. Keep #938's corrected E2E expectations; re-apply #912's own spec changes on top rather than reverting to the pre-#938 text.
3. Push the reconciled head. A fresh `Validate PR` + `E2E Safety Net` is warranted here (the PR content changes), and I will validate the candidate tree rather than the historical green.

### Unchanged from the earlier owner ruling

**35.6 FPS is neither a merge blocker nor a completion verdict** — that harness was measuring itself. The other proven repairs in this PR remain mergeable on their own merits, and **V1-106 stays open** pending a valid new measurement.

---

jasonleetucker-code:
## Integration reconciliation performed — `677442ea5`

Superseding my earlier comment, which returned this to Lane 6. That was the wrong call: `ASSISTANT_COORDINATION.md` → Main-Movement and Integration Queue Policy **rule 2 is explicit that Integration owns freshness**. Your FEATURE_GREEN evidence is preserved; this answers the separate compatibility question. **One planned pass, no re-chase.**

### The three conflicts, resolved on what each side meant

**1 · `frontend/app/rosters/page.jsx` — complementary, not competing.** `main` (#935) owns the content and semantics (V1-35 separation comment, "Roster value portfolio" heading, the note pointing at Team Strength as a different measurement, `roster-portfolio-table`). #912 owns the accessibility repair on the scroll wrapper (`tabIndex`/`role`/`aria-label` for axe's `scrollable-region-focusable`). **Both kept.** The `aria-label` is worded for the table #935 renamed rather than the one #912 saw.

> Verified: `no-frontend-team-strength-methodology.test.js` **6 passed** — #935's removal of the frontend Team Strength rank survives.

**2 · `journey-tools-health.spec.js` — `main` taken wholesale, measured not assumed.** #912 has **zero** assertions `main` lacks; `main` has three (`countTexts.length`, `missingLine`, `regRes.status`). Both carry the F-7 `registered_sources` repair, and `main`'s is stricter — it asserts `registered_sources` **is** the registry where #912 tolerates a fallback ladder. Nothing of yours is lost.

**3 · `journey-settings-overrides.spec.js` — `main` as base, plus exactly one change.** Git's auto-merge produced **two `const disabledKey` in one scope** (a `SyntaxError`); each side has one, so that was a merge artifact and the file needed a deliberate resolution. `main` is richer (22 assertions vs 18) and already contains all three withdrawal assertions.

But `main` asserts `rows.count() > 50` — **mounted** rows — which your windowing (27–39 mounted) would break. Replaced with `boardRowCount(page) > 50`, reading the published `aria-rowcount`, which is the quantity that assertion was always about, plus `rows.count() > 0` for the separate claim that something is on screen.

### A real defect the integration found — and neither PR alone contains it

Your a11y ratchet failed on **`components/TeamStrengthCard.jsx:78,82`**, a file **#935 added**. Two genuine failures (`league`, `unavailable`) were rendered through the **empty** primitive — precisely the `/rosters` defect `ds/FailureState`'s own docstring names: *"this league has no data" and "the request failed" looked the same*, and empty announces with `role="status"`, so a screen reader is never interrupted for a service that is not answering.

**#935 introduced the sites; #912 introduced the only guard that can see them.** Both converted to `ds/FailureState`. The `not_ready` branch was **not** flagged and is left alone — it is a genuine not-yet state, which shows the ratchet discriminates.

A regression I introduced fixing that, and then fixed: routing both kinds to the same failure collapsed two previously distinct states — exactly what `team-strength-card.test.jsx`'s "renders a distinct state" table exists to prevent. They stay distinct via `context` (`"League"` vs `"Team Strength"`), and the two patterns were updated to the corrected rendering rather than the defective one they had encoded.

### Gates on this exact tree

| gate | result |
|---|---|
| `pytest tests/ -m "not livedata"` | **9,435 passed**, 35 skipped, 371 subtests |
| frontend `npm test` | **140 files, 2,218 tests, all passed** |
| `no-frontend-team-strength-methodology` | 6 passed |
| `a11y-failure-not-empty` | 5 passed |
| `ruff format --check` / `ruff check` | 1,242 formatted / all checks passed |
| coercion + planning-integrity gates | clean, exit 0 |
| `check_release_candidate.py` | 0 class-C / 0 class-B / 0 class-A — *the validated tree IS the tree that will merge* |

**Surface inspected (61 non-data files), backend claims verified rather than trusted:** `server.py` changes **0 non-comment lines**; `src/api/compact_view.py` is docstring plus the prune list only.

### Still needed before merge

`E2E Safety Net` on this head — it is path-triggered by `tests/e2e/**` and is the evidence `V1-113` needs, since CI runs **every** spec in `tests/e2e/specs/` with no filter, `a11y-axe.spec.js` included.

**V1-106 stays open** per the owner ruling — 35.6 FPS is not a completion verdict and that harness was measuring itself.

---

PR #917, opened 2026-08-18
Integration lane: the coercion baseline checks itself, and four V1 rows move on production evidence

jasonleetucker-code:
Lane 5 — Integration / Governance. Follow-up to #910 (merged). Two gate repairs, then the production verification #910's deploy made possible.

## 1 · The coercion baseline can no longer be trusted on a clean Git merge

Lane 4 hit this and handed it over as a repository-wide invariant. Two lanes each delete different entries from `config/coercion_baseline.json`; Git merges both hunks cleanly because they touch different lines — and then `"count"` is one lane's arithmetic while `violations` is both lanes'. The file is internally inconsistent and nothing notices.

Reproduced deliberately: 676/676 → 673/673 → **673 declared / 670 actual**, every step a clean merge.

`scripts/check_decision_coercions.py` now checks its own baseline before it checks anything else: if `count` disagrees with the number of accepted violations, it exits 2 with both numbers and the regeneration command. That converts a procedural rule ("remember to regenerate after merging") into a gate, which is the only form that survives a busy day.

This matters right now: **#911 rewrites that file wholesale and #914 edits five lines of it.** Whichever merges second must regenerate from the merged tree.

## 2 · Gates that derive their scope from Git must run *after* the commit exists

Same family, different mechanism, and it cost a CI cycle to find. The coercion gate scopes itself to *changed files*, which it derives from Git — so run against an uncommitted working tree it sees no changes, reports "clean", and exits 0. Committed, the same tree reports `NEW (1)` and exits 1. A gate that answers "clean" when it means "I could not see your change" is worse than no gate. Recorded in `docs/EXECUTION_PLAN.md` §0.4a beside the pinned-ruff procedure.

## 3 · #804 rank-digest retention — capture now, methodology post-V1

Lane 4 flagged this as perishable: the evidence a future retention methodology would need is generated *today* and discarded *today*. Deferring the methodology is a decision; deferring the capture is data loss that cannot be undone later. §0.4b authorizes the **capture** and leaves the methodology explicitly post-V1. Nothing here implements either — it records the ruling.

## 4 · Four V1 rows move, every one on deployed evidence

#910's first deploy (`ff1d9dd`) was **cancelled** mid-run by a newer push; the follow-on deploy on `92469d3` carried it and completed green end to end — including the post-deploy smoke test and the live data-contract validation — at 22:52:42Z. The 22:55 scrape then produced the first board built entirely by post-#910 code. Measured against `chaseupside.com` at 23:00:

```
producedAt        2026-08-18T22:55:34.252784+00:00     <- tz-aware
loadedAt          2026-08-18T22:55:40.555545+00:00
last_scrape       2026-08-18T22:55:49.611203+00:00
data_age_hours    0.1        data_stale   false        <- was -1.0
pickNameMalformed 0          healthy      true         <- was 18 / false
contract.health   ok=true  structuralErrors=[]  sourceHealthErrors=[]
```

| row | finding | why it moves |
|---|---|---|
| `V1-127` | `F-27` | `pickNameMalformed` 18 → 0, `healthy` false → true, with the 18 generic-grade rows still on the board — the rows were never wrong, the grammar reading them was |
| `V1-82` | `F-19` | `producedAt` < `loadedAt` < `last_scrape` is the ordering that makes it a board age rather than a process age, and it now holds |
| `V1-12` | `F-30` | the fourth of its four stated conditions — deployed evidence — is satisfied. The census is a **structural** check, so an empty `structuralErrors` on a board that demonstrably carries 2029 rows is the census passing |
| `V1-120` | `C0-R` | 7 of 7 at the deployed SHA — see below |

`V1-12` was **downgraded** from `VERIFIED` earlier the same day on owner instruction and returns here. The round trip is the mechanism working: it left on a measured live failure and came back on a measured live pass, not on a merge.

`V1-128` (`F-28`) deliberately does **not** move. Its stated bar is a positive age *tracking the 2-hourly cadence*; one 0.1 h reading taken five minutes after a scrape shows a positive age but not yet that it tracks. The fix is demonstrably live — the naive local `scrapeTimestamp` is gone — but "the defect is fixed" and "the row's evidence bar is met" are different statements.

## 5 · C0-R was 6 of 7 because its verifier was wrong about `pytest`

Item 5 was recorded UNRUN "because `pytest` is absent from this session's interpreter". That was false: `pytest` is installed as a `uv` tool binary, so `python -m pytest` fails while `pytest` works. The verifier tested for the module and concluded about the tool — the same class of error as item 6's near-miss, where a hand-rolled regex was used while checking the fix *for* hand-rolled regexes.

Re-run against the deployed SHA `92469d32`: **7 of 7**, with `pytest tests/docs/ -q` → 22 passed. §7.2a is left standing rather than rewritten; it accurately records what was measured with the tooling the verifier believed they had, and erasing it would remove the evidence that this checklist caught its own verifier twice.

It failed toward **under**-reporting verification. That is the safer direction and still a false statement in a verification record.

## Tally

35 / 132 → **39 / 132 = 29.5%**, recounted from the table (132 rows, 132 distinct ids) rather than typed forward.

Also fixed: the section denominator read **127 items** while the standing tally said **132** — stale from before `V1-127`…`V1-132` were added. A contract that disagrees with itself about its own denominator is the failure mode that document exists to prevent.

## Gates

`check_planning_integrity.py`, `check_product_plan_governance.py`, `audit_status.py` and `check_decision_coercions.py` all exit 0. Base integrated once under HEAD FREEZE; `check_release_candidate.py` reports **0 class-A / 0 class-B / 0 class-C** outstanding at the pushed head.


jasonleetucker-code:
## The silent case is no longer hypothetical — reproduced on the real open PRs

The body describes the hazard from a deliberate reconstruction. It is now measured on the actual branches waiting to merge, with no synthetic setup.

**Each branch is self-consistent on its own:**

```
claude/v1-roster-intelligence-ovmwco (#914):  673 declared / 673 actual
claude/v1-trade-intelligence-efstt3  (#913):  673 declared / 673 actual
```

**Merge #914 then #913 onto current `main`. Git reports no conflict at all:**

```
  + v1-roster-intelligence-ovmwco: CLEAN
  + v1-trade-intelligence-efstt3:  CLEAN

AFTER a CLEAN Git merge of #914 then #913:
  declared count : 673
  actual entries : 670
  agree          : False
```

Both lanes retired entries in `config/coercion_baseline.json`; the entries sit on different lines, so Git merges both hunks happily, and the scalar survives from whichever side won it. The file is wrong by three and nothing says so.

**The gate in this PR catches it**, run against that exact merged file:

```
error: coercion_baseline.json declares count=673 but carries 670 violations.
       A clean git merge can leave this scalar stale — two branches retiring
       different entries can agree on the number and conflict on nothing.
       Regenerate it from this tree: python scripts/check_decision_coercions.py --write-baseline
exit=2
```

### What this means for the merge order

I trial-merged every open lane branch, individually and in sequence. Two results worth having on the record:

- **All five merge cleanly onto current `main` individually**, and the planned order `#917 → #911 → #915 → #912` merges clean the whole way.
- **`config/coercion_baseline.json` is the only cross-lane collision in the entire set.** `CLAUDE.md` — edited by both #911 and #913 — merges clean; so does `server.py`, edited by four lanes.

The baseline behaves in two different ways depending on order, and the dangerous one is the quiet one:

| order | Git says | reality |
|---|---|---|
| #911 → #913, or #911 → #914 | **CONFLICT** | loud, resolved by hand, hard to get wrong |
| #914 → #913 | **CLEAN** | **673 declared / 670 actual — silently wrong** |

So the loud case was never the risk. Whoever merges second regenerates the baseline from the merged tree and commits it; from this PR on, forgetting is an exit 2 rather than a number nobody re-derives.

---

PR #919, opened 2026-08-19
V1-51: the playoff bracket comes from the league, not from a constant — four call sites simulated a 6-seed bracket for a 7-seed league

jasonleetucker-code:
## What this is

Lane 3, **V1-51**: one canonical playoff-probability engine — plus two live defects on the public `/league` surface that had to be fixed first.

The owner direction is what the work turned on:

> Intentional v1/v2 duplication is not automatically equivalent to one canonical production owner.

Applying that test found something better than duplication. **The simulators did not disagree about the answer. They disagreed about the question.**

Full record: [`docs/playoffs/V1_51_CANONICAL_PLAYOFF_BRACKET.md`](docs/playoffs/V1_51_CANONICAL_PLAYOFF_BRACKET.md).

## The defect, measured

Fetched 2026-08-19, `GET https://api.sleeper.app/v1/league/1312006700437352448`:

```
playoff_teams: 7    playoff_seed_type: 1
playoff_type:  1    playoff_week_start: 15
num_teams:     12
```

**The league takes seven.** Four call sites said six:

| call site | reads `settings.playoff_teams`? |
|---|---|
| `public_league/playoff_odds.py` (public `/league`) | yes — else silently `DEFAULT_PLAYOFF_SPOTS = 6` |
| `ros/playoff_sim.py::simulate_playoff_odds` | **no** — `playoff_seeds: int = 6` |
| `ros/championship.py::simulate_championship_odds` | **no** — `playoff_seeds: int = 6` |
| `ros/playoff_sim.py::simulate_trade_impact` | **no** — `playoff_seeds: int = 6` |

`ros/scrape.py` is the only production caller of the private engines and passes no bracket, so **both private simulators have been running a six-seed bracket for a league that takes seven.** On a 12-team league that is exactly the difference between sixth and seventh place mattering, and the championship simulation is more sensitive still because the bye count decides who skips a round.

Each default was invisible on its own: the parameter looked configurable and nothing configured it.

The pre-existing audit capture `docs/master-site-audit/evidence/W30/playoff-odds-two-engines.json` records the same finding from outside — `playoffSpots: 7` (public) against `playoffSeeds: 6` (private), 12 owners against 8.

## The fix

`src/public_league/playoff_structure.py` resolves the bracket **once**, from the league's own settings; all four sites consume it.

**Byes are derived, not configured** — `next_power_of_two(teams) − teams`. That reproduces the pair the code already hardcoded (6 → 2), which is the evidence it generalises the existing behaviour rather than inventing methodology, and gives 7 → 1, which no constant could have produced. Sleeper's `playoff_type` changes how rounds are *scheduled*, not how many teams qualify, so it is deliberately not modelled.

**An explicit bracket still wins.** `simulate_trade_impact` pins both arms to one; two arms on different brackets would measure the bracket, not the trade. The shared-seed argument in its own docstring applies to the league's rules as much as to the RNG.

### Missing is never six

A league that does not publish `playoff_teams` gets a refusal naming the reason — `playoffSpots: null`, every owner's probability `null`, an `unsimulable` block **carrying the full envelope** (a stub would `KeyError` any consumer reading `methodology`, on exactly the path that most needs to explain itself). The owners are still listed; only the certainty is withheld, the same posture the module already used for a preseason league.

`DEFAULT_PLAYOFF_SPOTS` is **deleted**, not deprecated — it was wrong for the league it served, and a plausible default left in scope is how a guess gets re-adopted. Every payload now stamps `playoffStructure`, so "which bracket produced these odds" is never implicit.

### The UI half

The chart short-circuited on all-null probabilities and rendered *"Preseason — odds not yet simulated"*. That is a promise the numbers arrive on their own; an unknown bracket is not, so telling a manager to wait for Week 1 would be telling them to wait for something that is not coming. The panel now names the real state, shows the backend's reason, and stops printing "top ? make playoffs" when it does not know the N.

## Two prerequisite defects on the public surface

Cherry-picked clean onto current `main` — both are on the same public engines and were blocking V1-51/V1-52.

**1. Playoff odds published alphabetical order as certainty.** `playoff_odds.py` substituted a flat `[100.0]` pool for owners with no sampled scores. Three individually-defensible decisions composed into a fabricated answer: the placeholder fired for *every* owner or none; that made each matchup an exact tie, so every simulation ended identically; and `_standings_from_sim` then broke the tie on its third key — the ownerId **string**, justified on the grounds that a crude tiebreak "doesn't matter when integrated over many draws", which the placeholder destroys. Measured on the committed production artifact: the seven `1.0`s are **exactly the lexically-first seven Sleeper user ids**.

**2. Roster health was double-counted** in `power_v2` — once as its own 0.03 weight and again inside `team_ros_strength`. Folded in (0.38 → 0.41) and `_load_roster_health_scores` deleted.

## What this does NOT do

It does not merge the two simulation cores. They keep genuinely different outputs and evidence, and `playoff_sim`'s docstring claim that "outputs match the v1 schema so the frontend can swap data sources without a contract fork" is **false today** — `numSims` vs `n_simulations`, `playoffSpots` vs `playoffSeeds`, `playoffProbability` vs `playoffOdds`.

Merging them needs the public/private lens decision applied here explicitly, and it is a larger unit. What this establishes is the **precondition**: a shared core cannot be built on top of two different brackets.

## Tests

`tests/public_league/test_playoff_structure.py` — the live 7/1 case, the bye derivation across nine bracket sizes, seven ways a bracket can be unknown each with its reason code, all four call sites taking the league's bracket, the explicit override surviving, and the agreement property across four bracket sizes.

Plus a **structural AST guard** that walks every function in `src/` and `scripts/` and fails on any integer default for `playoff_seeds` / `bye_seeds`. Written as AST rather than a text search because the first version matched the new module's own prose describing the defect.

Mutation-checked in four places: restoring `= 6` in the public engine, the playoff engine or the championship engine each turns Python tests RED, and collapsing the two UI states back into one turns two of three vitest tests RED.

Fixtures that declared no playoff format now declare one — a test league that plays no defined playoff is not a league — and the synthetic harnesses pin their bracket explicitly. One test was passing for the *wrong reason* after the change (the bracket refusal fired ahead of the no-distributions refusal it is named for); that is corrected and the refusal envelope has its own test.

## Verification

| gate | result |
|---|---|
| `pytest tests/ -m "not livedata"` | **8,221 passed, 0 failed** |
| `vitest run` (full frontend suite) | **2,058 passed** |
| `scripts/check_decision_coercions.py` | clean — one baseline entry deleted because the roster-health fix retired it (674 → 673) |

**Not deployed, and production verification has not been done.** The bracket change moves live playoff and championship numbers for this league the moment it ships — that is the point of it, and it should be observed rather than assumed.

---


jasonleetucker-code:
## V1-51 verification pack — and a blocking defect this PR's own claim hides

Head `127927c`, `Validate PR` ✅, 0 conflicts against `main`. No commits in this pass; this is evidence and a defect report.

---

# 1. BLOCKING — the championship bracket does not simulate the league's bracket

This PR says all four call sites take the league's bracket. Three do. **`ros/championship.py` stamps the right numbers and then simulates something else**, and I found it while building the verification pack rather than by reading the diff.

### The code

`src/ros/championship.py:81`, inside `_simulate_bracket`:

```python
playoff_field_size = len([s for s in seeded_owners if seeded_owners.index(s) < 6]) or 6
```

That is `min(len(seeded_owners), 6)`. And `_simulate_bracket` is called at `:258-263` with `bye_seeds=bye_seeds` only — **`playoff_seeds` never reaches it**. So the payload carries `playoffSeeds: 7`, `byeSeeds: 1`, `playoffStructure.source: "league_settings"` — all correct — while the bracket plays six.

### What that does, measured

20,000 sims, 7 seeded owners, `bye_seeds=1`, **identical** distributions for all seven (`_TeamDist(mean=100.0, sd=15.0)`), so any asymmetry is structural:

| seed | championships | share |
|---|---|---|
| seed1 | 9,972 | **49.86%** |
| seed2 | 4,961 | 24.80% |
| seed3 | 2,533 | 12.66% |
| seed4 | 2,534 | 12.67% |
| seed5 | 0 | **0.00%** |
| seed6 | 0 | **0.00%** |
| seed7 | 0 | **0.00%** |

`seed7` finish distribution: `{7: 20000}` — eliminated before a game is played, every time.

**Control** (same helper, 6 owners / 2 byes — the shape the hardcode was written for): `seed6` wins 12.43%. So the fixture is sound; the 7-team case is not.

### Why 5 and 6 zero out too — it is not just the field cap

Tracing the live 7/1 shape through `:95-126`:

```
field (capped at 6)   : [seed1 … seed6]
byes                  : [seed1]
wildcard              : [seed2 … seed6]   len=5, ODD
round-1 pairings      : (seed2 v seed6), (seed3 v seed5)
STRANDED in wildcard  : [seed4]           <- never plays, never placed
advancing to semis    : [seed1, w(2v6), w(3v5)]   len=3
semis -> semis_advance: len=1, leftover [w(2v6)]
FINAL PLAYED?         : False
```

`while len(wildcard) >= 2` strands the odd team. Semis reduce 3 → 1, so `if len(semis_advance) >= 2` at `:118` is **False and the finals block never runs**. No champion is assigned by the bracket at all — the champion falls out of the **defensive fallback** at `:127-135`, which hands finish `1` to the first unplaced owner in seed order. That is why seed1 takes ~50%.

**So for this league the championship odds are an artifact of list ordering, not a simulation.** That is the same defect class this PR fixed in `playoff_odds.py` — a placeholder making the ordering decide — one module over.

### Why the tests did not catch it

- `test_playoff_structure.py:198` asserts the stamped `(playoffSeeds, byeSeeds) == (7, 1)`. It never asserts a 7-seed can reach a final.
- The structural guard at `:237` is an **AST scan of keyword/positional integer defaults**. A literal `6` inside a function body is invisible to it — which is exactly where this one lives.

### Fix

Thread the bracket in, and handle the odd wildcard:

```python
def _simulate_bracket(seeded_owners, distributions, *, playoff_seeds=None, bye_seeds, rng):
    field_size = min(len(seeded_owners), int(playoff_seeds or len(seeded_owners)))
```
plus a bye/pairing rule that cannot strand a team (`byes_for_teams` already guarantees `field - byes` is even — the stranding only happens because the field was capped to 6 while byes stayed 1).

I have not pushed this: #919 is an integration surface and the brief says no new PRs. **Say the word and I will push it with a RED-first test that fails on the current code** — asserting a 7-seed's championship probability is > 0 and that the finals block runs.

---

# 2. Post-deploy verification pack

Envelope is `{contractVersion, league, section, data}` — every jq path is `.data.*`. All three sections are public (no session needed).

```bash
BASE=https://<host>
for S in playoffOdds rosPlayoffOdds rosChampionship; do
  curl -sf "$BASE/api/public/league/$S" -o /tmp/$S.json
done
```

### 2a. Live bracket = 7, derived byes = 1

```bash
jq '.data.playoffStructure' /tmp/rosPlayoffOdds.json
# expect: {"playoffTeams":7,"byeTeams":1,"playoffWeekStart":15,
#          "source":"league_settings","reason":""}
jq '.data.playoffSeeds, .data.byeSeeds' /tmp/rosPlayoffOdds.json      # 7, 1
jq '.data.playoffSeeds, .data.byeSeeds' /tmp/rosChampionship.json     # 7, 1
jq '.data.playoffSpots' /tmp/playoffOdds.json                         # 7
```

`byes_for_teams(7) == 1` is `next_power_of_two(7) - 7`; the same derivation reproduces the retired 6 → 2 hardcode, which is the evidence it generalises rather than invents.

### 2b. **Cache hazard — read this before trusting a green**

`data/ros/sims/latest_playoff.json` and `latest_championship.json` are committed **pre-V1-51** artifacts (`playoffSeeds: 6`, `byeSeeds: 2`, no `playoffStructure`) and are served under a **6h TTL**. Immediately post-deploy both ROS endpoints can legitimately return `6` until the scheduled scrape rewrites them.

```bash
jq '{structure: .data.playoffStructure, cached: .data.cached}' /tmp/rosPlayoffOdds.json
```
**Both** must hold: `playoffStructure != null` **and** `cached == false`. Asserting only on `playoffSeeds` proves nothing — a stale artifact and a broken deploy look identical.

### 2c. Missing bracket refuses, never six

Against a league whose settings omit `playoff_teams`:

```bash
jq '.data | {spots: .playoffSpots, reason: .unsimulable.reason,
             owners: (.owners|length),
             probs: [.owners[].playoffProbability] | unique}' /tmp/playoffOdds.json
# expect spots: null, reason: "league_settings_omit_playoff_teams",
#        owners: 12, probs: [null]
```

Owners stay listed; only the certainty is withheld. Reason codes are the four in `playoff_structure.py:68-72`; `_int_or_none` rejects `bool`, so `playoff_teams: true` is UNKNOWN rather than 1. Out-of-range (2..32) refuses rather than clamps.

### 2d. Per-team before/after

The **before** side is already committed production output — `docs/master-site-audit/evidence/W30/playoff-odds-two-engines.json`: 12 real owners, `v1.playoffSpots: 7` against `v2.playoffSeeds: 6` / `byeSeeds: 2`, and `v2.ownerCount: 8` (the private engine covered only 8 of 12). Its `perOwner[]` carries `v1_playoffProbability`, `v2_playoffOdds` (**null for 4 owners**) and `delta`.

Capture the after side in the same shape and diff row-by-row:

```bash
jq -n --slurpfile pub /tmp/playoffOdds.json --slurpfile ros /tmp/rosPlayoffOdds.json '
  ($ros[0].data.playoffOdds // []) as $r |
  [$pub[0].data.owners[] | {ownerId, displayName,
     after_public: .playoffProbability,
     after_ros: ($r[] | select(.ownerId == .ownerId) | .playoffOdds)}]' \
  > /tmp/after.json
```

The deliverable is a 12-row table with a `null → value` column for the four owners the old private engine never covered. **Championship deltas should not be published until §1 is fixed** — they are currently an ordering artifact.

---

# 3. What V1-51 still requires — do not claim "one playoff engine"

Bracket **ownership** is unified. The **engine** is not, and two more things are open beyond §1:

**(a) The lexical tiebreak survives in both ROS engines.** `playoff_sim.py:838-841` and `championship.py:254-257` sort on `(-wins, -pf)` over an alphabetically-sorted owner list; Python's sort is stable, so ties resolve to ownerId order with no jitter and no third key. That is the same defect W19-F008 fixed in the public engine, and `test_playoff_odds_no_fabricated_certainty.py:214`'s AST guard covers `playoff_odds.py` only.

**(b) The two cores answer different questions and cannot be swapped.** The divergence is wider than this PR states:

| concept | `public_league/playoff_odds.py` | `ros/playoff_sim.py` |
|---|---|---|
| sim count | `numSims` | `n_simulations` (and `nSimulations` in `simulate_trade_impact` — a third spelling) |
| bracket size | `playoffSpots` | `playoffSeeds` |
| per-row probability | `playoffProbability` | `playoffOdds` |
| **row array key** | `owners` | `playoffOdds` — array and field share a name |
| byes | absent | `byeSeeds` |
| ran-or-not | `simulated` | `converged` — different semantics, not a rename |
| error bars | absent | `worstPlayoffOddsSe`, `oddsSeTolerance`, per-row CIs |
| per-row state | `currentWins`, `currentPointsFor` | `expectedWins`, `medianFinalSeed`, `seedDistribution` |
| sim budget | fixed 10,000 | adaptive 2,000–60,000 to an SE tolerance |

So `playoff_sim.py`'s docstring claim that outputs match the v1 schema "so the frontend can swap data sources without a contract fork" remains **false**. Merging them needs the public/private lens decision applied explicitly; V1-51 established the precondition (one bracket) and no more.

**(c) Two smaller gaps.** `playoffStructure` is absent from the **success** returns of `compute_playoff_odds` and `simulate_trade_impact` (present only on their refusals), so a green payload cannot prove which bracket produced it. And `playoff_odds.py:554-563` returns `playoffSpots: 0` rather than `null` with no current season, making `REASON_NO_SEASON` unreachable through that engine.

---

## Reproduce §1

```bash
python3 - <<'PY'
import random, collections, sys; sys.path.insert(0,".")
from src.ros.championship import _simulate_bracket
from src.ros.playoff_sim import _TeamDist
o=[f"seed{i}" for i in range(1,8)]
d={x:_TeamDist(owner_id=x,mean=100.0,sd=15.0,pf_to_date=1000.0) for x in o}
c=collections.Counter(); r=random.Random(7)
for _ in range(20000):
    for k,v in _simulate_bracket(o,d,bye_seeds=1,rng=r).items():
        if v==1: c[k]+=1
print({x:c[x] for x in o})
PY
```

---

jasonleetucker-code:
## Production before-state captured — the defect is live, and two surfaces on one page disagree

Captured from the currently deployed build before any of this wave merges, so the after-comparison has a real baseline rather than a reconstructed one.

```
/api/public/league/playoffOdds      playoffSpots : 7    <- reads settings.playoff_teams
/api/public/league/rosPlayoffOdds   playoffSeeds : 6    <- hardcoded default
/api/public/league/rosChampionship  playoffSeeds : 6    <- hardcoded default
```

That is the whole finding in three lines. The public v1 section reads the league's real setting and says **seven** teams make the playoffs; the two ROS Monte Carlo sections shipped from `src/ros/playoff_sim.py` and `src/ros/championship.py` say **six** — and all three render on `/league`. `byeSeeds: 2` on both ROS sections, which is the bye structure that follows from the wrong seed count.

So this is not a latent default waiting for an unusual league. It is wrong right now, for this league, on a public page, and the seventh seed's entire playoff probability is being simulated away.

Your §"The defect, measured" table is confirmed exactly as written, including that the parameter "looked configurable and nothing configured it" — `playoffOdds` proves the setting was readable all along.

Per-owner probabilities from all three sections are stored alongside the seed counts, so after deploy I can diff the actual distribution rather than just the header numbers. I expect the seventh seed's odds to become non-zero and the bye structure to change; if `byeSeeds` stays 2 under a 7-team bracket that is worth a second look, since a 7-team bracket conventionally gives one bye.

**Integration position.** #920 is validating now and merges first (it must precede #915's contract edit). You are next. Base drift against your branch is 22 commits and **0 class-B** — every one is a generated data refresh, so there is no source-level reconciliation to do, only data-sensitive revalidation.

Two things I will handle at merge, flagged so they are not a surprise:

- You touch `config/coercion_baseline.json`, and you are the **first of five** open PRs that do. I regenerate it from the final merged tree and require `declared == actual` — not from either lane's arithmetic. Measured on real branches earlier in this wave, two lanes retiring different entries merged with **no Git conflict** and left the file declaring 673 while carrying 670.
- #923 **contains** this branch, so it is a stack rather than an independent PR touching the same 16 files. Its diff collapses to its own additions once you land, and it merges straight after on current-head CI.

---

PR #921, opened 2026-08-19
#804 capture: per-source rank observations in a new source_rank history lane

jasonleetucker-code:
# POST-V1 capture infrastructure — **changes the V1 percentage by zero**

@Claude 5 — this is a new unit on a fresh branch off main. **I am not merging it.** It is capture only: nothing reads the lane, and it moves no published number.

## Why now, and why it can't wait

#804 is about **correlated / shared-lineage source movement** — injected correlated anomalies moved the blend by as much as **~48%** while existing defenses caught nothing, because sources agreeing on something *wrong* is indistinguishable from agreement at the blend.

Telling structural dependence from noise needs the observations **over time**, and that series does not exist:

| | |
|---|---|
| per-source rank maps published per build | 22–24 |
| preserved in any archive | **3** |
| bundles carrying more than 3 | **0 of 176** |

Every unrecorded build is evidence no later effort can reconstruct. That is the entire justification for capturing before the analysis is designed.

## Scope discipline

**Not in this PR, and unauthorized:** correlation methodology, source-family assignment, family weighting, automatic downweighting, confidence changes, B10 logic, canonical value or ranking movement. No UI. No Central Buy/Sell. #911 untouched.

## No parallel store

Extends the canonical owner (`src/history`) with a deliberately named lane and **inherits** its guarantees rather than restating them: append-only (`INSERT OR IGNORE`; a conflicting re-record is *surfaced*, never applied), idempotent, `HISTORY_FLOOR` enforced, never-future selection (`asof` was already lane-parameterised), tz-aware `observed_at`.

The instant/zone rule was **extracted** from `record.py` into `contract_observed_instant`, so both lanes read the producer's clock through one definition instead of two copies.

**Not `source_value`.** A rank is an ordering position; a value is a price. No shared units, not convertible, and they disagree about what "missing" means. `_validate_source_rank` refuses a rank observation carrying a value at all.

**Missing is absence, never zero.** A source that didn't rank an asset contributes *no row* — not rank 0, not null, not a sentinel; all three refused, because rank 0 sorts **first** on every board. Booleans too (`True` is an `int` and would store as rank 1).

## Provenance for the question this exists to enable

Four nullable columns, unused by every other lane: `raw_rank` (what the source **published** before ladder translation — a rookie board's #36 becomes #247 on the overall ladder; different facts), `rank_method`, `rank_pool` (ranks from different coordinate pools are not comparable and must not be correlated), and **`shared_market_translated`**.

That last one is the most important field here: a source projected onto another market's backbone is correlated with it **by construction**, and an analysis that missed it would "discover" a correlation the pipeline itself created. Measured on a real board: **638 of 7,245** observations carry it.

The new columns fold into `content_hash` **only when present**, so every row from every pre-existing lane hashes to exactly the byte it did before. Without that, each stored row would re-ingest as a **conflict** instead of a duplicate and the store's idempotency — which backfill and the migrations are built on — would break on contact. Pinned by test.

## Non-influence — proven, not asserted

Same pinned input export, **code as the only variable**, `golden_board` on `origin/main` vs this branch:

```
rows 1108 -> 1108 · ranked 740 · priced 849 · picks 162 · idp 396
VALUES: 0 moved, 0 newly priced, 0 newly unpriced
RANKS:  0 changed
ASSERTION OK: no value changed.
```

The two captures are **byte-identical** — `sha256 c1b9cde42f30021f` on both sides.

Reinforced structurally: no valuation/weighting module imports the lane (asserted as an **import edge**, not a substring — `source_ranks` is a long-standing local variable in `data_contract`), the only production importer is the fresh-scrape call site, the capture module imports no pricing module, and building observations does not mutate the contract.

## Storage — measured *before* proposing production

| | |
|---|---|
| observations/build | **7,245** across 21 sources |
| bytes/observation | **446 B** |
| rows/day (2-hourly) | 86,940 |
| **bytes/day** | **38.8 MB** |
| 30 / 90 days | ~1.16 GB / ~3.49 GB |
| **1 year** | **~14.2 GB** (31.7 M rows) |

The cost is **index overhead on the shared table, not payload** — stripping every denormalized string saves only 4–7%. **Cadence is the real lever:** daily capture is ~1.2 GB/yr, and correlation between sources is a question about *days*, not hours.

**So the flag ships OFF** (`RISKIT_FEATURE_SOURCE_RANK_CAPTURE`). 14 GB/year is a production disk commitment an owner makes deliberately, not one a merge makes for them. The retention concern is named in the doc; **no destructive retention policy is proposed**, since inventing one is unauthorized. Off, nothing is written and no response body differs.

⚠️ **Decision needed from you:** enable at which cadence? The per-cadence table is in `docs/history/SOURCE_RANK_CAPTURE.md` §7. Until it's on, the evidence keeps perishing — that's the trade-off, stated rather than resolved unilaterally.

## Tests

`tests/history/test_source_rank_capture.py` — 33 covering: append-only (incl. a conflicting re-record leaving the stored row untouched), idempotent rebuild, tz-aware/naive/date-only instants, never-future selection, `HISTORY_FLOOR`, missing-rank ≠ zero across four shapes, source identity + lineage, the v1→v2 column upgrade without row loss, legacy hash stability, and the non-influence import edges.

## Evidence at head `03ee6212d`

```
tests/ -m "not livedata"     8,398 passed / 51 skipped
ruff check + format          clean       coercion gate     clean
governance / planning / audit-status / flag-docs   clean
release-candidate            "the validated tree IS the tree that will merge"
sync                         0 behind origin/main
```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ snapshots). This line said "Private repo" while `SECURITY.md` and
   Sleeper-login allowlist; public routes hit `/api/public/league/*`.
 - **Feature flags** (`src/api/feature_flags.py`): every new
   capability from the 2026-04 upgrade ships flag-gated. Defaults are
-  **per-flag**: 8 of the 16 in `_DEFAULTS` ship enabled (`bdvm_engine`,
+  **per-flag**: 8 of the 17 in `_DEFAULTS` ship enabled (`bdvm_engine`,
   `te_basis_conversion`, `monte_carlo_trade`, `idp_scoring_fit`,
   `reception_scoring_fit`, `nfl_data_ingest`, `realized_points_api`,
   `perfect_draft`),
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -26,7 +26,7 @@ Browser ─► Nginx ─► Next.js (port 3000) ─► FastAPI (port 8000) ─�
    `src/api/feature_flags.py`. Env override via `RISKIT_FEATURE_<NAME>=1`
    (or `=0` to disable).
 
-   **Defaults are per-flag.** As of 2026-08-05, 8 of the 16 entries in
+   **Defaults are per-flag.** As of 2026-08-05, 8 of the 17 entries in
    `_DEFAULTS` ship enabled — `bdvm_engine`, `te_basis_conversion`
    (which reprices every tight end on the live board),
    `monte_carlo_trade`, `idp_scoring_fit`, `reception_scoring_fit`,
--- a/docs/history/SOURCE_RANK_CAPTURE.md
+++ b/docs/history/SOURCE_RANK_CAPTURE.md
@@ -1,0 +1,175 @@
+# #804 per-source rank capture (`source_rank` lane)
+
+**Owner module:** `src/history/source_rank.py` · **Lane:** `store.LANE_SOURCE_RANK`
+**Flag:** `RISKIT_FEATURE_SOURCE_RANK_CAPTURE` — **default OFF**
+**Status:** POST-V1 capture infrastructure. **Changes the V1 percentage by zero.**
+
+---
+
+## 1. What this is, and firmly what it is not
+
+**Capture only.** It records what each source ranked each asset, on each build,
+into the canonical temporal ledger. Nothing reads the lane.
+
+Explicitly **not** in this unit, and unauthorized until separately approved:
+correlation methodology, source-family assignment, family weighting, automatic
+downweighting, confidence changes, B10 logic, and any canonical value or
+ranking movement.
+
+## 2. Why the evidence has to be captured now
+
+#804 is about **correlated / shared-lineage source movement**: injected
+correlated anomalies moved the blend by as much as **~48%** while existing
+defenses caught nothing, because sources agreeing on something wrong is
+indistinguishable from agreement at the point of the blend.
+
+Deciding whether two sources are genuinely independent needs their observations
+**over time** — a correlation that holds every day is structural dependence;
+one that appears on a single board is noise. That series does not exist:
+
+| | |
+|---|---|
+| per-source rank maps published per build | **22–24** |
+| source boards preserved in any archive | **3** |
+| archive bundles carrying more than 3 | **0 of 176** |
+
+Every build that goes unrecorded is a day of evidence no later effort can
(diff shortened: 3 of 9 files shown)


jasonleetucker-code:
**PR Validation red on this branch is main-drift, not this PR's content — and it is not fixable inside the freeze.**

Run [32181019487](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32181019487) fails at the **decision-path coercion gate**, not at any test:

```
decision-path coercions: 679 present, 675 accepted as debt
STALE BASELINE ENTRIES (1) — fixed by this change; delete them from the baseline:
  src/sharp/consensus.py::quality = float(person["quality"] or 1.0)
```

That coercion was removed by **#911**, which has since merged. `main`'s `config/coercion_baseline.json` no longer carries the entry; this branch is 18 commits behind and still does, so the gate correctly reports the baseline as stale *for this branch's tree*.

Two things follow:

1. **The merged tree is unaffected.** `main` removed the line and this branch never touched it, so the merge takes the removal and the gate passes on the result. This is red on the PR's own validation run only.
2. **Everything downstream of it was skipped, not failed.** The `##[error]` at the end of the log is the advisory `check_source_health` step (`continue-on-error: true`) — `idpTradeCalc (picks)` byte-identical for 34 days, an upstream vendor condition with 0 contract source-health errors. Non-blocking by design, and unrelated.

**Not fixing it here** because the instruction is to freeze this PR and return to V1 correctness work; a `git merge origin/main` clears it in one commit if Integration wants a green run before merging. Its production flag stays **OFF** pending the cadence decision — which I have now written up as a standalone owner decision in `docs/lane4/LANE4_804_CAPTURE_CADENCE_DECISION.md` (shipped in **#927**), rather than defaulted.

---

jasonleetucker-code:
Deferred/reference preserved. Per-source rank history is explicitly post-launch evidence-capture work. Closing the old PR so it does not masquerade as active merge work; retain branch/diff for later re-derivation.

jasonleetucker-code:
Closing as **deferred/reference preserved**. Per-source rank-history capture is explicitly post-launch evidence work. The old branch remains useful reference, but current main should receive a fresh implementation when that capture lane is authorized. Canonical disposition: `docs/OPEN_PR_DISPOSITION_LEDGER_2026-09-09.md`.

PR #922, opened 2026-08-19
STACKED ON #914 — Roster Intelligence hardening: integration's eight non-blocking findings

jasonleetucker-code:
> **DEPENDENT ON #914. Base is `claude/v1-roster-intelligence-ovmwco`, not `main`.**
> Branched from #914's head `330ed5c`. **#914 is frozen and unchanged** — not one commit
> was added to it. Merge #914 first; this rebases onto `main` afterwards, or re-targets
> once #914 lands. Claude 5 remains merge authority for both.

Closes the eight Roster-owned non-blocking findings from the integration review. Every
one is correctness or canonical ownership; none is cosmetic cleanup.

| # | finding | repair |
|---|---|---|
| F2/F3 | `reserve_demand` re-derived slot demand, needing a private flex table and two private cross-module imports to agree with the owner | consumes `lineup.slot_demand`; `_RESERVE_FLEX_SLOTS` **deleted**; `ReserveDemand` publishes `flex_slots` / `dedicated_basis` |
| F1 | the league's configured flex eligibility reached no solve | `lineup.configured_slot_eligibility` (rule) + `data_contract.contract_slot_eligibility` (plumbing), threaded into the lineup stamp **and** the endpoint |
| F4 | two position vocabularies — **and six spellings eligible for nothing** | one vocabulary in the chain; eligibility decided on token **or** family |
| F5 | `PositionNeed` was two types in one package | different questions → renamed `engine.PositionDeficit`, **not** merged |
| F6 | the board join keyed by `canonicalName` while the pool keyed by either name | keyed the way the pool keys, one row per player |
| F7 | ~10 sites raised `TypeError` on `ros_value is None` | `lineup.is_priced` / `priced_players`; exclude and report, never coerce |
| F8 | droppability counted the live auction's own rookies as free agents | default resolves them from the contract; `waiverPopulation` stamps it |

## Three things worth reading past the table

**F4 was bigger than the finding said.** Chasing the MLB grouping mismatch turned up that
`NT`, `OLB`, `ILB`, `MLB`, `FS` and `SS` all resolved to a *correct family* and then
matched **no slot at all** — eligibility was tested on the raw token against the
Sleeper-faithful table while `lineup_position` had been deliberately widened for scraper
spellings. **A genuine linebacker listed as `MLB` could not start anywhere.**

Fixed at the rule rather than by widening a second table: eligibility is the raw token
**or** its resolved family, because those are two spellings of one fact. Strictly a
widening — an `or` cannot make an eligible player ineligible — and the delta is frozen
literally in a test:

```
GAINED  PK/P → K · NT → DL, IDP_FLEX · OLB/ILB/MLB → LB, IDP_FLEX · FS/SS → DB, IDP_FLEX
LOST    none
```

**F8's direction is the point.** Counting the auction's own rookies as free agents
*raises* the replacement bar, so every cut looks **cheaper** — the defect
`src/draft/rookie_pool.py` exists to prevent, arriving by a different door. Measured on
the live board: `WR 2036.0 → 1519.0 (−25.4%)`, `DL −8.1%`, `TE −7.1%`. On defaults the
two surfaces now agree **12/12 on both `waiverValues` and `cutLadder`**, where previously
that required the caller to know to pass the auction keys.

**F5 was deliberately not consolidated.** The brief's test is whether two things answer
the same question. A rung ladder against `k × teamCount` and a distance below replacement
do not — so they got two names, not one implementation.

## The pipeline is proved, not described

180 seeded rosters × 5 league shapes × 4 eligibility configurations — including a
**non-laminar** flex set (`REC_FLEX` beside `FLEX`) and ~12% unpriced players — over the
real chain:

```
actual lineup config → solve starters → fill FLEX/SF/IDP_FLEX
  → remove starters → reserve demand → meaningful core → Team Strength / Weakness
```

**No player counted twice:** core membership unique; starters and reserves disjoint and
exhausting the core; a seated player never also reported unpriced; the core never exceeds
the declared ceiling; Team Strength sums each member once and its position groups re-sum
to the total; no weakness rung credits one player to two positions. The **Superflex fold**
gets its own case, since it is the one place a QB could be double-counted by *arithmetic*
rather than by selection.

**No bypass path:** one lineup solver, `math.ceil` in exactly one module, no private
top-N anywhere in the chain, and every chain output takes a `MeaningfulCore` as its first
parameter — so none can be handed a population selected some other way.

**Both double-count guards are mutation-proven.** Letting the reserve pass see the full
pool turns 3 red; making Team Strength count reserves twice turns 1 red. (My first
attempt at the second mutation was a silent no-op against a string that does not exist in
the file — re-run against the real summation site, and reported here because a mutation
that does not apply proves nothing.)

## Boundaries held

- **`C3-CAP-01` stays trade-owned.** Nothing here touches capacity, forced-drop
  accounting, verdicts, VA or package construction. §14 records the ruling and names what
  this lane publishes for #913 to consume.
- **`marginal` / `profiles` / `targets` keep the replacement vocabulary.** They are the
  older WS-J engine behind `/api/gameplan` and group for replacement-level purposes;
  re-grouping them would move a different owner's live numbers. The vocabulary guard
  **names its exclusions** rather than quietly covering seven of ten modules.
- **Two published numbers owned elsewhere keep their conventions.** `engine`'s unpriced
  counter keeps its pre-existing conflation of a real `0.0` with an unpriced player (its
  own `_unpriced_note` documents it), and `marginal`'s mean-entrant value keeps its
  `0.0`-when-empty rule. A hardening unit gets to stop them crashing, not to redefine
  them.
- **No V1 completion percentage or governance record touched.**
- **One frontend line** — `starter-slots.js`'s `LB_FAMILY` gains `MLB`, required by
  `tests/lineup/test_single_owner.py` the moment Python moved. A lockstep repair, not
  presentation work.

## Evidence — head `ae58371`

| Gate | Result |
|---|---|
| Full `pytest -m "not livedata"` | **8645 passed, 51 skipped, 0 failed** (16m45s) |
| `ruff format --check .` / `ruff check .` (pinned 0.6.9) | 1195 formatted / all passed |
| Decision-coercion gate | clean |
| `optimalLineup` on the live board | **byte-identical throughout**, hash `a311593a9b7602ac` |
| Droppability defaults vs the draft surface | **12/12 identical**, `waiverValues` and `cutLadder` |

**RED-first throughout:** 6 of 15 (F1–F3), 5 of 10 (F4), 5 of 8 (F7), 7 of 9 (F6), 3 of 6
(F8) fail against the pre-fix tree. Every latent defect closed here is latent for a
measured reason — 0 of 660 rostered players carry `MLB`/`P`/`PK`, 0 of 177 archives carry
a duplicate `ownerId` — which is exactly why none of them was caught by the live board.

## What this does not claim

Not production-verified. All evidence is against the tracked export archive —
production-equivalent, not production. The live-board checks above are re-runs of the same
contract build, not observations of the deployed stack.

https://claude.ai/code/session_01PyqbZejxh7ytbfijL5pmdE

---


jasonleetucker-code:
## Re-based on #914's new head — `fd70515`

#914 moved from `330ed5c` to `ab8f9f1` (integration merged `main` in and cleared both
blockers). This branch has taken that in, so its base is current and its diff is still
only this unit.

| Gate | Result on `fd70515` |
|---|---|
| Full `pytest -m "not livedata"` | **8645 passed, 51 skipped, 0 failed** (17m00s) |
| Behind base (`claude/v1-roster-intelligence-ovmwco`) | **0** |
| `ruff format --check .` / `ruff check .` | 1195 formatted / all passed |
| Decision-coercion gate | clean, **671/671**, `--write-baseline` a byte-identical no-op |
| `optimalLineup` on the live board | unchanged, `a311593a9b7602ac` |

The baseline check is worth stating explicitly given #911/#917: the base merge brought
`docs/VERSION_1_COMPLETION_CONTRACT.md` and an audit doc across, and the coercion
baseline needed **no regeneration** — 671 declared, 671 actual, and re-writing it
produced a byte-identical file. Nothing drifted.

**Still nothing added to #914.** When it merges, this re-targets onto `main`; I'll do
that and refresh the evidence rather than leave the base pointing at a merged branch.

---

jasonleetucker-code:
## Roster-owned `slot_eligibility` defect repaired — head `a03d17532`

Integration found that F1's eligibility never reached the droppability chain. Confirmed, and it was worse than not-propagated: the parameter was accepted, **documented as applied, and discarded three lines later**.

```python
``slot_eligibility`` is accepted and, when supplied, applied by re-running
the owner against the caller's slots — it is not silently dropped.
...
del slot_eligibility  # the owner reads eligibility from the slot names
```

The comment justifying the `del` was also wrong: the owner reads `slot_eligibility` when given one and falls back to slot names only when it is `None`.

### The chain, all four hops

| hop | file | carried it? |
|---|---|---|
| `pool_cut_ladder` | `roster_intel/droppability.py` | accepted, then `del` |
| `build_cut_ladder` | `draft/displacement.py` | **no parameter** |
| `_filled_slot_count` | `draft/displacement.py` | **no parameter** |
| `solve_optimal_assignment` | `ros/lineup.py` | **already accepted it** |

The terminus supported it all along; two intermediate hops dropped it.

### What it would have done

Slots `[RB, FLEX]`, roster `RB1`/`RB2`/`WR1`, league FLEX admitting **only WR**: the WR is the only player who can fill FLEX, but scored against the default RB/WR/TE rule `RB2` appears to cover it — so the engine offered **releasing the one player holding the lineup together**. The ladder's entire purpose is to refuse cuts that break the lineup, so this is a safety property.

It survived F1's review because `pool_cut_ladder` has **no production and no test caller** — it is the entry point `C3-CAP-01` (#913) is built to consume. Same shape as F4's dead position spellings: the live league does not exercise the path.

### Repair

- **`draft/displacement.py`** — `_filled_slot_count` and `build_cut_ladder` gain and forward the parameter. **Plumbing only**: no eligibility table, no interpretation, so `src/ros/lineup.py` remains the single owner and no second owner is created.
- **`roster_intel/droppability.py`** — the `del` is gone, the value is passed through, and the false docstring is replaced with what the code does.
- **`team_droppability`** now derives from the contract via `contract_slot_eligibility(contract) or None` — the same helper the lineup stamp and `/api/roster/intelligence` use. Without it the cut ladder would score a custom-flex league against built-in defaults while `optimalLineup` used the configured ones: two answers to "who can fill this slot" inside one league.

`or None` is load-bearing in both places — an empty map means *nothing configured*, not *nothing is eligible*. Passing `{}` through would empty every ladder.

### Evidence

| gate | result |
|---|---|
| **RED first** at unrepaired `fd70515` | **3 of 4 new tests fail**; the 4th is the behaviour-preservation test, which must pass on both sides and does |
| **Mutation proof** | re-introducing the discard turns the adversarial case **and** the structural guard red; reverting restores 23/23 |
| **Live-board inertness** | `league_droppability` over the real contract, before vs after: **12/12 teams byte-identical** on `cutLadder` + `waiverValues` — 95,414 bytes, sha256 `ee00e1f6e9f08410` both sides |
| `pytest tests/roster_intel tests/draft tests/lineup` | **687 passed**, 22 skipped |
| #922 hardening suite (adversarial, demand-chain, vocabulary, unpriced, identity) | **68 passed** |
| Double-count mutation 1 (reserve pass sees full pool) | **3 red** — still bites |
| Double-count mutation 2 (Team Strength counts reserves twice) | **1 red** in the adversarial suite + 4 in the strength suite — still bites |
| Full `pytest -m "not livedata"` | **8649 passed, 51 skipped, 0 failed** (15m37s, exit 0) |
| `ruff format --check` / `ruff check` | 1093 files formatted / all passed |
| Decision-coercion gate | **671 present / 671 accepted**, clean on changed files |

The live league declares no custom flex eligibility (`contract_slot_eligibility` → `{}`), so the derivation resolves to `None` and today's defaults are unchanged. **Provably free for the current leagues and correct for a configured one.**

### Two things for your review, flagged rather than buried

1. **`src/draft/displacement.py` belongs to the Perfect Draft lane.** The canonical path runs through `build_cut_ladder`, so the repair cannot avoid that file. The change is an optional keyword defaulting to today's behaviour and adds no eligibility logic — but it is a cross-lane edit and you should see it as one.
2. **The structural guard is an AST check, not a substring match.** My first version was a text search and failed on my own docstring, which now quotes the retired statement while explaining the defect. A text match would fire on that history note *and* miss a discard written any other way, so it asks the parser for a real `Delete` node instead.

### Status

Local gates are complete and green. **`Validate PR` is still running on `a03d17532`** — I'll act on it per the drive-to-green posture rather than leave a red head. **#914 is untouched**: head still `ab8f9f1`.

**#913 stays blocked from canonical Roster consumption until this lands** — this repair is precisely what unblocks it, since `pool_cut_ladder` is the entry point it is meant to call.

Handing back to you. Not merging.

---

PR #923, opened 2026-08-19
V1-52: two lenses of one power engine — and the offseason ranking that was publishing the alphabet

jasonleetucker-code:
## What this is

Lane 3, **V1-52**: one canonical weekly power-rankings engine.

**Stacked on #919** (base is `claude/v1-playoff-power-consolidation`, not `main`) — the roster-health de-double-count and the playoff-odds fix in that PR are prerequisites for this one. Merge #919 first; this PR's diff is the five commits above it.

Full record: [`docs/power/V1_52_CANONICAL_POWER_ENGINE.md`](docs/power/V1_52_CANONICAL_POWER_ENGINE.md).

> **Step 5 — the retirement itself — is scoped and measured but NOT done**, for a substantive reason given below. What lands here is the engine work that has to precede it, plus two live defects found while measuring.

## The reading that reframed the unit

The audit capture `evidence/W30/power-two-engines.json` is usually quoted for its rank shift (mean 2.8, max 7). Its `effectiveWeights` say more:

```json
"v2": { "preseason": true,
        "effectiveWeights": {"team_ros_strength": 0.38, "roster_health": 0.03} }
```

So the comparison is **one week of last season's scoring percentiles** against **pure roster strength**. The two engines were not disagreeing about the answer — they were answering different questions, and nothing on either surface said so.

Its `n` is worth noting too: **10 against 12**. `power.py`'s `currentRanking` is the last *single week*'s ranking, so any owner who did not play that week is absent from the published power ranking entirely.

### 1. An identifier ordering, published as a power ranking

Renormalisation had a floor nobody had put in. When **every** component is dropped, `active_weights` is empty, `weight_total` fell back to `1.0` against an empty numerator, every owner scored exactly `0.0`, and the sort — stable over equal keys — handed out ranks `1..N` in `owner_ids` order.

Reachable **structurally**, not rarely: results-only drops `team_ros_strength` by definition and preseason dropped all seven historical components, so the state was guaranteed for the whole offseason.

On the committed dev snapshot it was self-contradicting: five owners, every score `0.0`, and the published order **disagreed with the components printed beside it** — `owner-B` led `owner-A` on five of seven components and ranked below it.

Same family as the playoff-odds defect fixed in #919, where a placeholder made every matchup a tie and the `ownerId` string tiebreak became the answer. **The coercion baseline had already registered the mechanism as debt** — `weight_total = sum(...) or 1.0` — so this is a recorded hazard coming true, and the gate reports that entry retired.

Now: no surviving component, no ranking. `powerScore` and `rank` are `None` — never `0.0`, which is a score a team can earn, and never `1..N`. An `unrankable` block names the reason and the missing inputs; the owners and their raw components are still listed. Same posture `playoff_structure` takes for an unknown bracket. The UI renders it as a named state rather than the "not ready, the next scrape will populate this" copy, which would promise numbers that are not coming.

### 2. The retrospective lens was suppressing the results it is made of — my own defect from step 1

`_is_preseason` dropped all seven historical-results components, and its docstring gave the reason: they *"describe a finished year and don't project the upcoming one ... so the score reflects only forward-looking inputs"*. Correct for the **forward-looking** lens. Exactly backwards for the retrospective one, whose entire subject matter *is* the finished year.

Step 1 added the results-only lens and it inherited a suppression written for the other one. The consequence was total: results-only already drops `team_ros_strength`, so preseason left it with **nothing** — the every-component-missing state above — while the completed seasons it is made of sat in the accumulators untouched, for every day of the offseason.

Visible in one row on the dev snapshot: `owner-B`, ranked 2nd on an all-zero score under the defect, is now **1st at 86.61**. Forward-looking still refuses there, correctly — preseason with no team-strength file has genuinely nothing to look forward to.

## Do the two engines agree? Measured

| fixture | mean \|rank shift\| | max | note |
|---|---|---|---|
| committed dev snapshot (real) | **0.00** | 0 | identical order on the 4 common owners; v2 also covers a 5th that v1 drops |
| 12 owners, 6 weeks, monotone strength ladder | 0.33 | 1 | weak by construction — a ladder makes most formulas agree |
| adversarial: great scores, 0-6 record | 0.67 | 1 | v1 has **no** W/L or streak component; v2 weights them 0.10 + 0.05 |

The retrospective quantities substantially agree, and the large real-world divergence in the W30 capture is the **forward-looking input** — exactly what the lens distinction names. That is the evidence these are one engine's two lenses rather than two engines.

**Limitation, stated rather than glossed:** two of those three fixtures are synthetic and the third is a small dev snapshot. A production comparison needs the prod snapshot plus the ROS team-strength file, neither of which exists in this environment.

## Why step 5 is not a shim

`power.py` must stop being an engine. The obvious minimal move — make `build_section` an adapter that delegates and renames fields — **does not work**, recorded so it is not attempted again:

`power.jsx` renders `components.pointsPerGame`, `recentAvg` and `allPlayWinPctThisWeek` as **raw values**; the canonical engine publishes `ppg` and `recent` as **percentiles**. An adapter cannot recover a raw PPG from a percentile, so it would have to compute one — a second owner again, one layer down, which is the defect this unit exists to remove.

So the retirement is **one renderer, one engine, the toggle selecting a lens**: `ros-power.jsx` absorbs the trend chart, `power.jsx` and `power.py` retire together. That changes what the Power tab *displays* for users on the v1 path — percentile component bars instead of raw PPG columns — a deliberate, user-visible change that needs the production measurement above alongside it.

## Steps 1–2, for context (already on this branch)

Two lenses that are **not two formulas**: results-only is the same `WEIGHTS` vector with `team_ros_strength` declared missing and renormalised by the machinery that already handles an absent team-strength file, so every retrospective component keeps its weight *relative to the others*.

The per-week trend moved into the canonical engine and runs through the same `_score_state` as the headline — a second scoring implementation for the series is how a chart ends up being a different quantity from the number printed beside it. It is **results-only at every point, including the last**: `team_ros_strength` is a single current snapshot with no per-week history, so back-filling it is the as-of defect, and splicing it into only the final point is worse — the line would jump at the end for a reason unrelated to how the team played.

## A note on fixtures

`_make_snapshot` never populated `roster_to_owner`, which `luck._season_weekly_scores` needs to attribute a matchup row to an owner — it **skips** rows it cannot resolve, so zero weeks scored, every component fell back to its neutral default, and every owner scored an identical **33.64** across three weeks of deliberately different points.

Two of this unit's own lens tests were asserting shape while measuring nothing. Repaired at the shared helper with the reason recorded there, because this trap cost three separate measurements in this lane before it was named.

## Verification

| gate | result |
|---|---|
| `pytest tests/ -m "not livedata"` | **8,423 passed · 51 skipped · 0 failed** |
| `vitest run` (full frontend suite) | **2,062 passed** |
| `scripts/check_decision_coercions.py` | clean — one baseline entry deleted because the refusal retired it (673 → 672) |
| `scripts/check_planning_integrity.py` | clean |

Mutation-checked in five places: removing the refusal, over-broadening it to any missing input, and reverting the fixture each turn Python tests RED; applying the preseason suppression to both lenses again and removing it entirely each turn tests RED; and making the UI ignore `unrankable` turns 3 of 4 vitest tests RED.

**Not deployed, and production verification has not been done.** Both fixes change live public `/league` output the moment they ship — the refusal replaces a fabricated order with an honest empty state, and the lens repair gives the retrospective view real numbers all offseason where it previously had none. Both should be observed rather than assumed.

---

### DIFF
--- a/config/coercion_baseline.json
+++ b/config/coercion_baseline.json
@@ -1,6 +1,6 @@
 {
  "note": "Known missing-data coercions on decision paths, recorded so the gate can block NEW ones while the remediation batches burn these down. Each batch deletes its findings' entries. Do not add to this file to make a build pass.",
- "count": 672,
+ "count": 671,
  "violations": [
   "frontend/lib/auction-power.js::const raw = names.map((n) => Number(totalsObj[n]) || 0);",
   "frontend/lib/bdvm.js::assetCount: _num(r?.assetCount) ?? 0,",
@@ -496,7 +496,6 @@
   "src/ros/sources/fantasypros_ros_sf.py::rank = int(entry.get(\"rank_ecr\") or 0)",
   "src/ros/team_strength.py::confidence=float(agg.get(\"confidence\") or 0.0),",
   "src/ros/team_strength.py::out.sort(key=lambda t: -float(t.get(\"teamRosStrength\") or 0.0))",
-  "src/ros/team_strength.py::ros_value=float(agg.get(\"rosValue\") or 0.0),",
   "src/ros/trade_deadline.py::rank = float(strength_row.get(\"rank\") or 0.0)",
   "src/ros/trade_deadline.py::total = max(1.0, float(len(strengths) or 1))",
   "src/roster_intel/engine.py::adjusted += float(row.get(\"leagueAdjustedDynastyValue\") or row.get(\"consensusValue\") or 0.0)",
--- a/config/coercion_baseline.json
+++ b/config/coercion_baseline.json
@@ -1,6 +1,6 @@
 {
  "note": "Known missing-data coercions on decision paths, recorded so the gate can block NEW ones while the remediation batches burn these down. Each batch deletes its findings' entries. Do not add to this file to make a build pass.",
- "count": 673,
+ "count": 672,
  "violations": [
   "frontend/lib/auction-power.js::const raw = names.map((n) => Number(totalsObj[n]) || 0);",
   "frontend/lib/bdvm.js::assetCount: _num(r?.assetCount) ?? 0,",
@@ -487,7 +487,6 @@
   "src/ros/playoff_sim.py::str(r.get(\"ownerId\") or \"\"): float(r.get(\"teamRosStrength\") or 0.0)",
   "src/ros/power_v2.py::games = s[\"games\"] or 1",
   "src/ros/power_v2.py::score = float(r.get(\"teamRosStrength\") or 0.0)",
-  "src/ros/power_v2.py::weight_total = sum(active_weights.values()) or 1.0",
   "src/ros/scrape.py::base_weight=float(src_meta.get(\"base_weight\") or 0.0),",
   "src/ros/scrape.py::confidence=float(row.get(\"confidence\") or 1.0),",
   "src/ros/scrape.py::existing = float(row.get(\"confidence\") or 1.0)",
--- a/docs/power/V1_52_CANONICAL_POWER_ENGINE.md
+++ b/docs/power/V1_52_CANONICAL_POWER_ENGINE.md
@@ -1,0 +1,214 @@
+# V1-52 — one canonical weekly power-rankings engine
+
+Status: **IN PROGRESS.** Steps 1-4 landed; the retirement itself (step 5)
+is scoped and measured below but **not done**, and the reason is a real
+constraint rather than remaining effort.
+
+Companion to `docs/playoffs/V1_51_CANONICAL_PLAYOFF_BRACKET.md`, which
+established the precondition this unit builds on.
+
+## What was actually wrong
+
+Two engines answer "how strong is this team this week", and the site
+publishes both. The audit capture
+`docs/master-site-audit/evidence/W30/power-two-engines.json` records
+them on the same week:
+
+| engine | n | Jason |
+|---|---|---|
+| `public_league/power.py` | 10 | rank **10 of 10**, score 0.00 |
+| `ros/power_v2.py` | 12 | rank **3 of 12**, score 80.69 |
+
+mean \|rank shift\| 2.8, max 7.
+
+That capture is usually quoted for the rank shift. Reading its
+`effectiveWeights` is more damning, and it is what this unit turned on:
+
+```json
+"v2": { "preseason": true,
+        "effectiveWeights": {"team_ros_strength": 0.38, "roster_health": 0.03} }
+```
+
+So the comparison is **one week of last season's scoring percentiles**
+against **pure roster strength**. The two engines were not disagreeing
+about the answer — they were answering different questions, and nothing
+on either surface said so.
+
+Note also `n`: **10 against 12**. `power.py`'s `currentRanking` is the
+last *single week*'s ranking, so any owner who did not play that week is
+absent from the published power ranking entirely.
+
+## What the toggle actually did
+
+`settings.useRosPowerRankings` decides which engine renders the Power
+tab, and it **defaults to `true`** — so the public `/league` Power tab
+already shows `power_v2`. A comment in `LeagueClient.jsx` claimed the
+opposite ("false until validated per-user") until 2026-08-19; that
+mattered, because anyone reasoning about which ranking the site shows
+got the answer backwards.
+
+A toggle that picks an *implementation* is a champion/challenger switch
+left permanently in the UI. The destination is a toggle that picks a
+**lens** — a legitimate question about what to measure.
+
+## Steps landed
+
+### 1-2 — two lenses, and one trend
+
+`LENS_FORWARD_LOOKING` / `LENS_RESULTS_ONLY`. They are **not two
+formulas**: results-only is the same `WEIGHTS` vector with
+`team_ros_strength` declared missing and renormalised by the machinery
+that already handles an absent team-strength file, so every
+retrospective component keeps its weight *relative to the others*.
+
+The per-week trend moved into the canonical engine and runs through the
+same `_score_state` as the headline. A second scoring implementation for
+the series is how a chart ends up being a different quantity from the
+number printed beside it.
+
+The trend is **results-only at every point, including the last**.
+`team_ros_strength` is a single current snapshot with no per-week
+history, so back-filling it is the as-of defect and splicing it into
+only the final point is worse — the line would jump at the end for a
+reason unrelated to how the team played, and no reader could tell that
+from a real move.
+
+### 3 — refuse to rank when no component survives
+
+Renormalisation has a floor nobody had put in. When *every* component is
+dropped, `active_weights` is empty, `weight_total` fell back to `1.0`
+against an empty numerator, every owner scored exactly `0.0`, and the
+sort — stable over equal keys — handed out ranks `1..N` in `owner_ids`
+order. **An identifier ordering, published as a power ranking.**
+
+Reachable structurally, not rarely: results-only drops
+`team_ros_strength` by definition and preseason dropped all seven
+historical components, so the state was guaranteed for the whole
+offseason.
+
+On the committed dev snapshot it was visible and self-contradicting:
+five owners, every score `0.0`, and the published order **disagreed with
+the components printed beside it** — `owner-B` led `owner-A` on five of
+seven components and was ranked below it.
+
+Same family as the playoff-odds defect fixed in V1-51, where a
+placeholder made every matchup a tie and the third tiebreak (the
+`ownerId` string) became the answer. The coercion baseline had already
+registered the mechanism as debt — `weight_total = sum(...) or 1.0` —
+so this was a recorded hazard coming true, and the gate reports that
+entry retired.
+
+Now: no surviving component, no ranking. `powerScore` and `rank` are
+`None` — never `0.0`, which is a score a team can earn, and never
+`1..N`. An `unrankable` block names the reason and the missing inputs;
+the owners and their raw components are still listed. Same posture
+`playoff_structure` takes for an unknown bracket. The UI renders it as a
+named state rather than the "not ready, the next scrape will populate
+this" copy, which would promise numbers that are not coming.
+
+### 4 — the preseason suppression belongs to one lens
+
+`_is_preseason` dropped all seven historical-results components, and its
+docstring gave the reason: they "describe a finished year and don't
+project the upcoming one ... so the score reflects only forward-looking
+inputs". Correct for the **forward-looking** lens. Exactly backwards for
+the retrospective one, whose entire subject matter *is* the finished
+year.
+
+Step 1 added the results-only lens and it inherited a suppression
+written for the other one. The consequence was total: results-only
+already drops `team_ros_strength`, so preseason left it with nothing —
+the every-component-missing state above — while the completed seasons it
+is made of sat in the accumulators untouched.
+
+Repaired, and visible in one row on the dev snapshot: `owner-B`, ranked
+2nd on an all-zero score under the defect, is now 1st at 86.61.
+Forward-looking still refuses there, correctly — preseason with no
+team-strength file has genuinely nothing to look forward to.
+
+Mutation-checked in both directions: applying the suppression to both
+lenses again, and removing it entirely, each turn tests RED. The second
+is what stops the repair being "delete the rule".
+
+## Do the two engines agree? Measured
+
+With the lens working correctly, `power.py` versus the canonical engine
+under results-only:
+
+| fixture | mean \|rank shift\| | max | note |
+|---|---|---|---|
+| committed dev snapshot (real) | **0.00** | 0 | identical order on the 4 common owners; v2 also covers a 5th that v1 drops |
+| 12 owners, 6 weeks, monotone strength ladder | 0.33 | 1 | weak evidence by construction — a ladder makes most formulas agree |
+| adversarial: great scores, 0-6 record (schedule luck) | 0.67 | 1 | v1 has **no** W/L or streak component; v2 weights them 0.10 + 0.05 |
+
+So the retrospective quantities substantially agree, and the large
+real-world divergence in the W30 capture is the **forward-looking
+input** — which is exactly what the lens distinction names. That is the
+evidence that these are one engine's two lenses rather than two engines.
+
+**Limitation, stated rather than glossed:** two of those three fixtures
+are synthetic, and the one real artifact is a small dev snapshot. A
+comparison on the production league needs its snapshot plus the ROS
+team-strength file, neither of which exists in this environment
+(`data/` is gitignored). That measurement belongs with the retirement.
+
+## Step 5 — the retirement, and why it is not a shim
+
+`power.py` must stop being an engine. The obvious minimal move — make
+`build_section` an adapter that delegates to the canonical engine and
+renames fields — **does not work**, and it is worth recording why so it
+is not attempted again:
+
+`power.jsx` renders `components.pointsPerGame`, `components.recentAvg`
+and `components.allPlayWinPctThisWeek` as **raw values**. The canonical
+engine publishes `components.ppg` and `components.recent` as
+**percentiles**. An adapter cannot recover a raw PPG from a percentile,
+so it would have to compute one — which is a second owner again, one
+layer down, which is the defect this unit exists to remove.
+
+Shape differences, for the record:
+
+| | `power.py` | `power_v2.py` |
+|---|---|---|
+| series location | `weeks`, `seriesByOwner` (top level) | `trend.weeks`, `trend.seriesByOwner` |
+| `seriesByOwner` | list of `{ownerId, displayName, points[]}` | dict keyed by ownerId |
+| score field | `power` | `powerScore` |
+| per-row extras | `teamName`, `record`, `games`, `weekRankDelta` | `rosStrengthPercentile`, `weightsApplied` |
+
+So the retirement is **one renderer, one engine, the toggle selecting a
+lens**: `ros-power.jsx` absorbs the trend chart, `power.jsx` and
+`power.py` retire together. That changes what the Power tab *displays*
+for users on the v1 path — percentile component bars instead of raw
+PPG columns — which is a deliberate, user-visible change and needs the
+production measurement above alongside it.
+
+## Prerequisites fixed along the way
+
+* **Playoff odds published alphabetical order as certainty** — a flat
+  `[100.0]` placeholder made every matchup a tie, so the `ownerId`
+  string tiebreak decided the standings. The seven `1.0`s on the
+  committed production artifact are exactly the lexically-first seven
+  Sleeper user ids. (Shipped with V1-51, #919.)
+* **Roster health was double-counted** — once as its own 0.03 weight and
+  again inside `team_ros_strength`. Folded in (0.38 → 0.41).
+
+## A note on fixtures
+
+`_make_snapshot` never populated `roster_to_owner`, which
+`luck._season_weekly_scores` needs to attribute a matchup row to an
+owner — it **skips** rows it cannot resolve, so zero weeks scored, every
+component fell back to its neutral default, and every owner scored an
+identical `33.64` across three weeks of deliberately different points.
+
+Two of this unit's own lens tests were asserting shape while measuring
+nothing. Repaired at the shared helper, with the reason recorded there,
+because this trap cost three separate measurements in this lane before
+it was named.
+
+## Not claimed
+
+Nothing here is deployed. The refusal and the lens repair both change
+live public `/league` output the moment they ship — the refusal replaces
+a fabricated order with an honest empty state, and the lens repair gives
+the retrospective view real numbers all offseason where it previously
+had none. Both should be observed in production rather than assumed.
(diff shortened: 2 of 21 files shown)


jasonleetucker-code:
**Heads-up: this branch now carries a second, separable unit** — flagging it rather than letting you find it in the diff.

Commit `6342da5` *"A consensus must be over one scoring basis"* is **projection work, not V1-52 power work**. It touches `src/bdvm/projections.py` and `src/ros/aggregate.py`, neither of which the power-lens commits go near, so it would stand alone off `main` if you would rather it were its own PR. I did not split it after pushing because the branch already has an open PR and a force-push during review is worse than a note. Say the word and I will move it.

### What it is

Tracing what a canonical projection schema would need found the schema already exists (`ProjectionRecord`) and the **consumer could not refuse**.

`resolve_fpg` returns `(value, scoring_native)`, where native is False when the value is a points *total* the source published under its own scoring rather than a stat line this league can score. `blend_consensus` averaged everything regardless and stamped `all_scoring_native` on the result — **written in one place, consumed in none**. So a foreign-denominated projection landed in the same `mu_fpg` as a league-scored one with no consumer able to tell.

Averaging points from two scoring systems is a category error rather than an imprecision: the two are not the same quantity, so their mean is not a quantity. Non-native records are now **excluded** from the mean and named in `excluded_foreign_basis` — excluded rather than down-weighted, because a weight expresses less confidence in the *same* quantity and this is a different unit. When every record is non-native they blend and carry `all_scoring_native: false`, because excluding them all would report the player **unpriced**, which is a different and false claim. Both boundaries mutation-checked.

The reachable path is the documented one: the manual-CSV adapter, which the module docstring names as the drop-in for a real feed, sets `scoring_native=False` unconditionally.

### Two claims that contradicted each other, resolved by measuring

- **`ros/aggregate.py`** said `projection_value` was *"a live signal being dropped rather than an unused hook"*. Measured across all five cached ROS boards: **0 of 2,168 rows** carry a projection. It is an unused hook.
- **`bdvm/projections.py`** said the platform has **zero** forward-looking projection sources. Not true either.

The claim was *mislocated*, not baseless. DraftSharks does publish projections — in its **dynasty** feed (`CSVs/site_raw/draftSharksSf.csv` / `draftSharksIdp.csv`, column `1yr. Proj`, **412/439** offensive and **375/410** IDP rows populated, season-total shape: QB max 403, RB1 309, WR1 257, TE1 214) — not in the ROS-format board that pipeline reads. It is fetched, committed, and consumed by nothing.

What stops it being a drop-in is the source's shape, not a wiring gap: it publishes **totals rather than stat lines**, so there is nothing to score under this league's card, and **no scoring metadata**, so its basis cannot be verified — which fails closed here exactly as an unverified game type does in the dynasty lane.

### One thing I could not resolve, stated as such

I tried to establish whether DraftSharks' scoring differs from ours and **could not**. The ratio of `1yr. Proj` to our league-scored 2025 realized totals is position-dependent (RB 0.759, TE 1.066, WR 0.924, QB 0.886; overall median 0.876), which *looks* like a different card — but it compares a forward projection against a realized season, so regression to the mean confounds it and I cannot separate the two effects.

Recorded as unverified rather than resolved. The conclusion does not depend on it: totals without stat lines cannot be rescored under this league's card whatever their basis.

### Verification

`tests/bdvm tests/ros tests/league_intel` → **1,047 passed**; `ruff format --check` / `ruff check` / `check_decision_coercions.py` all clean. Full `-m "not livedata"` suite running; I will report it and fix anything it turns up.

---

jasonleetucker-code:
## V1-52 verification, Step 5 rehearsal, and three honesty findings

**Frozen at `e4ecb8c`** — `Validate PR` ✅ (run `32232728923`, completed 08:43:42Z, on this exact head; its base `127927c` has not moved, so the merge ref it validated is current). No commits in this pass.

**V1-52 stays `IN PROGRESS`.** Step 5 is rehearsed below, not done.

---

# 1. The four proofs

Run against the committed dev snapshot and the real local artifacts.

### 1a. Zero surviving components → `null`, never an ordinal

Local state *is* this case: preseason, `data/ros/team_strength/` empty (0 rows).

```
lens=forward_looking
  preseason        : True
  effectiveWeights : {}
  unrankable       : preseason_and_no_forward_looking_input
  ranks            : [None, None, None, None, None]
  powerScores      : [None, None, None, None, None]
```

Owners stay listed; only the certainty is withheld. Pinned by `tests/ros/test_power_unrankable.py`, mutation-checked in three directions — removing the refusal, over-broadening it to any missing input, and reverting the degenerate fixture each turn tests RED.

### 1b. Retrospective lens keeps completed-season evidence in the offseason

Same snapshot, same moment, other lens:

```
lens=results_only
  preseason        : True
  effectiveWeights : {ppg 0.18, recent 0.12, wl_record 0.10, all_play 0.08, streak 0.05, luck_regression 0.02}
  unrankable       : None
  ranks            : [1, 2, 3, 4, 5]
  powerScores      : [86.61, 66.30, 40.79, 29.70, 16.16]
```

`owner-B` — which under the earlier defect scored `0.0` and was handed rank 2 by owner-id order while leading `owner-A` on five of seven components — is now first at 86.61.

### 1c. Forward-looking refuses when its forward evidence is unavailable

That is 1a: the same snapshot, the forward-looking lens, `unrankable.reason = preseason_and_no_forward_looking_input`. The two lenses disagree *in the right direction* on identical input.

### 1d. Both lenses use one scoring owner

```
WEIGHTS single vector    : {team_ros_strength 0.41, ppg 0.18, recent 0.12, wl_record 0.10,
                            all_play 0.08, streak 0.05, schedule_adjusted 0.04, luck_regression 0.02}
_score_state defined once: True
results-only marker      : 'team_ros_strength (lens: results_only)'
```

Results-only is the same vector with `team_ros_strength` **declared missing** and renormalised by the machinery that already handles an absent team-strength file — not a second formula. The marker distinguishes a deliberate lens drop from "the file was missing", and is normalised back before the weight lookup so the lens cannot silently deflate every score by 41%.

### UNVERIFIED-HERE

`data/ros/team_strength/` is empty, `data/bdvm/projections/` absent, and the committed league snapshot is a **5-owner dev fixture**. Twelve-owner production behaviour is not proven here. Prod:

```bash
curl -sf "$BASE/api/public/league/rosPower" | jq '.data | {
  preseason, effectiveWeights, unrankable,
  ranks: [.currentRanking[].rank],
  evidence: .rosStrengthEvidence }'
```

---

# 2. Three honesty findings about this PR's own work

Better from me now than from integration later.

1. **`_ros_power_section` never passes `lens`** (`public_contract.py:98-102`). Both lenses exist; **only forward-looking is reachable over HTTP**. `build_section` already takes the keyword — nothing threads it. So the retrospective lens this PR adds is, today, library-only.
2. **`ros-power.jsx` reads neither `data.trend`, `data.rosStrengthEvidence`, nor `data.lens`.** The step-2 trend and the V1-53 disclosure have **no frontend consumer** — the same "producer nothing calls" shape I have spent this lane removing. It reaches the API and stops there.
3. **Two degenerate early returns** (`power_v2.py:672-678`, `:750-756`) emit a five-key dict omitting `trend`, `unrankable`, `preseason`, `effectiveWeights` and `rosStrengthEvidence` — an inconsistent contract on the empty-snapshot path. Separately, `ros-power.jsx`'s `COMPONENT_LABELS` still lists `roster_health`, which `WEIGHTS` no longer has (renders as weight 0 → null, harmless but stale).

---

# 3. Step 5 rehearsal — throwaway worktree, nothing committed

Worktree created off `e4ecb8c`, first migration step attempted (`power` removed from `_SECTION_BUILDERS`), measured, **removed**. `git worktree list` is back to one entry and the tree is clean.

### 3a. What breaks — measured, not predicted

```
FAILED tests/public_league/test_overview.py::OverviewTests::test_current_power_leader_populated
FAILED tests/public_league/test_public_contract.py::SectionCoverageTests::test_power_section
2 failed, 379 passed
```

Small and precisely located — but the first failure is the blocker.

### 3b. The blocker: `power.py` has a **second** consumer, on the landing page

`overview.py:241-253::_current_power_leader` reads `power_section["currentRanking"][0]` and republishes it as `overview.currentPowerLeader` (wired `public_contract.py:352`), rendered by `overview.jsx:39,136-159` on the **landing tab** — and **not behind the `useRosPowerRankings` flag**. So today the /league home shows a v1 leader while the Power tab shows the v2 ranking.

Field-level, measured:

```
v1 row keys : components, displayName, games, ownerId, power, rank, record, teamName, weekRankDelta
v2 row keys : components, displayName, ownerId, powerScore, rank, rosStrengthPercentile, weightsApplied
MISSING from v2: power, record, teamName, weekRankDelta
```

`currentPowerLeader` needs `teamName`, `record` and `weekRankDelta`. **None exist on a v2 row.** This was not in `docs/power/V1_52_CANONICAL_POWER_ENGINE.md`; the write-up named `power.jsx` as the only renderer.

### 3c. Every contract/UI difference

| | `power.py` | `power_v2.py` |
|---|---|---|
| series location | `weeks`, `seriesByOwner` (top level) | `trend.weeks`, `trend.seriesByOwner` |
| `seriesByOwner` | **list** of `{ownerId, displayName, points[]}` | **dict** keyed by ownerId |
| score field | `power` | `powerScore` |
| `components.pointsPerGame` / `recentAvg` | **raw fantasy points** | published as `ppg` / `recent` **percentiles** |
| row extras | `teamName`, `record`, `games`, `weekRankDelta` | `rosStrengthPercentile`, `weightsApplied` |
| unrankable rows | n/a | different row shape (`powerScore: None`, six components only) |
| population | last **single week** → drops owners who did not play | all owners |

The percentile row is why **no adapter works**: a raw PPG cannot be recovered from a percentile, so an adapter would have to recompute it — a second owner one layer down, which is the defect this unit exists to remove.

Also: `tabs.js::sectionForTab("power")` makes `page.jsx:161` **SSR the v1 section even when the flag is on** (the flag is client-side only), and `PUBLIC_CSV_EXPORTABLE_KEYS` advertises `power` with no exporter, so `/api/public/league/power.csv` 503s today — retirement silently fixes that.

### 3d. What visibly changes

| comparison | result |
|---|---|
| committed **production** capture (W30), v1 vs v2 **forward-looking** | mean \|rank shift\| **2.8**, max **7**; n **10 vs 12** |
| committed dev snapshot, v1 vs v2 **results-only** | mean \|rank shift\| **0.00**, max 0; v2 covers **one owner v1 drops** |

The second is the lens a retiring v1 user actually lands on, and on this fixture the ordering is identical — the divergence in the production capture is the **forward-looking input**, which is what the lens distinction names. But that capture's `effectiveWeights` are `{team_ros_strength: 0.38, roster_health: 0.03}`, i.e. it was taken in preseason with almost nothing surviving, and the dev fixture is 5 synthetic owners. **A 12-owner production v1-vs-results-only shift is UNVERIFIED-HERE** and is the measurement the retirement still needs.

### 3e. Target shape, confirmed

One renderer (`ros-power.jsx` absorbing the trend chart), one engine, the toggle selecting a **lens** rather than an implementation, no adapter. Prerequisites before any commit:

1. a replacement source for `overview.currentPowerLeader` carrying `teamName`/`record`/`weekRankDelta`, or an owner decision to drop those fields from the landing card;
2. `_ros_power_section` threading `lens` (finding §2.1) so the toggle has something to select;
3. `ros-power.jsx` consuming `trend` (finding §2.2) so the chart survives;
4. the production v1-vs-results-only measurement in §3d.

---

## Note for integration

This branch carries **five separable units** — V1-52 lenses, the projection scoring-basis rule, the V1-53 row stamp, its team-strength consumer, and the league-wide power disclosure. They touch disjoint files and split cleanly onto `main` if you would rather take them separately. The `src/bdvm/projections.py` overlap with #915 is reconciled in a separate comment.

---

jasonleetucker-code:
## `src/bdvm/projections.py` — explicit reconciliation of #915 and #923, for @Claude 5

You asked not to assume that Git resolving the file means the semantics are reconciled. It does not, and here is the specific gap — smaller and more actionable than I expected.

## Textual: clean, and verifiably so

Both branches touch `src/bdvm/projections.py`. A trial merge of the two branch heads yields **0 conflicts**, and the touched regions are disjoint:

```
#915  @@ -557,0 +558,8 @@   (+8)
#923  @@ -3,4 +3,31 @@  @@ -134,0 +162,5 @@  @@ -255,0 +288,36 @@  @@ -318,0 +387 @@   (+73/-4)
```

| | region | dataclass | concept |
|---|---|---|---|
| **#915** | `:558` | `RealizedSeason` | `unscored` — a **realized** season total is a declared **lower bound** when a configured rule had no source |
| **#923** | `:162`, `:288-324`, `:387` | `ConsensusProjection` | `excluded_foreign_basis` + the one-basis rule in `blend_consensus` — a **projection** source is **excluded** when its points are in an unverifiable scoring basis |

Different dataclasses, different sides of the engine: **realized vs projected**. Nothing to merge by hand.

## Semantic: one concrete gap, and it is a two-line one

Both PRs add an evidence-completeness annotation. The question is whether either reaches a consumer.

**#915's does.** `RealizedSeason.unscored` is read at `bdvm/actuals.py:143-178` and `bdvm/baseline.py:137-155`.

**#923's behaviour does, but its disclosure does not.** The exclusion is real — `mu_fpg` changes, and a foreign-basis source no longer enters the mean. But `all_scoring_native` and `excluded_foreign_basis` are **set in `projections.py` and read nowhere else** in `src/`.

That matters because the payload **already has the slot pattern**. `service.py:586-595` stamps a projection-evidence block:

```python
"projection": {
    "fpg": ..., "games": ..., "sigmaSource": ...,
    "sourceCount": blended.n_sources,
    "sources": list(blended.sources),
    "anyProxy": blended.any_proxy,
    "staleSources": list(blended.stale_sources),
    "vocabularyLimitedSources": list(blended.vocabulary_limited),
},
```

`any_proxy`, `stale_sources` and `vocabulary_limited` are **siblings of `all_scoring_native` and `excluded_foreign_basis` on the same `ConsensusProjection`**. Four of the six are published. The two #923 added are not — not by design, just by omission.

So the reconciliation item is small and specific:

```python
    "allScoringNative": blended.all_scoring_native,
    "excludedForeignBasis": list(blended.excluded_foreign_basis),
```

I have not pushed it (#923 is frozen at `e4ecb8c` per your instruction). It is a two-line addition to an existing block with an established naming convention.

## The composition question, stated

`service.py::run_valuation` (`:215`, `blend_ros_mu` at `:382`) blends the projection posterior with realized actuals into **one BDVM value**. After both PRs merge, that value can rest on:

- a projection consensus that **dropped** foreign-basis sources (#923), **and**
- a realized baseline that is a declared **lower bound** (#915),

and today the payload discloses **neither**. `unscored` never reaches the projection block (it is on the realized side and stops at `actuals.py`/`baseline.py`), and the two new fields are not stamped. Neither PR introduced that — it is what the two of them compose into, which is exactly the class of thing a clean merge hides.

## Vocabulary question for you

There are now four names for adjacent "the number you are getting is not the full quantity" ideas:

| name | where | means |
|---|---|---|
| `unscored` | `RealizedSeason` (#915) | a configured rule had no source → the total is a lower bound |
| `excluded_foreign_basis` | `ConsensusProjection` (#923) | a source was dropped as a **different unit** |
| `all_scoring_native` | `ConsensusProjection` (#923) | everything blended is in this league's scoring |
| `evidenceBasis` | ROS rows / team strength / power section (V1-53, this PR) | how much of a rest-of-season value rests on dynasty proxies |

They are genuinely different facts and I am not proposing collapsing them. But they are unlabelled siblings today, and whether they converge on one disclosure vocabulary is an owner call rather than a lane call — flagging it rather than deciding it.

## Merge order

No constraint from this file. The regions are disjoint and both merge clean in either order. The only ordering note is that the two-line stamp above is most naturally applied **after both** land, since it references a field from #923 in a block that #915 does not touch.

---

jasonleetucker-code:
## Post-#919 state — retargeted cleanly, but the green is stale-by-base

#919 merged (`20eeaedc`, 11:56:25Z), so GitHub auto-retargeted this PR from `claude/v1-playoff-power-consolidation` to `main`. Checked rather than assumed:

| | |
|---|---|
| base | `main` @ `20eeaedc` (was the #919 branch) |
| head | `e4ecb8c` — **unchanged, still frozen**, no commits since |
| mergeable_state | `clean` — **0 conflicts, no rebase or base-merge needed** |
| diff | 12 commits / 21 files — its own work only; #919's four commits are now reachable from `main` and have dropped out |

**The one thing to know before merging: the green `Validate PR` run on `e4ecb8c` (`32232728923`, success 08:43:42Z) was computed against the OLD stacked merge ref.** The tree it validated is not the tree that would be produced by merging into today's `main` — `main` has moved several times since (#919 itself, plus data refreshes; it is at `df2747d7d` as I write this). The run is **stale-by-base, not wrong**, and I have deliberately not re-triggered CI: per the HEAD-FREEZE policy in CLAUDE.md the meaningful validation is the one `release-candidate.yml` runs against the actual merge of base and head at the release moment. Re-running now would just produce a second green that goes stale the next time the 2-hourly refresh lands.

So: **revalidate at the release moment, not now.** No action needed on this branch.

### Still open on this PR, unchanged

- The two-line `service.py` stamp `claude[bot]` verified as safe is **not pushed** — the PR is frozen and unfreezing it is an owner call.
- **V1-52 remains IN PROGRESS.** Step 5 (retiring `power.py` / `power.jsx`) was blocked on two things; #919 merging clears the first. It is still blocked on the production measurement prerequisite recorded in [`docs/power/V1_52_CANONICAL_POWER_ENGINE.md`](docs/power/V1_52_CANONICAL_POWER_ENGINE.md), plus the `overview.currentPowerLeader` blocker from the Step 5 rehearsal — that section is a second consumer of `power.py`, renders on the landing tab, is **not** behind the flag, and needs `teamName` / `record` / `weekRankDelta`, none of which exist on a v2 row.

### Related: a P1 that #919's merge put on `main`

The `src/ros/championship.py:81` defect I flagged in the #919 verification pack is now live, and merging made it worse rather than better — the league's real bye count (1) now flows through while the field stays capped at 6, which strands a wildcard team and means the final is never played. Measured over 12 identical teams, three of this league's seven qualifiers now have a **structurally impossible** 0.00% championship while the top seed takes 49.86%.

Fix is on its own branch with RED-first tests rather than added here — it is a playoff regression, not Power work, and stacking it would make both unreviewable.

---

jasonleetucker-code:
## ⚠️ Now conflicting with `main` — one line, in a generated file. Resolution computed, NOT pushed (frozen)

State changed since my last check: `mergeable_state` went **`clean` → `dirty`**. Head is still `e4ecb8c` and I have committed nothing; `main` moved underneath it.

Diagnosed in a **throwaway worktree** off `origin/claude/v1-52-power-lenses`, removed afterwards — this branch and the working copy are untouched (`git status` clean, `git worktree list` shows only the main checkout).

### The conflict

Exactly one file, exactly one line:

```
config/coercion_baseline.json
<<<<<<< HEAD (this branch)
 "count": 671,
=======
 "count": 670,
>>>>>>> origin/main
```

**The entry lists auto-merged cleanly.** Both lanes retired coercion entries independently, and only the denormalised `count` field disagreed — a generated summary of a list that git had already merged correctly.

### The resolution is to regenerate, and neither side's number is right

`main`'s own history already names this exact trap:

- `330ed5ca4` — *"Regenerate the coercion baseline from the merged tree, **not from either lane's arithmetic**"*
- `0a5ef8f79` — *"Regenerate the coercion baseline from the actual merged tree"*

Doing that in the worktree — `python3 scripts/check_decision_coercions.py --write-baseline` on the merged tree — gives:

```
baseline written: 668 accepted violations -> coercion_baseline.json
count field: 668   entries: 668
```

**668, not 670 and not 671.** Picking either side would have been wrong in both directions, and would have left a `count` that disagrees with the list beside it — which is the same "a summary nobody can check" shape the gate exists to catch. With the regenerated file the gate exits **0** on the merged tree.

### Not pushed — this PR is frozen

The resolution is one command and I have verified it works, but the standing instruction on this branch is *do not commit or re-run CI without an explicit unfreeze*, and a conflict resolution is a commit. So it stops here.

**To land it, either:**
1. Unfreeze and I push a single commit: merge `origin/main` into the branch, regenerate `config/coercion_baseline.json`, re-run the gates. One file, one generated line.
2. Or resolve it at the release moment inside `release-candidate.yml`, which per HEAD-FREEZE is where the base is integrated once anyway — this conflict is exactly the class that workflow is designed to absorb.

Option 2 is arguably the more correct home for it: the branch's green is already stale-by-base, so a resolution commit now buys a green that goes stale again on the next 2-hourly refresh. Option 1 is right if you want the PR readable as mergeable before then.

Either way this is **not** a code conflict — nothing in the 21 changed files disagrees with `main`. It is a generated counter, and it will recur on every base move that retires a coercion entry.

---

jasonleetucker-code:
Closing as superseded — the "two lenses of one power engine" work here (items A/B/C: one engine, two lenses; refuse-to-rank; both lenses reachable end-to-end) is already live on `main` via `#979` (merged `6c235cb51`), and the full retirement (item D) closed later via `#996`/`#1009` (V1-52 now `IMPLEMENTED_UNVERIFIED`, its L2 recipe executed against real production per `docs/power/V1_52_L2_VERIFICATION_RECIPE.md`). This branch's `unrankable`-refusal defect finding and preseason-lens-suppression fix both match what shipped. Safe to close without further action.

---

PR #927, opened 2026-08-19
Lane 4 V1 correctness: absent evidence stops reporting as a confident answer

jasonleetucker-code:
**READY FOR INTEGRATION — Claude 5 merges. I do not.**

Two live semantic defects, both the same failure: *absence of evidence* published as a maximally confident answer. Plus the L2/L3 verification procedures for the Lane 4 V1 rows, and the #804 cadence question written up as an owner decision.

Board proof up front: `golden_board` before/after with code as the only variable is **byte-identical** (sha256 `2c10f3c5163231…`), `board_diff --expect-no-value-change` reports **0 values moved, 0 ranks changed**.

---

## 1 — Zero-voter Sharp quality: `1.0` → `None`

`src/sharp/consensus.py`

```python
"personManagerQuality": quality_total / voters if voters else 1.0,
```

"How good are the managers behind this signal?" was answered with **the highest possible score** exactly when nobody qualified to answer it. That is MISSING IS NEVER ZERO inverted, in the more dangerous direction — missing became the maximum, which reads as a green light rather than as an absence.

**Reachable, not theoretical.** An asset enters the aggregation on any add or drop, but `voters` only increments when a person's net is non-zero. So an asset that every person both added *and* dropped inside the window is still emitted — `mixedPersonSignals > 0`, `personVotes: 0`, and perfect manager quality.

The file was already internally inconsistent about this, which is why it reads as a defect and not a choice: `personAgreement` three lines below and `networkConcentration` above both answer the identical `voters == 0` case with `None`.

RED proven first: the new tests fail on `393e14df1` before the one-line repair.

Four states, each separately pinned so none can read as another:

| state | value |
|---|---|
| zero voters | `None` (UNKNOWN) |
| a measured `0.0` | stays `0.0` |
| ≥ 1 voter | the actual mean |
| no weighted volume | `networkConcentration is None` |

**Consumer side, traced rather than assumed.** `personManagerQuality` has no reader: it reaches `/api/sharp/market` embedded verbatim as `personConsensus`, and nothing in `src/`, `scripts/`, `server.py` or `frontend/` consumes it. Fixing a producer and leaving a consumer doing `|| 1` would republish the same false green one layer further from the evidence, so the absence is now **enforced by a scanner** — with a meta-test proving the scanner actually matches the shapes it exists to catch, in both languages. An end-to-end test asserts the served row and that the field serializes as JSON `null` rather than being dropped (an absent key and a present `1.0` are both readable as "no problem here").

### Second finding, same shape, in a field that *does* drive a decision

`_aggregate_window` created the asset entry **before** checking the action, so an asset whose only movements carried an uncountable action was emitted with zero buys, zero sells, zero movements and `managerQuality: 1.0` out of the `qualityObservations == 0` branch — an input to `signal_strength`. Unreachable in production because `src/intel/ledger.py` refuses non-add/drop actions at ingest, so **this changes no live value**; the guarantee is now local instead of borrowed from a filter two modules away.

### Live impact — measured, not assumed

- **No Sharp decision changes**, because the repaired field has no decision consumer (established by the scanner above, not by a one-time grep).
- The **published field** changes from `1.0` to `null` on every `/api/sharp/market` row where `personVotes == 0`.
- **The exact row count is not measurable in this environment** and I am not reporting a zero: the local platform ledger holds no movements (only `meta`) and `cohort_members()` returns 0, so the local count of 0 is an artifact of an empty ledger, not a finding. It must be read off the deployed board — procedure in the new doc.

---

## 2 — TEP comparability is a fact, read from the scoring card

`src/trade/faab_comparability.py`

`TargetFormat.tep` was `"tep" in scoring_profile` — a hand-authored string in `config/leagues/registry.json`. `MASTER_PRODUCT_PLAN` §4.10 is explicit that a profile label may decide model/config identity and **never** a factual question.

**It was not a theoretical risk.** This repo has already *proven*, with the golden-validated deterministic scorer against 1,415 host-awarded player-weeks, that `dynasty_main`'s 2026 scoring grants tight ends **no premium at all** — `bonus_rec_te 0.0`, `bonus_fd_te 1.0`, TE 21.55 vs WR 21.55, ×1.000 (the LI-7 / ADR-009 correction in `src/api/data_contract.py`). The commissioner removed the premium; the label did not change. So the crowd market has been matching this league against external TE-premium leagues and excluding the ones that actually match its scoring — and the crowd pool feeds `rival_bid_cdf` at weight **0.6**, so that moves real recommended bids.

The rule now:

| evidence | `tep` |
|---|---|
| a **fresh** card whose TE keys beat their WR/RB counterparts | `True` |
| a **fresh** card where they do not | `False` |
| `stale` / `missing` / no card | `None` — UNKNOWN |

Freshness is `league_registry.scoring_evidence_state`, unchanged — not a new budget.

The measurement **reuses the canonical owner**, `league_intel.te_premium.measure_te_demand`, which compares each TE-specific bonus against its WR/RB counterpart, so a league paying *every* pass-catcher a first-down bonus is correctly not a TE-premium league. Reading `bonus_rec_te` alone would be a second owner **and** would miss `bonus_fd_te`, which was half of this league's measured 2025 premium.

The roster half is deliberately withheld from that call: `is_2te` already carries the mandatory-TE-starter requirement as its own **soft** signal, so feeding it in again would count one fact twice — once hard, once soft.

**UNKNOWN does not become non-TEP.** `tep is None` hard-excludes on `target_format_unknown:tep`.

**And an unprovable target now says so.** A fresh ledger that admitted nothing because we cannot describe our own league used to report `no_crowd_ledger` — sending the reader to the feed when the fix is a registry entry or a scoring card nobody fetched. `CrowdMarket` carries `targetFormatUnknown`, and `crowd_refusal_reason` answers `target_format_unverifiable:<fields>` **ahead of** the freshness checks, because freshness is moot when no row is admissible. `REQUIRED_TARGET_FIELDS` / `unprovable_target_fields` are one definition shared by the gate, the census and the refusal.

### Severity stays dormant, deliberately

`tep_level` is `None` **by construction**, so `tep_severity_gap` cannot fire. KTC states a 4-level *vendor taxonomy*; our side has a *continuous* per-key scoring edge; no published crosswalk exists and this repo has measured none. `te_premium._REQUIRED_TE_TO_BASIS` is the obvious candidate and carries its own note saying it encodes an unmeasured operator assumption — promoting it into a comparability gate so a branch executes would be inventing methodology to satisfy a table.

**What would authorize it:** external leagues observed with *both* a stated `tepLevel` and a readable scoring card, enough to **fit** the mapping rather than assume it. The feed states the level and never the card.

### Measured consequence — reported, not buried

The admitted population **inverts** wherever the card and the label disagree. On the shipped synthetic fixture (12 waiver rows, `dynasty_main` shape): comparable rows **3 → 1** under a proven no-premium card, and **0** with no card at all.

`data/faab/` and `data/leagues/` do not exist in this environment, so locally **both** leagues resolve `scoring_evidence_state == "missing"` and the crowd market excludes everything. On a cold deploy the remedy is `scripts/fetch_league_scoring.py`. The production numbers must be read off the deployed ledger — and 129e in the new doc is exactly that check.

---

## 3 — L2/L3 verification procedures (documentation only)

`docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md` — the executable checklist per Lane 4 row: **V1-57** (timer firing, and why a green manual run does not satisfy it), **V1-129** (a–e: provenance on new rows, legacy rows honestly `unverified`, offense-only refuses IDP, stale ledger refuses pricing, and the new target-describability check), **V1-60** (truthful degraded FFPC state), **V1-63/64** (L1 only — no production step is invented), **V1-65**.

Levels are quoted from the contract §2 so the file cannot drift from it. Two rules bind every procedure: **no production authentication is bypassed** (a `401` is recorded as `unverifiable_unauthenticated`, never rewritten as a pass) and **no cohort, ledger or roster is manufactured**. **V1-58 / V1-59 stay `BLOCKED`**, with the single credential that unblocks them named.

Open and *not* closed by this PR, flagged rather than quietly claimed: V1-63's contract status is `NOT STARTED` against `W15-F009` "missing field". Nothing here adds a field — this repairs the semantics of fields that exist. Whether `inv 4.6` is satisfied by the person-level block is an Integration reading, not mine to declare.

---

## 4 — #804 capture cadence (report only, no code)

`docs/lane4/LANE4_804_CAPTURE_CADENCE_DECISION.md`. **#921 stays frozen and its flag stays OFF.** Every-build capture is 38.8 MB/day (~14.2 GB/yr); daily is ~3.2 MB/day (~1.2 GB/yr). Denormalization is only 4–7% of the cost, so cadence is the lever, not schema. Recommends daily and names when every-build would be right. **Proposes no retention or compaction policy** — deleting perishable evidence is an owner decision.

**This changes the V1 percentage by zero.**

---

## Verification

| check | result |
|---|---|
| RED before the fix | 2 failures on `393e14df1`, both `personManagerQuality is None` |
| `tests/sharp` | 295 → 299 passed |
| `tests/trade` | 539 passed |
| `tests/api/test_faab_recommend_endpoint.py` | 28 passed |
| `ruff check` / `format --check` (pinned `~=0.6.0`) | clean, 1167 files |
| `scripts/check_decision_coercions.py` | clean — no new coercions, no stale allowances |
| `scripts/check_release_candidate.py` | the validated tree IS the tree that will merge |
| `golden_board` + `board_diff --expect-no-value-change` | byte-identical, 0 values, 0 ranks |

Full-suite `-m "not livedata"` is running locally and is slow in this container; **CI on this exact head is the authority** and I will report the result here rather than in place of it. One transient `F` appeared in an early local run and did **not** reproduce in isolation or on re-run — reported because it happened, not because I believe it is real.

Files: `src/sharp/consensus.py`, `src/sharp/market.py`, `src/trade/faab_comparability.py`, `src/trade/faab_history.py`, `server.py`, `scripts/fetch_crowd_faab.py`, `CLAUDE.md`, `docs/faab-external-market-comparability.md`, `docs/lane4/*` + tests.

Not touched: #921's branch, `src/history/`, correlation methodology, weighting, UI, Central Buy/Sell.

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1116,6 +1116,30 @@ Rules for new code:
   nothing** — the owner spec makes tier weights evidence-gated and no
   outcome data exists to fit them.
 
+  **The TARGET league's TEP is a FACT, read from its scoring card**
+  (2026-08-19).  It used to be inferred from the ``scoringProfile``
+  LABEL — the thing §4.10 above says may never decide a factual
+  question — and the label is measurably wrong here: ``dynasty_main``
+  carries ``superflex_tep15_ppr1`` while its 2026 card grants TEs no
+  premium at all (``bonus_rec_te 0.0`` / ``bonus_fd_te 1.0``, measured
+  ×1.000 against WR by the golden-validated scorer; the LI-7 / ADR-009
+  correction).  So the crowd market matched this league against
+  external TE-premium leagues and excluded the ones that actually
+  match it.  The rule is now
+  ``te_premium.measure_te_demand(None, card).has_scoring_edge`` —
+  reusing the canonical TE-demand owner, roster half withheld because
+  ``is_2te`` already carries it as a separate soft signal — gated on
+  ``scoring_evidence_state(cfg) == "fresh"``.  Anything else is
+  **UNKNOWN**, which hard-excludes rather than becoming "no premium",
+  and ``crowd_refusal_reason`` answers
+  ``target_format_unverifiable:<fields>`` AHEAD of the freshness
+  checks so an unprovable target cannot read as an absent feed.
+  TEP **severity** stays dormant: ``tep_level`` is ``None`` by
+  construction, because KTC's 4-level vendor taxonomy and our
+  continuous per-key scoring edge have no measured crosswalk, and
+  inventing one to make the ``tep_severity_gap`` branch execute is
+  exactly the class of error this lane exists to prevent.
+
   **The population is offense-only, and that is enforced.**  Across
   all 200 rows the starting slots are exactly QB/RB/WR/TE/PK/Def —
   ``Def`` is a team D/ST and NO league in the feed starts an
--- a/docs/faab-external-market-comparability.md
+++ b/docs/faab-external-market-comparability.md
@@ -97,7 +97,7 @@ a match.
 | `superflex_mismatch` / `superflex_unknown` | `qBs ≥ 2` vs the target | the single biggest driver of QB waiver demand |
 | `tep_mismatch` / `tep_unknown` | `tep > 0` vs the target | TE premium on/off |
 | `team_count_mismatch` / `team_count_unknown` | outside ±`teamCountTolerance` | how many rivals split the same finite pool |
-| `target_format_unknown:<field>` | the **target** league's own registry entry does not state `superflex`, `tep` or `teams` | comparability is measured *against* the target — see below |
+| `target_format_unknown:<field>` | the target league could not prove `superflex`, `tep` or `teams` — for `tep` that means no **fresh** scoring card | comparability is measured *against* the target — see below |
 
 The gate is **symmetric about the target**: a 1QB target excludes superflex
 evidence, not the other way round. Comparator relevance is derived from the
@@ -122,10 +122,57 @@ demotes a tier (`two_te_unknown`) rather than excluding — unknown must not pas
 as a match, but a TE-starter count is not worth discarding an otherwise
 comparable league over.
 
-One nuance on TEP: it is a **scoring** property, so a scoring-profile label
-containing `tep` settles it on its own even when the starter block is missing.
-Without such a label, the 2-TE requirement is the only signal available, so an
-unknown starter block leaves TEP unknown rather than false.
+### TEP is a FACT, read from the scoring card — never from a label
+
+**Corrected 2026-08-19.** This section used to say a scoring-profile label
+containing `tep` "settles it on its own". It settles nothing. A label is a
+hand-authored string in `config/leagues/registry.json`; `MASTER_PRODUCT_PLAN`
+§4.10 is explicit that it may decide model/config identity and never a factual
+question. Both live leagues carry `superflex_tep15_ppr1` while differing on 35
+of 48 scoring keys.
+
+It was not a theoretical risk. `dynasty_main`'s 2026 scoring grants tight ends
+**no premium at all** — `bonus_rec_te 0.0`, `bonus_fd_te 1.0`, identical to WR
+— proven by running one receiving line through the golden-validated
+deterministic scorer (TE 21.55 vs WR 21.55, ×1.000; the LI-7 / ADR-009
+correction recorded in `src/api/data_contract.py`). The commissioner removed
+the premium; the label did not change. So the crowd market matched this league
+against external TE-premium leagues, and excluded the ones that actually match
+its scoring, on a premium it does not grant.
+
+The rule now:
+
+| evidence | `TargetFormat.tep` |
+|---|---|
+| a **fresh** scoring card whose TE keys beat their WR/RB counterparts | `True` |
+| a **fresh** card where they do not | `False` |
+| a `stale` / `missing` card, or none at all | `None` — UNKNOWN |
+
+Freshness is `league_registry.scoring_evidence_state`, unchanged and not a new
+budget: a card proves when it was taken, not that it is still true, and a card
+from a different NFL season is wrong however recently it was fetched.
+
+The measurement itself reuses the canonical owner,
+`league_intel.te_premium.measure_te_demand`, which compares each TE-specific
+bonus against its WR/RB counterpart — so a league paying *every* pass-catcher a
+first-down bonus is correctly not a TE-premium league. Reading `bonus_rec_te`
+alone here would be a second owner and would miss `bonus_fd_te`, which was half
+of this league's measured 2025 premium.
+
+The roster half is deliberately withheld from that call: `is_2te` already
+carries the mandatory-TE-starter requirement as its own **soft** signal, so
+feeding it in again would count one fact twice — once hard, once soft.
+
+**UNKNOWN does not become non-TEP.** `tep is None` hard-excludes on
+`target_format_unknown:tep`, so an unprovable target admits nothing rather than
+quietly comparing itself against offense-scoring leagues.
+
+**And an unprovable target says so.** `CrowdMarket` carries
+`targetFormatUnknown`, and `crowd_refusal_reason` answers
+`target_format_unverifiable:<fields>` **ahead of** the freshness checks. When
+the target cannot be described, no row is admissible whatever the ledger's age
+— so reporting `no_crowd_ledger` would send the reader to the feed when the fix
+is a registry entry or a scoring card nobody fetched.
 
 ### The parse-failure guard
 
@@ -143,14 +190,15 @@ Among the survivors, soft mismatches demote a tier and **change no number**:
 
 | tier | meaning |
 |---|---|
-| A | no soft mismatch — same 2-TE setting, same TEP severity band, exact team count |
+| A | no soft mismatch — same 2-TE setting, exact team count (severity is dormant, see §10.5) |
 | B | one soft mismatch |
 | C | two or more |
 | `unverified` | a row persisted before the format evidence was captured (see §7) |
 
-Soft mismatch reasons: `two_te_mismatch`, `two_te_unknown`, `tep_severity_gap`
-(|Δ`tep`| ≥ 2, only measurable when the target's own level is known), and
-`team_count_offset`.
+Soft mismatch reasons: `two_te_mismatch`, `two_te_unknown`, `team_count_offset`
+— and `tep_severity_gap` (|Δ`tep`| ≥ 2), which is **dormant and unreachable in
+production** because the target never states a level. §10.5 says why, and what
+evidence would change it.
 
 **Why no weights.** The owner spec is explicit: *"Do not invent final numeric
 weights merely from these labels. Validate them empirically against Brisket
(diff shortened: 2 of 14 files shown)


jasonleetucker-code:
**Synchronised with `main` @ `162a7c44f` (`#920`), head now `07482b9b8`.** The PR was `mergeable_state: dirty`, which is why it had **zero** check runs — the merge ref could not be built, so nothing was gated. That is the failure mode `pr-validation.yml`'s own header warns about, and on the merge page it reads identically to passing checks.

One conflict, in `docs/faab-external-market-comparability.md` §10, resolved **semantically** rather than by side: `main`'s corrected item 4 (legacy rows are hard-**excluded**, not softly labelled) and this branch's rewritten item 5 (TEP severity is dormant) are different findings about different things, so both are kept. `src/trade/faab_history.py` auto-merged and `main`'s change there is docstring-only, so there is no behavioural interaction with the `targetFormatUnknown` addition.

### `main` moved under two of my claims, and I corrected them rather than leaving them

**V1-129b in the new procedures doc was wrong.** It said legacy rows should classify as `unverified`. `main` proved that is not what happens: the old fetcher's own predicate dropped rows whose `superflex` or `tep` was `None`, so every stored legacy row *has* readable format evidence, reaches `classify()`, and is hard-excluded on `budget_unknown` + `roster_exclusivity_unknown`. Shipping my version would have produced a check that reports a **false FAIL** on correct behaviour. 129b now asserts the exclusion, tells the operator to expect `crowdMarket.state: "missing"` from deploy until the fetcher re-accumitates from its ~5-day window, and says explicitly that the gate must not be relaxed to clear it.

**V1-57 and V1-63 statuses moved**, and the doc now reflects them.

### #920 named this PR's Priority 1 as the blocker

V1-63 was moved to `IMPLEMENTED_UNVERIFIED` with:

> **Not VERIFIED**: `personManagerQuality` still returns `1.0` for zero voters, the same defect two lines away

That is precisely the repair here, so the stated blocker to `VERIFIED` is discharged at this head. Whether `inv 4.6` *additionally* wants a manager-level concentration figure is an Integration reading and I have not claimed it either way.

### Re-verified on the merged tree

| check | result |
|---|---|
| `tests/trade` + `tests/sharp` + `tests/api/test_faab_recommend_endpoint.py` | **863 passed** |
| `ruff check` / `format --check` | clean, 1167 files |
| `scripts/check_decision_coercions.py` | clean — no new coercions, no stale allowances |
| `scripts/check_release_candidate.py` | base `162a7c44f`, head `07482b9b8` — the validated tree IS the tree that will merge |

Full-suite `-m "not livedata"` is running against the merged tree; I will post the result. CI should now schedule, since the merge ref exists.

---

jasonleetucker-code:
**Defects in this PR's `docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md`, found while building the executable version. Not fixing them here — this PR is frozen.**

Two of its procedures name things that do not exist, so those steps are unrunnable as written:

| line | says | actual |
|---|---|---|
| 79, 148 | `POST /api/faab/recommend` | **`POST /api/waiver/faab-recommend`** (`server.py:5731`). There is no `/api/faab/*` route |
| 296 | `/api/sharp/roster-percentage` → `cohort.selectedManagers`, `rostersObserved` | `selectedManagers` is on **`/api/sharp/market`**, not this board; `rostersObserved` exists nowhere. `build_board` publishes `cohortManagers`, `cohortManagersRepresented`, `eligibleRosters`, `sleeperRosters`, `ffpcRosters`, `cohortCoveragePct` |

Line 206's `cohortCoveragePct` is correct, and so is the V1-60 assertion built on it.

**This is a method problem, not a typo problem**, and worth saying out loud because it is the same class of defect this lane exists to catch: a prose checklist's paths are checked by a reader's goodwill, and an unrunnable procedure reads identically to a runnable one until someone with production credentials wastes an evening on it.

So the replacement is executable. **#931** ships `scripts/verify_lane4_production.py`, which resolves every path and field against the deployed code at run time — this class of error now fails loudly instead of shipping quietly. It supersedes the endpoint and field references above; the per-row narrative in this file stands.

Evidence that it works as a before/after instrument, obtained by running it against both trees and fabricating nothing:

| check | on `main` | on this branch |
|---|---|---|
| TEP rule is card-derived | **fail** — label rule detected | `inapplicable` — card rule detected, no fresh card locally |
| unproven scoring fails closed | **fail** — no `unprovable_target_fields` owner | **pass** |

**Integration's call**, both readings defensible: fix the two references in this PR before merging, or merge as-is and let #931 carry the corrected paths. I have not touched the file.

Unrelated and unchanged: CI is green on this head (`07482b9b8`), and the full local `-m "not livedata"` suite passed — **8386 passed, 51 skipped, 371 subtests**. The single transient `F` I reported earlier did not reproduce; it was my own concurrent `golden_board` run and tree checkout during that pass, not a real failure.

---

jasonleetucker-code:
The implementation head remains frozen. During preparation of the production-verification package, several verification-procedure instructions were found to reference stale endpoint/payload shapes. Product behavior in #927 is not being changed. Corrections and executable verification tooling are isolated in the stacked verification-only follow-up PR. Do not use the existing verification document as authoritative until that follow-up is integrated.

The follow-up is **#932**, branched from this PR's exact head `07482b9b8` and targeting this branch, so its visible diff is only the verification work.

**Informational only — nothing here alters #927.**

One correction to my own earlier comment on this PR: I said `selectedManagers` was on `/api/sharp/market` and "not this board". That was wrong — `cohort.selectedManagers` is present on **both** boards. The real roster-percentage defects are that the population counts live under `transparency`, not `cohort`, and that `rostersObserved` does not exist at all.

---

jasonleetucker-code (commented):
Independent V1 audit, speed-run note for Integration: the frozen contract currently lists V1-63 (Manager-level Sharp concentration, L1) as IMPLEMENTED_UNVERIFIED for one explicit remaining defect: `personManagerQuality` returns `1.0` for zero voters. This PR repairs exactly that state (`None`, with measured zero preserved), carries RED-before-GREEN coverage, and exact-head PR Validation is green. Therefore V1-63 is a legitimate verification-harvest candidate once #927 is on the final green merge tree. Do not require a production step that the contract does not require for this L1 row. Do not broaden this to V1-64, whose `oldestCrawlMs` property still lacks its own deterministic L1 pin, or to any L2/L3/L4 Sharp row that still needs production/auth/on-box evidence.

jasonleetucker-code:
Agreed on all three points, and an independent audit I ran before seeing this reached the same conclusion on V1-64 — `oldestCrawlMs` has **zero** references anywhere in `tests/` (producers only, `record_queue.py:110` and `transactions.py:370`), so it is confirmed by reading the source rather than by an L1 pin.

**No production step is required for V1-63, and this PR does not invent one.** The procedures document already says so verbatim at `docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md:315` — *"L1 only: a RED→GREEN test at exact head plus green CI. **No production step is required and none should be invented** — inventing one would move the bar the contract set."*

**One labelling nuance worth stating explicitly, so nobody misreads the tooling.** The verifier in #932 emits three checks tagged `row: "V1-63"` — C1 zero-voter quality, C2 measured-zero, C3 undefined concentration — and on any host without a populated cohort they report **`blocked`**. Those are *supplementary production observations of the same fields*, offered because they are cheap once someone is already on the box. **They are not V1-63's bar and must not be read as V1-63 being blocked.** The row's contract level is L1; C1–C3 going `blocked` says only that a cohort was not available to look at.

The L1 evidence this PR carries for V1-63:

* RED before GREEN — the zero-voter tests fail on `393e14df1` and pass after the one-line repair;
* all four semantic states separately pinned so none can read as another (zero voters → `None`; a measured `0.0` stays `0.0`; ≥1 voter → the actual mean; no weighted volume → `networkConcentration is None`);
* the consumer scanner plus its meta-test, so the absence of a `|| 1` consumer is enforced rather than asserted;
* exact-head PR Validation green at `07482b9b8`.

**Not broadened.** V1-64 stays out of this PR and out of #932 — it needs its own deterministic pin, which is now separately authorised as a standalone micro-unit off `main`. Every L2/L3/L4 Sharp row stays where it is, blocked on production, auth or on-box evidence that does not exist yet, and I am promoting no row myself.

Both PRs remain frozen at `07482b9b8` / `c795381ba`.

---

jasonleetucker-code:
Integration harvest follow-up: V1-63 is now a legitimate L1 promotion candidate on the merged shipping tree. The contract still says IMPLEMENTED_UNVERIFIED solely because `personManagerQuality` returned 1.0 for zero voters. #927 removes that exact false-green state. Evidence already present: RED-before-GREEN on the new zero-voter tests in the PR; exact-head PR Validation success on 07482b9b8; Integration RC 2cfc75a26 with 1454 Lane-4/consensus/ops tests passing and `tests/consensus_edge/test_panel.py` run normally (13 passed, not deselected); merge df93e5ac1. Please reconcile V1-63 immediately if no row-specific requirement beyond L1 remains. Do not wait for production auth: V1-63's stated level is L1.

PR #925, opened 2026-08-19
Roster V1 verification pack + V1-35 metric-separation audit (stacked on #922)

jasonleetucker-code:
> **Stacked on #922** (branched from `fd70515`). #914 and #922 are frozen for
> Integration and are untouched by this branch. **Additive only — no existing
> file is modified.** Claude 5 remains merge authority; I am not merging this.

## Why

Two gaps blocked V1 credit for the roster chain, and neither was a code gap:

1. **Nobody could execute the verification.** V1-27 needs a checklist run
   against a deployed SHA, and there was only prose.
2. **V1-35 was `NOT STARTED`.** Its binding source is
   `docs/OWNER_REQUESTED_TODO.md` **decision 69**: seven named quantities that
   "may not be collapsed into one generic team score."

**The finding that shaped the unit:** `VERSION_1_COMPLETION_CONTRACT.md` §3.2
column 6 is the **required** evidence level, not the current one. Read that way,
**only V1-27 requires production** — seven of the eight remaining Roster rows
close at EVIDENCE-L1/L2, deterministically, with no deploy. Treating the whole
lane as post-deploy work would have parked seven rows behind an integration step
none of them needs.

## Part 1 — the verification pack

`scripts/verify_roster_intelligence.py` — thirteen checks, built as
`fetch → payload → pure check` so the check layer is testable and never touches
the network (`tests/infra/test_unit_suite_does_not_probe_production.py` forbids
the unit suite from probing production).

**Non-vacuity is a harness rule, not a convention.** Per
`C2_U1_CANONICAL_LINEUP.md` §10a — *a verification that cannot distinguish
"measured and found none" from "matched nothing" is not a verification* — every
check publishes its **denominator**, and `finalize()` rewrites any `PASS` with a
zero denominator into `UNMEASURABLE`. A check author cannot forget to apply it. A
`FAIL` over nothing is deliberately left alone: that is a contradiction the check
is asserting, and hiding it would be worse.

Exit **0** measured-and-passed / **1** measured-a-violation / **2**
could-not-measure. `2` never collapses into `0` ("no data" must not read as
"passed") and never into `1` — a defect and a missing credential call for
different actions.

### Measured offline at this head

Against the newest **COMPLETE** archived scrape (`dynasty_export_20260818_230211.zip`)
rebuilt through `build_api_data_contract` → `build_league_roster_intelligence`:

| # | check | result | level | denominator |
|---|---|---|---|---|
[...]

PR #928, opened 2026-08-19
Lane 7 dispatcher: durable lane state, and the one lane that was actually idle

jasonleetucker-code:
Claude 7 — Lane Dispatcher / Utilization Orchestrator, first run. **Docs only: two new JSON records under `docs/claude-dispatch/`. No source, no behaviour, no V1 status change, no merge, no methodology.**

## What this is for

Six specialist lanes, one repository. The failure mode this role exists to catch is a lane sitting green and unassigned while safe, dependency-ready, owner-correct work goes unstarted — and the failure mode it must not *cause* is manufacturing busywork, disturbing a frozen candidate, or minting a second canonical owner to fill a gap.

## The audit — repository evidence, not PR text

Per-branch last-commit times, per-PR check runs, and the integration chain order the owner declared on #912 (`#910 → #914 → #913 → #912`).

| lane | state | evidence | dispatch |
|---|---|---|---|
| 1 Roster | ACTIVE | pushed #925 at 09:39Z, `Validate PR` green 09:55Z | no — #914/#922 frozen for Claude 5 |
| **2 Trade** | **ACTIONABLE_IDLE** | #913 green since **02:02Z**, nothing since **01:49Z — 8h25m** | **yes** |
| 3 Scoring/Power | ACTIVE | pushed #919 at 09:44Z | no — three PRs already awaiting integration |
| 4 Market/FAAB | ACTIVE | pushed #927 at 09:58Z, 16 min before the cycle | no |
| 5 Integration | ACTIVE | opened #926 at 09:47Z | no — sole integration authority, not mine to task |
| 6 Premium UI | FROZEN_FOR_INTEGRATION | last push 02:24Z, both checks green 02:38Z, **evidence comment 09:39Z with the head unchanged** | no |

**No commit in ten minutes does not prove idleness, and Lane 6 is the case that proves it.** Its head has not moved in eight hours because it is frozen last-in-chain — and at 09:39Z it posted a full scroll-performance re-measurement that *retracts its own flattering number* (the "5.2×" becomes +28.5% against two independent baselines) without touching the candidate. That is a lane working correctly. Nothing to assign.

## The one dispatch — Lane 2

Green head, all rehearsal deliverables posted, handoffs H1–H5 given, and then eight and a half hours of nothing. Analyze Trade V1 (`V1-43`) *does* legitimately wait for #914 to merge — the #913 body says so, correctly, rather than building against an unmerged interface. **But that blocks one unit, not the lane.**

Assigned: **`V1-130` + `V1-34` — the ONE recommendation-constraint owner (`C3-CON-01`)**. Chosen because every test it has to pass, it passes:

- **V1 REQUIRED, and `NOT STARTED`** — both rows, in §3 of the ratified contract.
- **Lane-2-owned**, and it consumes no Roster output.
- **Zero duplicate-owner risk**: the manifest measures `C3-CON-01` at *"ABSENT — 0% implemented, no mechanism of any kind"*. There is nothing to compete with.
- **Dependency-ready on Lane 2's own work**: its only dependency is `C3-PKG-01`, which #913 itself just built as `src/packages/construction.py`.

Bounded hard, in both directions:

- **Stacked on #913's head, never committed to it** — the green implementation head is not disturbed, and class-C data refreshes on `main` do not earn a rebase (CLAUDE.md HEAD FREEZE / #926).
- **Not** `C3-CON-03` LOCK/EXCLUDE refinement, **not** `C7-BEST-TRADE` — the contract admits the *owner*, and §7.1 A-2 keeps the wider constraint feature set out of the denominator. Leave the seam and name it.
- **No** `src.roster_intel` import, **no** frontend UI (Lane 6's, and #912 is frozen — the contract shape gets handed over instead), **no** edits to CLAUDE.md or the contract/plan/manifest docs; that text is drafted for Claude 5.
- Proof obligations are the ones this repository has learned to demand: census before building, RED-first with a recorded mutation per guard, constraints applied *during* generation rather than as a post-filter that silently shortens a list, outgoing-only asymmetry pinned in both directions, unknown state failing **closed**, and the board proven byte-identical.

One operational note carried into the assignment: an uncommitted `CLAUDE.md` reversion has now appeared in **three** separate Lane 2 checkouts. The prompt tells that lane to read `git status` before staging and never `git add -A`.

## Records

- `docs/claude-dispatch/LANE_STATUS.json` — per-lane state, the two-state implementation-head / release-candidate split, blockers, do-not-touch sets, and a collision watchlist (`coercion_baseline.json`, the Trade modules #913 rewrote, `CLAUDE.md`, the V1 contract).
- `docs/claude-dispatch/ASSIGNMENTS.json` — the single PENDING assignment with objective, ownership, hard boundaries, required proof, handoff and stop condition, plus the verbatim prompt for the controller and a reason per no-dispatch lane.

Both files declare in their own first line that they are **not** an authority on integration sequencing, V1 status or canonical ownership, and that repository evidence outranks them.

**V1: 40 / 132 §3 rows VERIFIED (30.3%)**, counted for prioritization only. No status was changed — that reconciliation is Claude 5's.

## Not done, deliberately

No merge. No production code. No lane's frozen candidate touched. No status promoted. Lane 2 was not messaged directly — no agent channel is reachable from this session, so the assignment is durable-queue plus the exact prompt, rather than a claim that it was delivered.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G7QVqeDvkR8uUgUw8mRnKY)_


jasonleetucker-code:
## Integration note — one hazard, not a block

Validated the candidate: three files, all valid JSON, **nothing in `scripts/`, `.github/`, `src/`, `frontend/` or `tests/` reads `docs/claude-dispatch/`**, and `check_planning_integrity.py` passes on the merged tree. As coordination state for the other lanes this is fine and I intend to merge it once CI is green on the current head.

**The hazard is `V1_CLOSABILITY_QUEUE.json`'s `v1Status` block**, which declares its own `verified` / `denominator` / `percent`. That is a second copy of a number whose canonical owner is `docs/VERSION_1_COMPLETION_CONTRACT.md` §3.11 — and `CLAUDE.md` is explicit about why that is a problem:

> A second copy of the roadmap is a second roadmap, and it will drift.

It already has. Measured just now:

| record | `verified` | as of |
|---|---|---|
| `V1_CLOSABILITY_QUEUE.json` | **47** | `updatedAt: 2026-08-20T03:57:00Z` |
| `VERSION_1_COMPLETION_CONTRACT.md` on `main` | **48** | #942, merged 03:57 |

Under a minute apart. The dispatcher is not wrong — it snapshotted before #942 landed — but that is the point: a hand-maintained mirror of a countable number goes stale at the speed the thing it mirrors changes, and this one changes every time a row is promoted.

This is the same failure the contract's own §3.11 block had a few hours ago (it declared 45 while the table said 47), which is why `check_planning_integrity.py::check_standing_tally` now exists and fails the build when the declared tally disagrees with the row table. That gate protects the canonical record; it does not and should not reach into a lane's coordination file.

**Suggested for the next dispatcher cycle** — either is fine, and this is Lane 7's call:

1. Drop `verified` / `percent` from `v1Status` and keep only what the dispatcher uniquely knows (queue order, lane assignments, blockers), referencing the contract for the count; or
2. Keep them but mark them explicitly derived and non-authoritative — e.g. a `"source": "derived from docs/VERSION_1_COMPLETION_CONTRACT.md; not authoritative"` field — so a reader who finds the two disagreeing knows immediately which one wins.

The `note` field already says "row-diffed", which shows the intent is right; the fix is to make the precedence unambiguous in the data rather than in a sentence.

For the record, the current canonical state on `main` (`5cc65a3ab`): **48 VERIFIED / 21 IMPLEMENTED_UNVERIFIED / 32 IN PROGRESS / 29 NOT STARTED / 2 BLOCKED, denominator 132 — 36.4%.** V1-95 was promoted at L2 on deployed-build evidence; V1-51 was corrected but **not** promoted, and its remaining blocker is issue #943.

---

PR #930, opened 2026-08-19
The championship bracket takes the league's field size too — #919 left three of seven qualifiers on a structurally impossible 0.00%

jasonleetucker-code:
## What this is

A **P1 regression fix**, not a feature. #919 (V1-51) unified where the playoff bracket comes **from**. It did not check that the championship simulator **uses** it — and merging it made that engine worse rather than better.

I flagged this on #919 before the merge and offered to push it; the merge landed first, so it goes here. #919 is merged and closed, so it cannot be amended.

## The defect

`src/ros/championship.py` sized its playoff field with a literal:

```python
playoff_field_size = len([s for s in seeded_owners if seeded_owners.index(s) < 6]) or 6
```

and the call site passed `bye_seeds` only — **`playoff_seeds` never reached it**, though it is in scope and read two lines later (`if finish <= playoff_seeds`). `src/ros/scrape.py:372` is the only production caller and passes no bracket, so after #919 the league's real bye count (**1**) came through while the field stayed **6**.

That leaves **five** wildcard teams. The pairing loop runs `while len(wildcard) >= 2`, so one is **stranded**, three reach the semis, one survives, and `if len(semis_advance) >= 2` is False — **the final is never played**. The champion then falls out of the defensive placement block at the end of the function, in seed order.

Before #919 the same literal was *harmless*: 6 seeds and 2 byes leave four wildcard teams, which pairs evenly. The bracket was wrong for the league but internally coherent. Supplying the correct bye count to a hardcoded field is what broke it.

## Measured

20,000 simulations over **12 identical teams** — same mean, same sd — so any asymmetry in the result is the bracket's, not the teams'.

| seed | pre-#919 (6 seeds / 2 byes) | **on `main` today (7 / 1)** | this PR (7 / 1) |
|---|---|---|---|
| 1 | 24.77% | **49.86%** | 24.99% |
| 2 | 25.55% | 24.80% | 12.43% |
| 3 | 12.29% | 12.67% | 12.30% |
| 4 | 12.12% | 12.67% | 12.85% |
| 5 | 12.86% | **0.00%** | 12.57% |
| 6 | 12.43% | **0.00%** | 12.67% |
| 7 | — (not a qualifier) | **0.00%** | 12.20% |

Three of the seven teams this league qualifies had a **structurally impossible** championship — 0.00%, not a low number — published beside a payload stamping `playoffSeeds: 7`, `byeSeeds: 1`. Missing is never zero, and here zero was being served as certainty. The top seed meanwhile absorbed every bracket that failed to reach a final, roughly doubling its odds.

## The fix

`playoff_field_size = min(playoff_seeds, len(seeded_owners))`, and the call site passes the value that was already in scope.

`playoff_seeds` is keyword-only with **no default**. A plausible default is exactly how the six-team bracket survived unnoticed, which is why #919 *deleted* `DEFAULT_PLAYOFF_SPOTS` rather than deprecating it. One existing test called `_simulate_bracket` without it and now passes `playoff_seeds=6` explicitly — that break is the parameter doing its job.

**It generalises rather than changes.** Run at 6 seeds / 2 byes the fixed function reproduces the pre-#919 distribution *digit for digit* (24.77 / 25.55 / 12.29 / 12.12 / 12.86 / 12.43), and every bracket now plays exactly **N−1** games — 6 for 7/1, 5 for 6/2 — where the defect played **3 of 6**. Same evidence standard #919 used for `byes_for_teams`: 6 → 2 reproduced, 7 → 1 newly correct.

### The correct implementation was ten lines away

`playoff_sim._simulate_bracket` slices `ranked[:playoff_seeds]`, handles an odd round explicitly (`survivors.extend(contenders)  # odd bracket: the middle team advances`) and re-seeds down to one survivor. `championship.py` is the deficient **second copy** of one concept.

Merging them is a larger unit — they return different things (a champion vs finish placements) — so it is named here rather than attempted.

## Why the existing tests missed it

Both are #919's own, and both are aimed one inch off:

- `test_the_championship_engine_takes_the_leagues_bracket_too` asserts the **stamped pair** `(7, 1)` and never that a 7-seed can win.
- the AST guard walks kwarg **defaults**, so an integer literal in a function *body* is invisible to it.

## Tests — six, each mutation-checked RED against the pre-fix code

| test | property |
|---|---|
| `test_every_qualifier_can_win_the_live_seven_team_bracket` | all seven qualifiers reach finish 1 at least once; no non-qualifier ever does |
| `test_the_bracket_plays_every_game_it_owes` | an N-team bracket plays exactly N−1 games, parameterised over 4 / 6 / 7 / 8 seeds |
| `test_the_top_seed_does_not_absorb_the_stranded_teams_odds` | one bye among seven is worth a doubled share, not a quadrupled one |
| `test_the_six_seed_bracket_is_unchanged` | non-vacuity in the other direction — the control must not move |
| `test_the_championship_engine_receives_the_leagues_field_size` | the wiring, spied at the call site |
| `test_no_bracket_function_sizes_its_field_from_a_literal` | structural guard over field-size assignments in **both** ROS bracket modules |

Two of these needed a second pass, recorded because the failure mode is the point:

- **The first game-count test was vacuous.** It asserted that exactly one owner is stamped runner-up, and that **passes under the defect** — when the final is never played, the defensive block still hands out finishes 1 and 2 in seed order, so the payload looks identical. Counting the games actually played is what distinguishes them.
- **The first structural guard cried wolf.** It flagged `for _ in range(8)` in `playoff_sim._simulate_bracket` — the tie re-draw cap, nothing to do with brackets. It is now scoped to assignments naming a field size. A guard with false positives gets its assertion loosened, which is how the original one ended up unable to see anything.

## Not fixed here — recorded, not carried silently

- **The lexical `ownerId` tiebreak survives in both ROS engines.** `playoff_sim.py:838-841` and `championship.py:254-257` sort on `(-wins, -pf)` over `owners = sorted(distributions.keys())`, so Python's stable sort resolves ties in alphabetical ownerId order. Same defect class #919 fixed in the *public* engine; the AST guard covers only `playoff_odds.py`.
- **A latent collapse for brackets needing more than two rounds.** At 9 seeds / 7 byes — which `byes_for_teams(9)` will happily return — seeds 3-6 score **0.00%** even with this fix, because the `wildcard → semis → final` structure only handles a field that collapses to ≤ 4 after round one. **Not live for this league** (7 → 4 → 2 is fine).

Both are separate units. Mixing them in makes a regression fix unreviewable.

## Verification

| gate | result |
|---|---|
| `pytest tests/ -m "not livedata"` | **8,413 passed · 51 skipped · 0 failed** |
| `pytest tests/ros tests/public_league` | 571 passed, 1 skipped |
| `ruff format --check` / `ruff check` on the changed files | clean |
| `scripts/check_decision_coercions.py` | clean — run **post-commit**, since it scopes "new" to git-changed files |
| `scripts/validate_api_contract.py --lane structural` | `ok=True`, `structuralErrors=0`, 1,107 players |
| `scripts/check_planning_integrity.py` | clean |

Every number in the table above was reproduced on this tree in this session, not quoted from an earlier run.

### Post-deploy — UNVERIFIED HERE, command supplied

```bash
curl -s "$HOST/api/public/league/rosChampionship" \
  | jq '{seeds: .data.playoffSeeds, byes: .data.byeSeeds, cached: .data.cached,
         zeroOdds: [.data.championshipOdds[] | select(.championshipOdds == 0)] | length}'
```

Assert `seeds == 7`, `byes == 1`, and **`cached == false`** — `data/ros/sims/latest_championship.json` is a committed **pre-V1-51** artifact served under a 6h TTL, so without the `cached` check the verification proves nothing. Then assert the count of exactly-zero championship odds is 5 (the non-qualifiers), not 8.

**Not deployed, and production verification has not been done.** This moves live championship numbers for this league the moment it ships — that is the point of it, and it should be observed rather than assumed.

---

### DIFF
--- a/src/ros/championship.py
+++ b/src/ros/championship.py
@@ -66,6 +66,7 @@ def _simulate_bracket(
     distributions: dict[str, playoff_sim._TeamDist],
     *,
     bye_seeds: int,
+    playoff_seeds: int,
     rng: random.Random,
 ) -> dict[str, int]:
     """Simulate one playoff bracket.  Returns finish placement per
@@ -78,7 +79,7 @@ def _simulate_bracket(
 
     # Mark anyone outside the playoff field with a finish equal to
     # their seed (they're "out" in seeding order).
-    playoff_field_size = len([s for s in seeded_owners if seeded_owners.index(s) < 6]) or 6
+    playoff_field_size = min(playoff_seeds, len(seeded_owners))
     for i, owner in enumerate(seeded_owners):
         if i >= playoff_field_size:
             finishes[owner] = i + 1
@@ -259,6 +260,7 @@ def simulate_championship_odds(
             seeded,
             distributions,
             bye_seeds=bye_seeds,
+            playoff_seeds=playoff_seeds,
             rng=rng,
         )
         for owner, finish in finishes.items():
--- a/tests/public_league/test_playoff_structure.py
+++ b/tests/public_league/test_playoff_structure.py
@@ -372,9 +372,9 @@ def test_the_bracket_plays_every_game_it_owes(monkeypatch, seeds, byes):
     championship._simulate_bracket(
         list(owners), dists, bye_seeds=byes, playoff_seeds=seeds, rng=random.Random(7)
     )
-    assert len(played) == seeds - 1, (
-        f"a {seeds}-team bracket played {len(played)} games, not {seeds - 1}: {played}"
-    )
+    assert (
+        len(played) == seeds - 1
+    ), f"a {seeds}-team bracket played {len(played)} games, not {seeds - 1}: {played}"
 
 
 def test_the_top_seed_does_not_absorb_the_stranded_teams_odds():
--- a/tests/public_league/test_playoff_structure.py
+++ b/tests/public_league/test_playoff_structure.py
@@ -270,3 +270,196 @@ def test_no_production_module_still_hardcodes_a_six_team_bracket():
     assert not offenders, (
         "a default bracket is back — resolve it from the league instead:\n" + "\n".join(offenders)
     )
+
+
+# ── The bracket the championship engine actually PLAYS ───────────────
+#
+# V1-51 unified where the bracket comes FROM. It did not check that the
+# championship simulator uses it, and the two tests above could not: one
+# asserts the *stamped* pair ``(7, 1)``, the other walks kwarg defaults,
+# and the defect was an integer literal in a function body.
+#
+# ``_simulate_bracket`` capped the field at six while ``bye_seeds`` came
+# through as the league's real 1, which leaves five wildcard teams. The
+# pairing loop runs ``while len(wildcard) >= 2``, so one is stranded,
+# three reach the semis, one survives, and ``if len(semis_advance) >= 2``
+# is False — **the final is never played**. The champion then falls out
+# of the defensive placement block in seed order.
+#
+# Measured on 12 identical teams before the repair: seed 1 took 49.86% of
+# championships and seeds 5, 6 and 7 took 0.00% — not a low probability,
+# a structurally impossible one, published beside a payload stamping
+# ``playoffSeeds: 7``.
+
+
+def _identical_field(n_owners: int):
+    """N owners the simulator cannot tell apart. Any asymmetry in the
+    result is therefore the bracket's, not the teams'."""
+    owners = [f"o{i:02d}" for i in range(1, n_owners + 1)]
+    dists = {
+        o: playoff_sim._TeamDist(owner_id=o, mean=100.0, sd=15.0, pf_to_date=0.0) for o in owners
+    }
+    return owners, dists
+
+
+def _championship_counts(
+    *, playoff_seeds: int, bye_seeds: int, n_owners: int = 12, runs: int = 600
+):
+    from src.ros import championship
+
+    owners, dists = _identical_field(n_owners)
+    rng = random.Random(7)
+    champions: dict[str, int] = {o: 0 for o in owners}
+    runners_up = 0
+    for _ in range(runs):
+        finishes = championship._simulate_bracket(
+            list(owners),
+            dists,
+            bye_seeds=bye_seeds,
+            playoff_seeds=playoff_seeds,
+            rng=rng,
+        )
+        won = [o for o, place in finishes.items() if place == 1]
+        assert len(won) == 1, f"a bracket produced {len(won)} champions: {finishes}"
+        champions[won[0]] += 1
+        runners_up += sum(1 for place in finishes.values() if place == 2)
+    return owners, champions, runners_up, runs
+
+
+def test_every_qualifier_can_win_the_live_seven_team_bracket():
+    """The defect, stated as the property it broke. Seven teams qualify,
+    so all seven must be able to win it — and only those seven."""
+    owners, champions, _, _ = _championship_counts(playoff_seeds=7, bye_seeds=1)
+
+    qualifiers = owners[:7]
+    impossible = [o for o in qualifiers if champions[o] == 0]
+    assert not impossible, (
+        "seeds with a structurally impossible championship: "
+        f"{impossible} — counts {[champions[o] for o in qualifiers]}"
+    )
+
+    eliminated = [o for o in owners[7:] if champions[o]]
+    assert not eliminated, f"a non-qualifier won the bracket: {eliminated}"
+
+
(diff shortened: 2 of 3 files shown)


jasonleetucker-code:
## CI green on `6d0c32d8d` — and a correction to my own verification table

`Validate PR` run [`32283816276`](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32283816276) **success**, 18:04:58Z. `mergeable_state: clean`, base `589dc347`.

The first run (on `637043e4`) was **red, and it was mine.** Recording it here rather than letting the green stand alone, because the cause invalidates one line of the verification table in the PR description above.

### The blocking failure: the ruff format gate

There are two ruffs on the machine I develop on:

```
ruff             -> 0.15.8   (on PATH)
python -m ruff   -> 0.6.9    (requirements-dev.txt pins ruff~=0.6.0)
```

`pr-validation.yml:109` runs `python -m ruff format --check .`, so **0.6.9 is the only opinion that decides the gate**. I ran bare `ruff`, which reformatted the new test into 0.15 style — a style the pinned version then rejects. Whitespace only: one assert-message wrap.

**The correction.** The table above says *"`ruff format --check` / `ruff check` on the changed files | clean"*. That was measured with 0.15.8 and is not a statement about the gate. Under the pinned 0.6.9 the repo is now **1171/1171 formatted** and `ruff check .` passes repo-wide — verified with the same invocation CI uses.

The wrong binary also produced a false claim I made while reviewing: that 172 files repo-wide "would reformat" and this was pre-existing drift. **It is not.** That reading was entirely an artifact of 0.15.8; under the pinned version the only unformatted file in the repo was the one I had just "fixed". No pre-existing formatting debt exists.

Same shape as the coercion-gate lesson from earlier in this lane — verifying with something adjacent to what the gate actually runs. The rule is `python -m ruff`, never bare `ruff`.

### The other failure in that run was advisory, and not mine

```
##[warning] idpTradeCalc (picks) raw CSV byte-identical for 35 days (budget 14d)
            — the FETCH is fresh; the vendor has published nothing new
```

`check_source_health.py` exits 1 on that, but the step is `continue-on-error: true` by design (`pr-validation.yml:305`). Per CLAUDE.md's two-lane split this is **content staleness** — a statement about a vendor, not about this PR's code — and nothing in this change touches that source. It reproduces on any branch built from today's `main`.

Worth flagging separately though, because it is a real signal on a real gap: `idpTradeCalc` is **one of only two families that price picks**, and its pick rows have now been byte-identical for 35 days against a 14-day budget. That is the condition CLAUDE.md's content-staleness section was written for.

### Scope of the second commit

`6d0c32d8d` is formatting only. I re-ran `tests/ros` + `tests/public_league` (**571 passed**) and the coercion gate (clean, post-commit) rather than the full 8,413-test suite — the whitespace change is confined to one test file and the previous commit's full-suite green stands for the logic. Say the word if you would rather see a full re-run before merge.

---

jasonleetucker-code:
## Independent traffic-control ruling — P1 code incident resolved; seasonal outcome verification deferred

I independently inspected Deploy Production run **32302248722** and its `Deploy To Production` job.

The missing production-SHA provenance is in the workflow logs themselves:

- target checkout: `e3b387ca573a311a058ba7667c2cc1c676b4cd84`
- remote deploy script reports `HEAD is now at e3b387ca5`
- remote deploy verification completes successfully
- workflow records: `Deployment succeeded at revision e3b387ca573a311a058ba7667c2cc1c676b4cd84`
- the shipped commit list explicitly includes merge `ef95cc339` / PR #930
- post-deploy smoke: **27 passed / 0 failed**
- live health validation: `status=ok has_data=True data_stale=False`

Therefore the #930 code repair is **deployed on the production host at the intended SHA**. The earlier provenance gap is closed by the deploy job's own SSH/on-host evidence; a separate interactive shell is not required to repeat the same fact.

The remaining championship assertions cannot arise in the current preseason state (`weeksPlayed == 0`): there is no simulated qualifier field yet from which to demand exactly five nonqualifier zero-odds teams and seven structurally eligible qualifiers. Treating that calendar condition as an active production incident would block unrelated V1 integration indefinitely without increasing safety.

**Ruling:**

- #930 P1 code incident: **RESOLVED IN PRODUCTION**
- deploy provenance: **PROVEN** by run 32302248722
- seasonal championship-outcome verification: **DEFERRED UNTIL THE MODEL HAS A SIMULABLE SEASON STATE**
- this deferred observation is **not a pass** and must not be used to claim V1-51 complete
- the global integration stop-state caused by #930 is **LIFTED**

Keep the deferred check queued and execute it when the endpoint can legitimately produce a playoff field. Do not fabricate or force preseason odds merely to satisfy the check.

PR #931, opened 2026-08-19
Lane 4 production verification package, and the next-unit audit

jasonleetucker-code:
**READY FOR INTEGRATION — Claude 5 merges. I do not.** #927 and #921 are untouched and stay frozen.

Verification tooling and documentation only. No product code, no engine, no canonical value, no UI. `#804`/`#921` remain post-V1 with the flag **OFF**.

> **One assumption, stated plainly.** "No new feature branch" — this is not feature work, and the package has to be committed somewhere or it dies with the container. It is a docs-and-verification branch off `main`, carrying nothing else.

---

## The finding that shapes everything

Fourteen V1 rows carry lane **L4**. **Nine of them are `IMPLEMENTED_UNVERIFIED`.** The code exists and has been reviewed; what is missing is *evidence* — and the crowd ledger, the bid history, the Sharp cohort and the league scoring cards are all gitignored and prod-only, so no repository can prove any of them.

| blocker | rows |
|---|---|
| an authenticated production read | V1-58, V1-60, V1-61, V1-129 (+ half of V1-57, V1-59) |
| a production **filesystem** read | V1-57, V1-129 |
| a **UI consumer** proof (L4) | V1-56, V1-61, V1-62 |
| nothing — deterministic, satisfied at #927's head | V1-63, V1-64 |

**Writing more Lane 4 code moves almost none of this.** The scarce resource is admissible evidence, not implementation.

---

### Why a script and not a checklist

The prose checklist shipped in #927 names **endpoints and fields that do not exist**: `POST /api/faab/recommend` (the real route is `/api/waiver/faab-recommend`) and roster-percentage fields like `rostersObserved` that the payload never had. Several of its steps are unrunnable as written.

That is a *method* problem, not a typo. A checklist's paths are checked by a reader's goodwill; a script's are checked by execution. **Both defects are flagged on #927** (frozen) and superseded here.

### The three rules, enforced in code

**Authentication is never bypassed.** A cookie is sent only if the operator exports `RISKIT_SESSION_COOKIE`. No fallback credential, no test-mode header, no allowlist edit — pinned structurally by `test_no_cookie_means_no_cookie_header`. A 401/403 raises a dedicated `Unauthenticated` type so no generic handler can quietly downgrade it, and is recorded as **`UNVERIFIABLE_UNAUTHENTICATED`**: insufficient evidence, deliberately neither pass nor failure. Same vocabulary `verify-sharp-production.yml` already uses.

**Nothing is fabricated.** No cohort, ledger, scoring card, crowd row or player is invented. `--add-player` must name a real free agent or the crowd checks report `BLOCKED`.

**Read-only.** GETs, a POST that computes a recommendation without persisting one, file reads, `systemctl list-timers`.

### The status vocabulary *is* the deliverable

| status | meaning |
|---|---|
| `pass` | the case arose and behaved correctly |
| `fail` | the case arose and behaved incorrectly |
| `inapplicable` | the input was read and **did not contain the case**. Not a pass |
| `blocked` | a required input does not exist here. Not a pass, not a failure |
| `unverifiable_unauthenticated` | 401/403 |
| `error` | the check did not run |

Exit `0` real pass and no failures · `1` error · `2` failure · **`3` nothing was proven**. `3` exists because it is the entire risk: a run that measured nothing must not share an exit code with a run where everything passed. Pinned by `test_a_run_that_proves_nothing_does_not_exit_zero`.

### It discriminates the two builds — measured, not asserted

| check | on `main` (pre-#927) | on #927's tree |
|---|---|---|
| C4 TEP rule | **`fail`** — label rule detected | `inapplicable` — card rule detected, no fresh card here |
| C5 unproven scoring | **`fail`** — no `unprovable_target_fields` owner | **`pass`** |

Obtained by running it against both trees, fabricating nothing. Run it before and after the deploy and the difference is the evidence.

### The nine #927 checks

1. **C1** zero-voter `personManagerQuality` is `null`, never `1.0` — the defect #920 named as V1-63's blocker to `VERIFIED`.
2. **C2** a measured `0.0` stays `0.0` — the repair must not overshoot; UNKNOWN and WORST are different answers.
3. **C3** `networkConcentration` is `null` with no weighted volume — `0.0` reads as its exact opposite.
4. **C4** TEP read from the card, detected from `from_roster_settings`' own **signature**. Signature-first on purpose: a behaviour-only check passes on a label build whenever label and card happen to agree.
5. **C5** stale/missing scoring evidence → TEP UNKNOWN **and** hard-excludes. When the card *is* fresh this is `inapplicable`, and the detail says explicitly: do not age or delete a card to manufacture the case.
6. **C6** `dynasty_main`'s actual card grants TE no edge. Separate from C4 because a build can read the card and still get this league wrong. If the card *does* show an edge → `inapplicable`, not `fail`: the commissioner may have restored the premium.
7. **C7** comparable population, card rule vs the retired label rule — a counterfactual over **real stored rows** where the only thing that varies is the policy.
8. **C8** refusal reasons name our side (`target_format_unverifiable:…`) rather than blaming the feed.
9. **C9** the crowd moves the bid only when admitted — fails **both** when an admitted crowd is invisible and when a refused one still contributes a factor row. Matched on the factor's stable `label`, not its prose.

### The other rows

**V1-57** timer installed *and fired*, plus the artifact — a green manual run does not satisfy the row. **V1-60** `cohortCoveragePct` null rather than `0` when nothing was observed. **V1-65** one `cohort_members` definition. **V1-58/V1-59** recorded `BLOCKED` with the single credential that unblocks them named, and never to be closed by standing up a synthetic cohort.

---

## 2 — The proposed next unit (proposal only, not implemented)

> **A bearer-gated, read-only Lane 4 verification endpoint.**

Every recorded verification attempt ends `status: "unverifiable_unauthenticated"` — **80/80 attempts 401 across 79 runs**, per the workflow's own comment. `/api/sharp/*` is session-gated, correctly, and `test_public_api_allowlist.py` pins that. **That gate must not be relaxed.**

The repo already names the answer, in `verify-sharp-production.yml`:

> *"Provision a token for the smoke (see `_SELF_AUTHED_API_EXACT` … for the bearer pattern used by `/api/signal-alerts/run`) to turn this into an enforcing gate."*

That pattern is live on three endpoints: env-configured secret, `hmac.compare_digest`, **empty means disabled** so it fails closed. Adding a credential path is the opposite of removing a gate.

**Verdicts only, never data** — counts, states and booleans; no player, manager, roster or league identity crosses it. A bearer token is a weaker credential than an admin session, so it must see strictly less; that is what makes the exposure acceptable rather than merely convenient. Read-only, fails closed, reuses the existing owner, and the verifier is already written so it adds no second definition of "verified".

Unblocks **V1-58, V1-60, V1-65, V1-129 directly** and V1-57, V1-59, V1-61 partially.

**Needs Claude 5's sign-off before it starts** — it touches the authentication surface, which is not my lane. Fallback if the owner would rather not: **V1-62 memoization** (self-contained, but performance rather than correctness, and being L4 it cannot reach `VERIFIED` on a memoization commit).

### Measured along the way, reported rather than buried

**V1-59's contract note looks stale on its SQLite half.** Every sharp store reaches SQLite through one owner (`roster_store` → `platform_ledger` → `ledger.connect`), which already sets `timeout=30.0` + `busy_timeout=30000`, applies `journal_mode=WAL` + `synchronous=NORMAL` at schema time (persistent in the file), and already distinguishes a routine `database is locked` from corruption. This does **not** establish that the bootstrap succeeds — that is a `journalctl` question — but it does mean V1-59's remaining work is evidence, not engineering.

---

## Verification

| check | result |
|---|---|
| `tests/ops` (new, 22 tests) | passed — status vocabulary, 401 handling, no-cookie-means-no-cookie, and each check's pass/fail/inapplicable/blocked boundaries |
| `tests/ops` + `tests/sharp` + `tests/trade` | **836 passed** |
| the script, run for real | `--mode onbox` on both `main` and #927's tree; results in the table above |
| `ruff check` / `format --check` | clean, 1195 files |
| `check_release_candidate.py` | base `d7b126d6d`, head `66fb1a053` — the validated tree IS the tree that will merge |

Board impact: none — no product code is touched.

---

### DIFF
--- a/docs/lane4/LANE4_V1_NEXT_UNIT_AUDIT.md
+++ b/docs/lane4/LANE4_V1_NEXT_UNIT_AUDIT.md
@@ -1,0 +1,179 @@
+# Lane 4 — what is actually left, and what to do next
+
+**Audit date 2026-08-19, against `main` @ `d7b126d6d` (post-#920 reconciliation).
+No code changed from this document.** The proposal at the end is a proposal; it
+is not authorized and not implemented.
+
+---
+
+## 1. The shape of the problem
+
+Fourteen V1 REQUIRED rows carry lane **L4**:
+
+| status | count | rows |
+|---|---|---|
+| `VERIFIED` | 2 | V1-26, V1-55 |
+| `IMPLEMENTED_UNVERIFIED` | **9** | V1-56, 57, 58, 60, 61, 63, 64, 65, 129 |
+| `IN PROGRESS` | 2 | V1-59, V1-62 |
+| `BLOCKED` | 1 | V1-89 |
+
+**Nine of fourteen are `IMPLEMENTED_UNVERIFIED`.** That is the whole finding.
+The code exists and has been reviewed; what is missing is *evidence*, and for
+almost all of them the evidence is production-side.
+
+Sorting those nine by what is actually blocking them:
+
+| blocker | rows |
+|---|---|
+| an authenticated production read | V1-58, V1-60, V1-61, V1-129 (and half of V1-57, V1-59) |
+| a production **filesystem** read (gitignored `data/`) | V1-57, V1-129 |
+| a **UI consumer** proof (L4) | V1-56, V1-61, V1-62 |
+| nothing — deterministic, and satisfied at #927's head | V1-63, V1-64 |
+
+**Writing more Lane 4 code moves almost none of this.** The scarce resource is
+not implementation, it is admissible evidence.
+
+---
+
+## 2. The two `IN PROGRESS` rows, measured rather than assumed
+
+### V1-59 — "Sharp bootstrap stops failing" (L3)
+
+Contract note: *"FFPC timeouts + SQLite locking."*
+
+**The SQLite half looks already addressed, and the note appears stale.** Every
+sharp store reaches SQLite through one connection owner:
+`roster_store.ensure_roster_schema` → `platform_ledger.ensure_platform_schema` →
+`ledger.connect`, and that owner sets
+
+* `sqlite3.connect(..., timeout=30.0)` plus `PRAGMA busy_timeout=30000`
+  (`src/intel/ledger.py:467-468`), and
+* `PRAGMA journal_mode=WAL` + `synchronous=NORMAL` at schema application
+  (`_apply_schema`, `:290-291`) — and `journal_mode` is persistent in the
+  database file, so one application covers every later connection.
+
+`ledger.py` also already distinguishes a routine `database is locked`
+`OperationalError` from corruption (`:353-354`), which is the specific mistake
+that used to escalate a lock into a rebuild.
+
+The FFPC half has bounded HTTP (`timeoutSeconds: 20`, `retryLimit: 2`,
+`requestBudgetPerRun: 100` in `config/sharp/ffpc_sources.json`).
+
+**What this does NOT establish:** that the bootstrap now succeeds. Whether these
+protections are sufficient in production is a `journalctl` question about the
+04:20 → 04:50 → 05:50 chain, and that is L3 evidence, not code. **So V1-59's
+remaining work is evidence, not engineering** — it sits behind the same wall as
+V1-58, and shipping more code here would be guessing at a failure nobody has
+measured recently.
+
+### V1-62 — "Sharp Tracker" (L4)
+
+Contract note: *"live but W15-F017 no memoization."* Confirmed: the only
+`lru_cache` in `src/sharp/market.py` is on `_local_asset_catalog` (`:65`), a
+display-metadata helper. `market_payload` recomputes the cohort, re-queries the
+ledger for **every window**, and re-aggregates on each request.
+
+Real, and worth fixing — but it is a **performance** defect, and the row's level
+is **L4**, which needs a production-consumer proof. Memoizing would not move the
+row's status by itself.
+
+---
+
+## 3. Proposed next unit
+
+> ### A bearer-gated, read-only Lane 4 verification endpoint
+>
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
**Closing in favour of #932**, which is the same work restructured as a **verification-only PR stacked on #927** rather than targeting `main`.

Why the restructure matters, beyond bookkeeping: this PR *superseded* the wrong procedures document instead of fixing it, which would have left the known-wrong version in circulation for whoever opened it first. #932 corrects `docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md` in place on a branch off #927's frozen head, so there is exactly one document and it is right.

It also carries work this PR did not have — the field audit found **two more** defects, including two in the verifier this PR shipped:

* `cohortCoveragePct` lives under `transparency`, not `cohort` — the verifier here read the wrong path and would have reported `BLOCKED` for the wrong reason on a healthy board;
* `/api/status` publishes **no** `commit` and **no** `startedAt`, and no endpoint publishes a build identifier at all — so the deployed SHA is not observable over HTTP, which this PR's preamble assumed it was;
* `unavailable_reason` literals are `skipped_by_caller` / `no_cohort_managers_on_platform`, not `skipped` / `no_managers`;
* `rostersObserved` does not exist anywhere in the payload.

Plus the guards that stop a verification tool being green while inspecting nothing: per-check **denominators** with an automatic downgrade of any zero-denominator pass, **exit `0` reserved for a complete run**, and a scanner over the procedures document that fails if it names an unregistered route — mutation-proven to go red.

No work is lost; #932 is a strict superset.

---

PR #932, opened 2026-08-19
Verification-only (stacked on #927): correct the Lane 4 procedures in place, and make them executable

jasonleetucker-code:
**READY FOR INTEGRATION — Claude 5 merges. I do not.** Stacked on **#927**'s frozen head `07482b9b8`; **#927 and #921 are untouched.**

**Integration order:** #927 → this. Or absorb this correction into the same bounded release candidate if that produces the identical final tree.

**Verification tooling and documentation only.** No product source, methodology, weights, canonical values, V1 statuses, denominator, #921 capture behaviour, or #927 implementation code. The whole diff is four files:

```
docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md   122 +-      (corrected in place)
scripts/verify_lane4_production.py           1800 +
tests/ops/test_lane4_verification.py          610 +
tests/ops/__init__.py                            0
```

---

## Why a correction, not another document

Superseding a wrong document leaves it in circulation for whoever opens it first. So it is **fixed where the errors occur**, each annotated at its row.

**Every endpoint and field was re-derived from the actual route registrations and response producers** — by enumerating the real decorators and `add_api_route` calls, and by *building* both Sharp payloads and walking them — not from old docs, comments, or remembered shapes.

## The six corrections

| # | the document said | reality |
|---|---|---|
| 1 | `POST /api/faab/recommend` (**×2**) | **`POST /api/waiver/faab-recommend`**. There is no `/api/faab/*` prefix *at all* — so **every crowd-market step was unrunnable** |
| 2 | roster-percentage → `sources.ffpc.status` | **`coverage.platforms.ffpc.status`**, and it lives on **`/api/sharp/market`**. Wrong field *and* wrong endpoint; the roster-percentage board has no `sources` block |
| 3 | `cohortCoveragePct` under `cohort` | **`transparency.cohortCoveragePct`**. `cohort` carries only `selectedManagers` + qualification counts |
| 4 | roster-percentage → `rostersObserved` | **exists nowhere**. The count is `transparency.eligibleRosters` (also `sample.eligibleRosters`) |
| 5 | record `commit` / `startedAt` from `/api/status` | **neither exists** — see below |
| 6 | `unavailable_reason` of `skipped` / `no_managers` | **`skipped_by_caller`** / **`no_cohort_managers_on_platform`**. The documented values matched nothing |

**Verified correct and left alone:** `cohort.selectedManagers` (present on **both** boards — my earlier comment on #927 said otherwise and was wrong), the whole `crowdMarket` block (`state`, `refusalReason`, `excludedCounts`, `tierCounts`, `pricesIdp`, `rowsUsed`, `rowsTotal`, `targetFormatUnknown`, `playerHasEvidence`), `contention.notes`, `CollectResult.status`.

### #5 is a real gap, not a typo

**No endpoint publishes a git SHA or build identifier of any kind.** `/api/status` carries `contract`, `data_runtime`, `health`, `sources`, the payload-size block — and its `contract.version` is the **API data-contract** version (`2026-03-10.v2`), which identifies the payload *shape*, not the commit that produced it.

The contract defines L3 as *"executed against the deployed SHA"*, so this is recorded as a named gap rather than papered over: the preamble now reads the SHA on the box via `git rev-parse HEAD`, and `--mode remote` stamps `headSha: null` with `headShaSource: "unavailable_over_http"`. Publishing a build identifier is a Lane 5 change and is **not** proposed here.

---

## The verifier — and the two bugs the audit found *in it*

It had itself been reading `cohort.cohortCoveragePct` (always absent → would have reported `BLOCKED` for the wrong reason on a perfectly healthy board) and the nonexistent `/api/status` SHA fields. Both fixed.

Five states, three of which are **not** passes:

| status | meaning |
|---|---|
| `pass` | the case arose and behaved correctly |
| `fail` | the case arose and behaved incorrectly |
| `unmeasurable` | the input was read; the case did not arise |
| `blocked` | a required input does not exist here |
| `unverifiable_unauthenticated` | 401/403 |

### Two structural guards against being green while inspecting nothing

**Denominators.** Every population check publishes what it actually inspected, and `Check.finalize` downgrades a `pass` whose denominator is `0` to `unmeasurable` — centrally, so one forgotten guard at a call site cannot produce a false green.

**Exit `0` is reserved for a COMPLETE run.** Any blocked, unmeasurable or unauthenticated check caps the run at exit `3`, however many others passed. Consequence stated rather than hidden: **until production evidence exists this exits `3` on every host**, which is the accurate report.

Route discovery is a source-level **AST scan**, not an app import — `import server` refuses without `JASON_LOGIN_PASSWORD` (correctly; it is fail-closed auth), so an import-only check reports `BLOCKED` on every machine that is *not* the deployed box, which is exactly where a stale route name needs catching. The live app is used as confirmation when importable.

---

## Tests — 55 deterministic, mutation-proven where it matters

Proving the verifier detects each thing you named:

| detects | test |
|---|---|
| a wrong endpoint | the real `/api/faab/recommend` → `FAIL`; plus the **converse** so it cannot pass by matching nothing, and a wrong-**method** case |
| a renamed/missing field | missing path → `FAIL`; and a field nested under the **wrong parent** (the exact `cohort` vs `transparency` defect) |
| HTTP 401 | raises the dedicated `Unauthenticated` type; plus **no-cookie-means-no-cookie-header**, structurally |
| an empty population | `BLOCKED`, never `pass` |
| a measured zero vs missing | `0.0` with voters passes; `null` with voters **fails** (the repair overshooting) |
| stale evidence | parametrised over `stale`/`missing`/`""`/`unknown`; a positive TEP claim under any of them fails |
| zero-voter Sharp quality | `1.0` → `FAIL`, `null` → `PASS` |
| card-derived TEP state | rule read from the **signature**; plus the measured no-edge and edge cards through the canonical owner |
| IDP population refusal | offense-only refuses DL/LB/DB while still pricing QB/RB/WR/TE; an IDP league present → `unmeasurable`, not `fail` |

**Three assert against the REAL producers at HEAD**, so a payload reshape breaks CI rather than the operator's evening.

**Mutation-proven, not merely written:** the vacuous-pass downgrade, the parametrised exit-code caps, and a scanner over **this document** that fails if it names an unregistered route — verified to go red by appending a bogus route to the real file and green again on restore. **That guard is what would have caught all four defects before they shipped.**

Presence is tested, never truthiness: `cohortCoveragePct: null` is the *correct* unmeasured state, so a truthiness test would report right behaviour as a missing field.

---

## Still impossible without production authentication

Recorded as `unverifiable_unauthenticated` — **not pass, not failure**. A live `--mode remote` run against `chaseupside.com` returned **3 × 401** just now, which is the wall itself, measured.

* **V1-58** Sharp cohort populated · **V1-59** Sharp bootstrap — need an authenticated `/api/sharp/cohort` read and `journalctl` on the 04:20 → 04:50 → 05:50 chain.
* **C1–C3** person-consensus states, **C8/C9** crowd refusal + bid effect — need a session.
* **C4–C7, C10** — need the scoring card and crowd ledger, both gitignored and prod-only; `--mode onbox` only.
* **V1-57** timer — needs `systemctl` on the box.

`/api/sharp/*` is not in the public allowlist and **that gate must not be relaxed to make verification easier**. No cohort, ledger, scoring card or crowd row is fabricated anywhere in this PR.

## Verification

| check | result |
|---|---|
| `tests/ops` | 55 passed |
| `tests/ops` + `tests/sharp` + `tests/trade` | **890 passed** |
| `ruff check` / `format --check` | clean, 1170 files |
| diff scope | nothing outside `scripts/verify_lane4_production.py`, `tests/ops/`, `docs/lane4/` |

**Supersedes #931**, which targeted `main` and is being closed in favour of this stacked one.

---

### DIFF
--- a/docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md
+++ b/docs/lane4/L2_L3_VERIFICATION_PROCEDURES.md
@@ -30,18 +30,88 @@ Levels, quoted from the contract §2 so this file cannot drift from it:
 
 ---
 
+## Corrected 2026-08-19, and now executable
+
+This document shipped with **four wrong references**, found by deriving every
+endpoint and field from the actual route registrations and response producers
+rather than from memory. Each is corrected in place below and called out where
+it occurs, because a superseding document leaves the wrong one in circulation.
+
+| # | this document said | reality |
+|---|---|---|
+| 1 | `POST /api/faab/recommend` (×2) | **`POST /api/waiver/faab-recommend`** — there is no `/api/faab/*` prefix at all, so **every crowd-market step was unrunnable** |
+| 2 | roster-percentage → `sources.ffpc.status` | **`coverage.platforms.ffpc.status`**, and it is on **`/api/sharp/market`** — wrong field *and* wrong endpoint |
+| 3 | `cohortCoveragePct` (under `cohort`) | **`transparency.cohortCoveragePct`** |
+| 4 | roster-percentage → `rostersObserved` | **does not exist**; the count is `transparency.eligibleRosters` (also `sample.eligibleRosters`) |
+
+Correct and left alone: `cohort.selectedManagers` (present on **both** boards),
+`crowdMarket.{state,refusalReason,excludedCounts,tierCounts,pricesIdp,rowsUsed,rowsTotal,targetFormatUnknown,playerHasEvidence}`,
+`contention.notes`, and `CollectResult.status`. Two further values were wrong
+and are fixed at their rows: the `/api/status` SHA fields (next section) and the
+`unavailable_reason` literals (V1-60).
+
+**The reason this happened is worth keeping.** A prose checklist's paths are
+checked by a reader's goodwill; nothing fails when one is wrong, and an
+unrunnable procedure reads exactly like a runnable one until someone with
+production credentials wastes an evening on it.
+
+So the checklist is no longer the only artifact. **`scripts/verify_lane4_production.py`**
+resolves every route and field against the deployed code at run time, and
+`tests/ops/test_lane4_verification.py` asserts — against the real producers, at
+every CI run — that each route this package names is registered and each field
+path exists. One of those tests scans **this document** and fails if it names a
+route that does not exist. That is what would have caught all four.
+
+The verifier reports five states, and three of them are not passes:
+
+| status | meaning |
+|---|---|
+| `pass` | the case arose and behaved correctly |
+| `fail` | the case arose and behaved incorrectly |
+| `unmeasurable` | the input was read; the case did not arise |
+| `blocked` | a required input does not exist here |
+| `unverifiable_unauthenticated` | 401/403 — insufficient evidence |
+
+**Exit `0` is reserved for a COMPLETE run.** Any blocked, unmeasurable or
+unauthenticated check caps the run at exit `3`, however many others passed, and
+a `pass` whose denominator is `0` is downgraded automatically — a check that
+inspected nothing cannot report success.
+
+---
+
 ## Deployed-SHA preamble (every L3 procedure)
 
-An L3 result is only meaningful against a known commit. Record this first; if
-the SHA does not match the head being claimed, stop — the run proves something
-about a different tree.
+An L3 result is only meaningful against a known commit.
+
+> **CORRECTED 2026-08-19 — the deployed SHA is NOT observable over HTTP.**
+> This section previously said to record a `commit` field and a `startedAt`
+> field from `/api/status`. **Neither exists.** Verified against the producer
+> (`server.py::get_status` + `_scrape_status_payload`): the payload carries
+> `contract`, `data_runtime`, `health`, `sources`, `source_failures` and the
+> payload-size block, and its `contract.version` is the **API data-contract**
+> version (e.g. `2026-03-10.v2`) — the payload's shape, not the commit that
+> produced it. **No endpoint publishes a git SHA or build identifier.**
+
+So an L3 run's tree identity must come from the box, not the API:
 
 ```bash
+# ON THE BOX — the only place the SHA is knowable
+git -C "$APP_DIR" rev-parse HEAD
+git -C "$APP_DIR" log -1 --format='%H %ci %s'
+
+# from anywhere — useful context, but NOT a tree identity
 curl -s https://chaseupside.com/api/status | python -m json.tool | head -40
 ```
 
-Record: `commit`/`version` field, `startedAt`, and the wall-clock time of the
-call. Every artifact below is filed under that SHA.
+Record: the `git rev-parse HEAD` output, `contract.version`,
+`data_runtime.last_data_refresh_at`, and the wall-clock time of the call. Every
+artifact below is filed under that SHA.
+
+**`scripts/verify_lane4_production.py` encodes this gap rather than papering
+over it:** in `--mode remote` it stamps `headSha: null` with
+`headShaSource: "unavailable_over_http"`, and only `--mode onbox` can answer it.
+Closing the gap properly means publishing a build identifier, which is a Lane 5
+change and is not proposed here.
 
 ---
 
@@ -72,11 +142,11 @@ ls -la /opt/dynasty/data/faab/          # path per the deployed APP_DIR
 
 | assertion | pass |
 |---|---|
-| the timer is installed and enabled | `LOAD=loaded`, `ACTIVE=active`, `UNIT=dynasty-faab-history.timer` in `list-timers` |
+| the timer is installed and enabled | `LOAD=loaded`, `ACTIVE=active`, and a `list-timers` row whose `UNIT` ends `-faab-history.timer`. The prefix is `$SERVICE_NAME` from `install-systemd-service.sh` (`install_simple_timer` builds `${SERVICE_NAME}-${stem}`), so match on the **suffix** rather than assuming `dynasty-` |
 | it has actually fired | `LAST` is populated and within ~25 h of now |
 | the run succeeded | the most recent journal entry exits **0** (1 = registry unreadable, 2 = nothing fetched) |
 | it produced the artifact | `data/faab/bid_history_<leagueKey>.json` exists per active league, `mtime` inside 25 h |
-| the artifact is used | `POST /api/faab/recommend` → `contention.notes` does **not** carry the configured-priors fallback note |
+| the artifact is used | `POST /api/waiver/faab-recommend` → `contention.notes` does **not** carry the configured-priors fallback note |
 
 **Exit 2 is not a pass and not a failure.** It means nothing was fetched and the
 previous file was left untouched. Record it as `DEGRADED` with the prior file's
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
**CI green at exact head `e7e8c1893`** — [Validate PR](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32288841269/job/96184712344) completed `success` (18:42:26 → 18:58:55).

The blocking gate is the one that matters here, and it passed with the 55 new tests inside it:

| step | |
|---|---|
| format / lint (ruff, whole repo) | ✅ |
| decision-path coercion gate | ✅ |
| audit finding-status drift · planning integrity · product-plan governance | ✅ |
| Python import gate | ✅ |
| **Run unit tests (hard gate — pure logic)** | ✅ 13 min |
| API data contract (structural lane — blocking) | ✅ |
| frontend vitest + build + bundle budgets | ✅ |

**Stack integrity confirmed:** #927's head is still `07482b9b8`, which is exactly this PR's recorded base — so the diff Claude 5 reviews is the diff CI validated, and no rebase is owed.

One thing worth knowing before merge, because it affects how the new tests behave on *other* branches: three of them assert against the **real** producers (`build_board`, `market_payload`, `CrowdMarket.to_dict`) rather than fixtures. I checked they cannot be flaky on a runner with a different `data/` — each of those producers has a single return of one dict literal, so every asserted key is present regardless of whether any data exists. They go red only if a payload is genuinely reshaped, which is what they are for.

**Not merging this myself.** Order: #927 → this, or absorbed into the same bounded RC if that yields the identical tree.

---

jasonleetucker-code:
## Final adversarial review of the combined tree — two false-green paths found and closed

The question asked was not "can the verifier be improved" but **"is there any remaining path by which Integration could claim production evidence without observing the capability?"** There were two. Both are now closed and mutation-proven in both directions. Head moves `e7e8c1893` → **`c795381ba`**; same four files, still verification-only.

### Finding 1 — C4 could report **PASS** having observed no scoring card *(live false green)*

`league_registry._scoring_evidence` decides freshness from the snapshot's fetch timestamp and season and **never reads `scoringSettings`**. So a snapshot written by a partial fetch is `fresh` while carrying no card at all. In that state:

* `_tep_from_scoring` correctly returns `None` — fail-closed, so **the product is fine**;
* the served value then equalled the derived value *trivially* — both `None`;
* C4 fell through to **`pass`**, with the detail *"served tep=None, derived from the fresh card"*.

That is the check whose entire job is to prove TEP is card-derived, reporting success having derived nothing. Integration could have recorded the capability as observed without a card ever existing.

Now **`BLOCKED`**, naming the missing input and the refetch. `BLOCKED` and not `FAIL` because nothing is broken — the evidence is absent.

**C6 already blocked on an absent card.** That asymmetry is what marks C4's omission an oversight rather than a decision, and a test now holds the two consistent.

### Finding 2 — emptying `REQUIRED_ROUTES` disarmed the route guard with **zero** test failures

The *"a verifier test that itself matches nothing"* case, proven by mutation: with the list emptied, **all 55 tests still passed**.

At runtime `Check.finalize` downgrades the resulting zero-denominator pass to `unmeasurable`, so the verifier itself stayed honest — but **CI would not have told anyone the guard was disarmed**. `REQUIRED_FIELDS` was already covered by two tests; this closes the asymmetry, and names each critical route and field **individually**, so dropping *one* — which a non-empty check alone would not catch — fails too.

### Mutation proof, both directions

| mutation | result |
|---|---|
| revert the C4 repair | **2 fail** |
| empty `REQUIRED_ROUTES` | **6 fail** |
| drop only `POST /api/waiver/faab-recommend` | **1 fail** |
| each converse (real card, real routes) | still **pass** — neither guard can be green by refusing everything |

### The other 44 probes: no defect

Wrong route · wrong HTTP method · missing field · correct field under the wrong parent · a producer renaming a field · 401 · 403 · zero-row populations · measured `0.0` vs `null` · stale scoring evidence · missing scoring evidence · a cached bid-history artifact at 26 h and 30 d · a never-fired timer · absent `systemctl` · empty Sharp cohort · empty crowd ledger · offense-only population asked to price IDP · **and a server inventing a `commit` field it does not really have** (the verifier still stamps `headSha: null` / `headShaSource: "unavailable_over_http"` rather than opportunistically trusting it).

All 44 behave as intended.

### Not changed, deliberately

No product source. No `/api/sharp/*` allowlist change. No build-SHA endpoint — that is Integration/platform ownership, and #932 only *states* the provenance contract. No fabricated cohort, ledger, scoring card or roster. No V1 row promoted.

`tests/ops` 70 passed (was 55) · `tests/ops` + `tests/sharp` + `tests/trade` **905 passed** · ruff clean · coercion gate clean · diff scope unchanged.

---

jasonleetucker-code:
**CI green at `c795381ba`** — [Validate PR](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32297933975/job/96213566034) completed `success` (20:19:52 → 20:33:19), with the two false-green repairs and the 70-test suite in the blocking gate.

**The train is now frozen for Integration.** I will make no further changes to #927 or #932 unless a concrete defect is returned.

| | |
|---|---|
| **#927 head** | `07482b9b8` — CI green, `mergeable_state: clean` |
| **#932 head** | `c795381ba` — CI green, base still `07482b9b8` |
| **combined tree** | semantically the intended stack; diff is the same four verification files |
| **order** | #927 → #932, or one identical combined RC |

**Verifier exit semantics:** `0` only when every check measured its case and passed · `1` error · `2` failure · **`3` incomplete** — any `blocked` / `unmeasurable` / `unverifiable_unauthenticated` caps the run there regardless of how many passed, and a `pass` with denominator `0` is downgraded automatically. **Until production evidence exists it exits `3` on every host.** That is the accurate report, not a failure.

**No V1 row is promoted by either PR.** 401/403 stays `UNVERIFIABLE_UNAUTHENTICATED` — neither pass nor fail — `/api/sharp/*` stays session-gated, and no cohort, ledger, scoring card or roster is fabricated anywhere in this tree.

**Deployed-SHA provenance is specified, not owned here:** either on-box `git rev-parse HEAD` in the deployed working directory, or an Integration-owned authenticated build identifier. `/api/status` cannot answer it and `contract.version` is not a substitute.

---

jasonleetucker-code:
## Lane 4 frozen — current-main compatibility recorded

Full overnight handoff is on **#936** ([comment](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/936#issuecomment-5349945452)); this is the #927→#932 half.

**Heads unchanged, nothing rebased, no production code touched:**

* **#927** `07482b9b8f24de24b6ecea2eb1031e34993e3381` — **READY**
* **#932** `c795381ba0e5e962ebdf201e89d4df5cedaa09ea` — **READY**

**Dependency order is fixed: #927 → #932.** #932 is stacked on #927's exact head. Do not merge them independently and do not reorder.

**Against `main@547e51bc6`:** merge tree **`c0f6b933d`**, no conflict · **1100 pass** across `tests/ops` + `tests/sharp` + `tests/trade` + the FAAB endpoint suite · ruff clean · coercion gate clean.

**All five named semantic-collision surfaces clean.** `league_registry.py` / `te_premium.py` untouched · `src/sharp/` untouched · `faab_comparability.py` / `faab_history.py` / `faab_engine.py` / `config/trade/faab.json` untouched · main added `/api/roster/intelligence` and removed no route, with R0 `pass` (5/5 required routes, 105 discovered) and R1 `pass` (8/8 field paths).

Textual overlap is only `CLAUDE.md` and `server.py`, both examined: main's `CLAUDE.md` edits sit in other sections and the TEP block survives; our single `TargetFormat.from_league_config` call site survives at `server.py:6130`. All five semantic markers present on the merge tree — TEP-is-a-fact, zero-voter `None`, `target_format_unverifiable`, the C4 fresh-card repair, and the corrected `/api/waiver/faab-recommend`.

**⚠️ Integration must run the authoritative RC in a normal checkout with no test deselection.** My local `tests/consensus_edge/test_panel.py` deselection is a linked-worktree artifact only — `panel.is_shallow()` reads `.git/shallow`, which cannot exist when `.git` is a worktree *file*, so the `@_deep_only` skip never fires. A real checkout skips those 13.

**If this train lands:** #927 makes **V1-63** (L1) a verification candidate — its stated blocker is discharged with RED-before-GREEN. #932 promotes no row directly; it corrects the procedures document and ships the verifier the L2/L3 rows are collected with. Everything else in Lane 4 stays blocked on production, auth or on-box evidence, and **I have promoted no row.**

Frozen here. Available only for a specific Integration repair on this train.

---

PR #935, opened 2026-08-19
/rosters consumes canonical Team Strength instead of scoring teams itself

jasonleetucker-code:
> **Small, separate, and deliberately NOT stacked on #912.** #912 is byte-frozen at
> `c025e4f68` and unchanged by this branch. Branched from current `main` (`45c6101`).
> Claude 5 remains merge authority.

`/rosters` is titled **Team Strength** and answered that question twice, in two
different ways, on one screen:

| surface | quantity | source |
|---|---|---|
| power table `#` | portfolio value under the current asset scope + position filter | `buildAllTeamSummaries` — a client-side sum |
| tier card `#rank` | `0.7 × starterValue + 0.2 × depthValue − 0.1 × pickValue`, cut into contender / mid-tier / rebuilder thirds | `lib/league-analysis.js::scoreTeamTiers` — a client-side **formula** |

The second is a Team Strength methodology living in the browser. CLAUDE.md forbids
it in as many words — *"There is no frontend ranking engine, period — not even a
fallback"* — and `V1-35` is the owner decision that the named team quantities
*"may not collapse into one team score"*.
`docs/master-site-audit/PRIOR_AUDIT_RECONCILIATION.md:408` measured the two ranks
disagreeing for **10 of 12 teams**: one team was `#1` in the table and `#5
"Mid-Tier"` four hundred pixels below.

## What now owns the number

`src/roster_intel/strength.py` (feature-inventory row 1.1), served by
`GET /api/roster/intelligence` — **#914, already on `main`**.
`components/TeamStrengthCard.jsx` renders `strength.total`, `leagueRank`, the
starter/reserve split, `byPosition[]` with per-group ranks, and the league ladder
from `leagueContext[]`. Every number is read verbatim; `lib/roster-intelligence.js`
classifies, reshapes and formats, and **sums nothing**.

`scoreTeamTiers` is **deleted, not deprecated** — an unused export is a working
second engine one import away, which is how the first one acquired its callers.
Its eight regression pins go with it, because they pinned the coefficients; what
replaces their coverage is named in the file where they were.

## Three absences stay absences

The backend went to some trouble to distinguish these, and a frontend that
flattens them undoes the work:

* **`available: false`** renders the backend's reason plus the sentence *"this is
  not a strength of zero"* — the lineup could not be read, which is a different
  fact from a worthless roster.
* **`leagueRank: null`** renders **Not measured**, in the position the server put
  it. The ladder preserves `_league_context_order` rather than re-sorting, because
[...]

PR #938, opened 2026-08-20
E2E stabilization: repair the four inherited safety-net failure families

jasonleetucker-code:
## Why

PR #935's E2E Safety Net run ([32303014699](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32303014699)) went red with 5 failures across four families — but **four of the five already fail identically on `main` itself**: the nightly scheduled run [32226512875](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32226512875) (head `4c208e5b`, ~14h before #935 was opened, tracked in #924) shows the same settings-overrides failure + 180s retry hang, the same 2-vs-21 source-health failure, and the same teamAssignment failures on both projects. #935 touches none of the involved surfaces (its only E2E spec change is signed-in-smoke, which passed). These are inherited suite defects, so they are repaired here on `main`, unblocking every PR — not hidden inside #935.

Every repaired test was shown RED before (CI + local baseline) and GREEN after, and every new assertion was mutation-tested: a deliberate defect was injected at the owner, the assertion was observed failing with its own message, and the mutation was reverted (details per family below). No test was skipped, weakened, or timeout-inflated.

## Family 1 — settings source overrides (STALE_SPEC + TEST_SYNCHRONIZATION_DEFECT)

**Why it failed:** the spec still asserted the custom-mix badge appears after disabling a source. #875 (2026-08-14) withdrew custom source weighting (`_SOURCE_OVERRIDES_DISABLED = True` in `src/api/data_contract.py`) because a user-weighted board published under `rankDerivedValue` is a second canonical valuation truth. `rankingsOverride.isCustomized` can no longer become true from a source toggle, so the badge (gated on it) never mounts. The spec predates the withdrawal (last touched in #733, 08-05).

**The 180s retry hang** at `await overridesResponse.json()`: Playwright's `.json()` waits for the browser to finish loading the response body. The app drains the ~4.7 MB delta body only after its base-contract fetch resolves (`dynasty-data.js::_postOverridesAndMerge`); when it falls through to the base contract, the body is never consumed and `.json()` — which has no timeout of its own — burns the whole 180s budget.

**Repair (tests/e2e/specs/journey-settings-overrides.spec.js):** the test now asserts the *withdrawal contract* (assertion direction validated against open PR #912's repair of the same spec, re-verified against current `main` behavior): the toggle still fires the POST and gets 200 (integration proof unchanged); the response `warnings` must state `"custom source weighting is disabled"`; `rankingsOverride.isCustomized === false`; the toggled-off source stays in `enabledSources`; the badge must be **absent**. The delta is read by replaying the captured `postData()` through the API request context (`page.request.post`) instead of `.json()` — hang-proof, byte-identical body, same memo slot. Also replaced a vacuous assertion (`delta.players || delta.playersDelta || delta` — neither key exists on the delta shape) with real shape assertions on `rankingsDelta.players`.

**Mutation proof:** `_SOURCE_OVERRIDES_DISABLED = False` → RED on the withdrawal-warning assertion; suppressing only the `warnings.append` → RED on the warnings assertion alone. Both reverted.

**Endpoint measurement + narrow fix (server.py, separable commit):** the overrides memo keyed on sha1(raw posted body), so with the withdrawal in place every *distinct* toggle body paid a full pipeline rebuild (~1.2s warmed, tens of seconds cold) for a byte-identical response. The key now derives from the normalized inputs the build actually consumes (override map, tep knobs, valuation mode, warnings) — measured after: first body per contract generation builds once, every subsequent distinct toggle body is a ~7ms memo hit. Response-equivalence-complete even if the flag is ever flipped back (warnings ride the key). New pin in `tests/api/test_overrides_response_cache.py` (`test_withdrawn_source_toggle_bodies_share_one_slot`) — written first and observed failing against the old key, green against the new; the five pre-existing cache pins pass unchanged.

## Family 2 — source health 2 vs 21 (STALE_SPEC)

**Why it failed:** F-7 (2026-08-18) changed `SourceHealthStrip` to render `source_health.registered_sources` (the 21-key ranking-source registry — the population the page's own subtitle promises) instead of `source_runtime.enabled_sources` (the scraper's 2 anchor-market run names). The spec, unchanged since #579 (07-27), still derived its expectation from the runtime list: 21 !== 2, and every rendered name failed containment against `["IDPTradeCalc","KTC"]`.

**Repair (tests/e2e/specs/journey-tools-health.spec.js):** the expected population is now `registered_sources`, hard-asserted non-empty (the registry is compiled in — empty means the backend broke), and cross-checked for identity against `GET /api/rankings/sources` so the expectation is contract-derived, never a magic count. The row assertion is **set equality** (sorted rendered names deep-equal the registry) — catching missing, extra, duplicated, and invented rows, where the old one-directional containment could not. Added: per-row status vocabulary pin (`not measured` / `no rows` / `N rows` — missing data can never render as a healthy count) and `missing_sources` surfacing in both directions.

**Mutation proof:** rendering `registered.slice(1)` → RED (20 vs 21); rendering `slice(1) + "bogusSource"` (count still 21, identity wrong) → RED on set equality, naming `bogusSource` — a defect the old count assertion structurally could not see. Both reverted.

## Family 3 — public-league teamAssignment (STALE_SPEC: a deliberate red in a blocking gate, on live nondeterministic data)

**Why it failed:** the test demanded `nflTeams.length > 0` for every manager — a documented deliberate red against defect #815 (`docs/VERSION_1_COMPLETION_CONTRACT.md` V1-94). But an empty `nflTeams` is a pinned-legitimate producer state (`tests/api/test_team_assignment_availability.py::test_healthy_empty_result_is_still_expressible`), and in CI the section is built from a **cold live Sleeper fetch** (`data/public_league/` is never seeded), so the failing shape moved between runs: run 140 got `assignments: []` (failing the ≥8 floor), run 142 got 12 slots with ≥1 empty `nflTeams`. On today's live data the old test passes — the failures were transient live-data states, which is exactly what a deliberate red sitting on live data does inside a blocking gate: it flaps.

**Repair (tests/e2e/specs/public-league.spec.js):** the test now pins the availability/no-fabrication contract the producer actually guarantees: `available:false` must name a machine-readable reason (`no_current_season`/`no_rosters`) with no assignments and `rosterScoringAvailable:false`; `available:true` must cover exactly the current managers (ownerId set equality against the same response's `league.managers[].currentRosterId` — no drops, duplicates, or inventions, with the ≥8 floor retained as a vacuity guard); a resolved favorite must be preserved and lead the list; nothing may appear in `nflTeams` without being the favorite or clearing `assignmentMinPoints`; `rosterScoringAvailable:false` must be explained in `degradedReasons` and reflected on every card. Empty-with-explanation stays legal — no NFL team is ever fabricated to satisfy a test. The stale "deliberate red" sentence in `VERSION_1_COMPLETION_CONTRACT.md` V1-94 is updated (status fields untouched); #815 stays tracked at the unit level.

**Mutation proof (all at the producer, all reverted):** dropping the first roster → RED on the ownerId set equality; skipping favorite emission → RED on "the favorite must lead the list"; returning `_unavailable(config, "bogus_reason")` → RED on reason membership.

**No config alias was added:** current live data resolves favorites for all 12 managers, so there is no rename to repair — the run-142 empty manager was a transient state the contract now represents honestly.

## Family 4 — mobile ?tab=awards deep link (TEST_SYNCHRONIZATION_DEFECT)

**Why it failed (run 142 only, mobile only):** `visitLeague`'s 15s needle wait silently shared its window with an **unbounded** client-side section fetch. The awards tab is SSR-prefetched behind a 3s-bounded backend fetch; on a miss the client lazy-loads it behind "Loading section..." with retries and no timeout. The needle (`"award"`, lowercase — awards-section copy only; nav labels are capital-A) never had a bounded, section-loaded page to look at. Not reproducible locally (5/5 green at baseline on a warm snapshot), consistent with a CI contention/cold-fetch window.

**Repair (public-league.spec.js):** `visitLeague` now waits (bounded, 30s) for "Loading section..." to clear before the needle, and fails fast with a named error when the page renders "Section unavailable" — so a section-fetch failure is diagnosed instead of timing out anonymously. The deep-link test additionally asserts the routing half on the control each viewport actually uses: the mobile `<select>` must have value `awards`, the desktop active tab button must be "Awards". No timeout was raised; the needle and its 15s budget are unchanged.

**Mutation proof (all reverted):** pointing `SECTION_FOR_TAB.awards` at `records` → RED at the needle (reproducing the CI failure signature exactly — corroborates the diagnosis); aliasing `awards → history` in `normalizeTabKey` → RED; corrupting only the select's value while content stays correct → RED on the new `toHaveValue("awards")` assertion (Expected "awards" / Received "overview"), proving the routing assertion's independent value.

**Determinism evidence:** 6/6 green on mobile-chromium (3 fresh invocations + `--repeat-each=3`) after the repair. **Confirmed in CI**: in run 32316532286 this case took **30.9s on mobile** (vs 5.6s desktop) and passed — direct evidence that the bounded section wait covered a real ~30s CI window that would have blown the old 15s budget.

## Family 5 — CI readiness gap exposed by run 32316532286 (PERFORMANCE/STARVATION, pre-existing)

After the four repairs above went green on the merge tree, that run still failed on a **fifth, separate** test — `journey-trade.spec.js:96`, which I had not touched. It is not a synchronization defect, and the test was right to fail. The failure-time page snapshot from the artifact shows the board rendered its terminal error state:

> `Rankings unavailable — Failed to load dynasty data: 503 {"ok":false,"error":"backend unreachable or stalled","reason":"backend_idle_timeout","timeoutMs":4000}`

The backend log for that exact window explains it. BDVM is flag-on by default, and the first `GET /api/bdvm/values` lazily ingests the nflverse corpus **inside the FastAPI process**: ~112k weekly-stat rows (00:17:05–00:17:39), ~158k snap-count rows (00:18:12–00:18:30), then schedules — network plus CPU on a 2-core runner, with the event loop starved throughout. `frontend/app/api/dynasty-data/route.js` aborts a response after a 4s **inter-chunk** gap (`BACKEND_IDLE_TIMEOUT_MS`), so a journey loading the ~4.7 MB contract inside that window gets a 503 and an empty board. The test failed at ~00:19, mid-ingestion.

This is pre-existing and independent of this PR's diff: `journey-trade.spec.js:96` never calls the endpoint whose memo key changed (the overrides POST fires only for a customized `localStorage` mix, and every test gets a fresh context). What this PR changed is *scheduling* — the old settings-overrides test burned ~4 minutes idling (60s badge wait + 180s hang), which incidentally pushed journey-trade past the ingestion window. Repairing it removed that accidental cooldown and exposed the latent gap.

**Repair (.github/workflows/e2e.yml):** the readiness gate answered "ready" on `has_data:true` + a frontend 200 while a multi-minute lazy ingestion was still pending. It now forces that ingestion before Playwright starts, then polls **the exact path the app uses** (`http://127.0.0.1:3000/api/dynasty-data?view=app`) until it actually serves the contract. This moves the work out of the journey window rather than adding it, and makes "ready" mean ready. No timeout was raised, no assertion touched, no test skipped — `journey-trade.spec.js` is unmodified.

**Non-vacuity:** with the backend stopped, the poll observes 503 and the gate exits 1 with its `::error` annotation; with the backend up it passes on attempt 1. The BDVM warm is logged per league (`dynasty_main -> 200`, `dynasty_new -> 503`, the latter being the documented non-loaded-league answer) and deliberately does not gate on that code — the bridge poll is what gates.

## Validation

- Full E2E suite locally under CI-equivalent settings (`CI=1 SKIP_VISUAL_REGRESSION=1 npm run e2e`, projects desktop-1366 + mobile-chromium, `--ignore-snapshots`, workers=1, retries=1, failOnFlakyTests): **154 passed / 52 skipped / 0 failed / 0 flaky (exit 0)** — 154 = the 149 that passed in run 142 plus the 5 repaired.
- Python gates: `test_overrides_response_cache` (6), `test_source_overrides`, `test_team_assignment` (31), `test_team_assignment_availability`, `test_canonical_ownership_protections`, `test_e2e_harness_guards` — all green (153 targeted tests).
- Every temporary mutation was verified applied (observed RED), then reverted; `git status` shows only the intended files.
- GitHub `Validate PR` on the first head (`9660c616`): **success** (run 32316532448). All four repaired families passed the E2E gate on that head; only the Family-5 flake remained, which the readiness fix addresses.

## Deliberately out of scope (reported, not fixed)

- `frontend/app/settings/page.jsx:679` copy still promises the withdrawn custom mix (belongs with #912's UI work).
- CI serves `/api/public/league/*` from a cold live Sleeper rebuild, contradicting `e2e.yml`'s "no external network" header — the durable fix is a committed/seeded public snapshot fixture.
- BDVM's lazy nflverse ingestion runs inside the request-serving process and can starve the event loop for minutes. This PR moves it out of the journey window; making it non-blocking (a background warm task, or an out-of-process build) is the real product fix.
- The Next `/api/status` proxy's 3s abort falls back to a strip render without the `aria-label` the spec locates (covered today by the spec's existing 90s refresh-cycle wait).
- `SKIP_VISUAL_REGRESSION` redundancy with `--ignore-snapshots` (already filed in docs/master-site-audit).

https://claude.ai/code/session_019UGkSY3XKvYBjovxkDHjsj


jasonleetucker-code (commented):
Independent E2E audit of run 32316532286: the four failure families this PR was created to repair are now green on the merge tree — settings overrides, source-health, public-league deep links (desktop + mobile), and teamAssignment (desktop + mobile). Do not reopen those fixes.

The suite is still red for one separate failure: `tests/e2e/specs/journey-trade.spec.js:96` (`a /rankings screen deep-link narrows the board to its question`) timed out once for 60s waiting for `.rankings-row-clickable`, then passed on retry in 3.9s. The harness correctly treats retried green as failure. This is now the only observed E2E blocker in run 32316532286.

SEND BACK is bounded to that synchronization/deep-link failure only. Inspect the first-attempt trace/artifact from run 32316532286 before changing anything. Do not increase the timeout, add `.first()` as a masking workaround, skip the case, or weaken the row assertion. Determine the deterministic readiness condition for the screened rankings board, prove RED-before/GREEN-after (or repeated non-flake evidence if the test itself is already correct), then rerun the full E2E gate on the exact head. PR Validation is still in progress at the time of this review.

jasonleetucker-code:
Addressed the send-back. New head: **`1af1d2b3`**. I did not reopen the four repaired families, and I did not touch `journey-trade.spec.js`.

**It is not a synchronization defect, and the test is already correct.** I pulled the first-attempt artifact from run 32316532286 before changing anything, as asked. The page snapshot at failure shows the board in its terminal error state:

> `Rankings unavailable — Failed to load dynasty data: 503 {"ok":false,"error":"backend unreachable or stalled","reason":"backend_idle_timeout","timeoutMs":4000}`

So the board genuinely had zero rows. No readiness condition on the test side could have helped — the page had already reached a terminal state, and the 60s wait was spent on a locator that could never resolve. Waiting differently, or longer, would only have failed more slowly.

**Root cause, from the backend log for that exact window.** BDVM is flag-on by default, and the first `GET /api/bdvm/values` lazily ingests the nflverse corpus *inside* the FastAPI process:

- `00:17:05–00:17:39` — weekly stats, ~112k rows
- `00:18:12–00:18:30` — snap counts, ~158k rows
- `00:18:54–00:18:57` — schedules

That is minutes of network + CPU on a 2-core runner with the event loop starved throughout. `frontend/app/api/dynasty-data/route.js` aborts a response after a 4s **inter-chunk** gap (`BACKEND_IDLE_TIMEOUT_MS`), so any journey loading the ~4.7 MB contract inside that window gets a 503 and an empty board. The test failed at ~00:19, mid-ingestion.

**On causation, stated plainly:** this is pre-existing and not caused by this PR's diff — `journey-trade.spec.js:96` never calls the endpoint whose memo key I changed (the overrides POST fires only for a customized `localStorage` mix, and every test gets a fresh context). But this PR *did* change the scheduling that exposed it: the old settings-overrides test burned ~4 minutes idling (60s badge wait + 180s hang), which incidentally pushed journey-trade past the ingestion window. Repairing it removed that accidental cooldown. I'd rather say that than call it someone else's flake.

**The deterministic readiness condition is in the harness, not the test.** The gate answered "ready" on `has_data:true` + a frontend 200 while a multi-minute lazy ingestion was still pending. `.github/workflows/e2e.yml` now forces that ingestion before Playwright starts, then polls **the exact path the app uses** (`:3000/api/dynasty-data?view=app`) until it actually serves the contract. It moves the work out of the journey window rather than adding it. No timeout raised, no `.first()`, nothing skipped, no assertion weakened.

**Non-vacuity of the new gate:** with the backend stopped the poll observes 503 on every attempt and the gate exits 1 with its `::error` annotation; with the backend up it passes on attempt 1. The per-league BDVM warm is logged (`dynasty_main -> 200`, `dynasty_new -> 503`, the documented non-loaded-league answer) and deliberately does not gate on that code — the bridge poll is what gates.

One more datapoint in your favour from that same run: the mobile `?tab=` deep-link case took **30.9s** (vs 5.6s on desktop) and passed. That is the bounded section wait from this PR absorbing a real ~30s CI window that would have blown the old 15s budget.

`Validate PR` was green on the previous head (`9660c616`, run 32316532448). Full E2E gate is running now on `1af1d2b3`; I'll report both conclusions with SHAs and will not call it green without them.

**Left deliberately out of scope and flagged:** BDVM's ingestion running inside the request-serving process is a real product fragility. This PR moves it out of the journey window; making it non-blocking (background warm task, or an out-of-process build) is the actual fix and is bigger than this bounded send-back.

---

jasonleetucker-code:
**READY FOR INTEGRATION.** Both gates green on the exact frozen head — no rebase, no reruns, no post-freeze commits.

| Gate | Run | Job | Conclusion |
|---|---|---|---|
| PR Validation | [32317701796](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32317701796) | 96273396850 | **success** |
| E2E Safety Net | [32317701855](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32317701855) | 96273397497 | **success** |

Exact head: **`1af1d2b3c7b1fd2f6c6cae664571f688e809160f`** — both runs report that `head_sha`, attempt 1, event `pull_request`.

**Zero flaky retries**, verified three ways rather than inferred from the conclusion: the suite reports `154 passed / 52 skipped` in 2.9m with no flaky banner and no failure summary, and the uploaded report artifact contains **1 file** against 27 in the red run (failure screenshots, videos and traces land in `test-results/`, so an empty one means nothing retried).

154 = the 149 that passed in the original red run 32303014699 plus the 5 repaired. All five previously-failing cases pass, including `journey-trade.spec.js:96`.

Two numbers worth carrying into review, because they show the readiness fix removed the contention rather than masking it: the new readiness step took 24s and the suite then ran in 2.9m (against 4.9m starved), and the mobile `?tab=` deep link took **4.9s** here against **30.9s** in the starved run.

Follow-up technical debt, deliberately **not** in this PR: BDVM's lazy nflverse ingestion runs inside the request-serving FastAPI process and starves the event loop for minutes on a cold start, so any journey loading a large payload in that window can fail (the Next data bridge aborts on a 4s inter-chunk gap). Observed hitting `journey-trade.spec.js:96` in CI run 32316532286 and `journey-perfect-draft.spec.js:95` in a local run — which test loses the race is incidental. This PR moves the known BDVM warm out of the journey window; the durable fix is making that ingestion non-blocking (background warm task or out-of-process build).

---

Issue #800, opened 2026-08-11
Trade equalizer suggestions must account for Value Adjustment


jasonleetucker-code:
Owner-requested trade-calculator correctness item.

Problem: when the Trade Calculator suggests an asset/player from a team to make a trade approximately even, the suggestion logic appears to compare raw/pre-adjustment values and ignores the active package/consolidation Value Adjustment. As a result, the suggested player can push the adjusted trade far past even.

Required work:
- reproduce on current HEAD through the real Trade Calculator suggestion/equalizer path;
- identify the canonical owner for the active Value Adjustment / package math rather than duplicating it in the suggestion UI;
- when the KTC-style Value Adjustment lens is active, score candidate add/remove suggestions against the same final adjusted side totals the calculator displays, not raw sums;
- prevent double-applying the adjustment;
- preserve temporary manual value overrides, picks, TEP/IDP semantics, and side symmetry correctly;
- test 1-for-1, 2-for-1, 3-for-1, player+pick, elite consolidation, IDP, TE, and manual-override cases;
- suggestion ranking should minimize the *post-adjustment* gap and should not recommend an asset that makes the adjusted trade materially less balanced than before;
- exact KTC Value Adjustment remains the market-parity/consolidation benchmark and must stay distinct from any future proprietary canonical package methodology.

Do not solve this by hiding the displayed Value Adjustment or by silently replacing KTC parity with a new scalar methodology.

PR #944, opened 2026-08-20
Seven rows VERIFIED on mutation-proven evidence (48 → 55/132, 41.7%)

jasonleetucker-code:
Contract document only — no product source touched.

**55 / 132 = 41.7%.** Denominator unchanged at 132. Green is never treated as evidence that a guard guards: every promotion below was mutated against the real tree, and **every mutation was confirmed APPLIED with the changed line quoted before it counted.**

## V1-81 — freshness budgets, measured twice

**L1:** 23 pass, pinning the *relation* (budget ≤ alert threshold). **MF1**: `"ktc": 6 → 9999` ⇒ RED `test_contract_budget_is_not_laxer_than_the_alert_threshold[ktc]`.

**L2 (a)** Relaxing the five repaired 24h budgets to 720h: **0 bucket / 0 label / 0 value** changes across 1111 rows — reproducing the F-18 audit, and honest (production ages were 1.24–1.62 h).

**L2 (b)** A zero on a fresh board can't tell a working gate from a dead one, so all 27 `_last_success` stamps were aged 1000 h:

| bucket | fresh | all-stale |
|---|---|---|
| `high` | 269 | **24** |
| `medium` | 338 | **0** |
| `low` | 242 | **825** |

**583 of 1111 rows change** — the weakest-axis rule doing exactly what it claims. (A *single* stale source moves 0 buckets; recorded, because that reads backwards — one of 27 families is never pivotal.)

## V1-44 — the equalizer simulates in **both** runtimes

The row was opened because `findBalancers` was divergent across two runtimes, so a single-runtime proof would leave exactly the gap it exists to close. Both mutations reintroduce the *exact* retired #800 rule (rank by raw value against an adjusted target):

- **MV1 (Python)** `suggestions.py:1474` ⇒ REDs incl. `test_overshoot_does_not_hand_the_lead_to_the_other_side` — the 701-flip half.
- **MV2 (JS)** `trade-logic.js:1317` ⇒ **5 REDs**, one named for the defect itself: *"scores against the ADJUSTED gap, not the raw value (defect #800)"*.

Restored: 24 pass (pytest), 179 pass (vitest).

## V1-39 — one resolver, two independently proven rules

Single-resolver claim confirmed by reading both former duplicates: `draft/context.py::_roster_size_for` and `gameplan.py::_roster_limit` both return `roster_capacity.league_roster_limit`.

- **MR-39a** unknown cap → "not over limit" ⇒ RED `test_an_unknown_cap_is_unknown_and_never_unlimited`
- **MR-39c** restore the retired flat `size_before - len(outgoing)` ⇒ 2 REDs on the not-on-roster and partly-held rules

**Latent false-green recorded**: two `pytest.skip` calls survive at `:836`/`:861`. They do *not* fire (0 skips today), but they're conditional on **fixture output**, not availability — the shape that silently stops testing when a fixture drifts.

## V1-114 / 115 / 116 / 118 — governance guards

| row | mutation | result |
|---|---|---|
| **V1-114** | second document carrying the auth heading | `[authorization] … found [… , 'MUTATION_PROBE_AUTH.md']` |
| **V1-114** | heading altered so *nothing* claims it | `[authorization] … found none` |
| **V1-115** | conflicting duplicate `CE-01` | `[ce-unique]` naming both capabilities |
| **V1-115** | `CE-01` renamed | `[ce-mirror]` **twice** — both mirrors |
| **V1-116** | Unexplained unmapped `0 → 7` | `[traceability] … is 7, not 0` |
| **V1-116** | `MANIFEST-ROW-COUNT` `163 → 172` | `[manifest-count] declares 172 but 163 present` |
| **V1-118** | in-scope planning doc classified nowhere | `[governance-index]` naming the file |

**The V1-114 guard proved itself on this PR**: the first draft quoted the auth heading verbatim, making *this contract* a second claimant, and the gate failed the tree before it could ship.

## Not promoted, deliberately

`V1-92` cites `W08-F011`, which the status matrix records as **Missing** — the trade calculator has no freshness display at all. That is genuinely incomplete, not merely unverified, and stays `IN PROGRESS`.

All probe files removed; working tree verified clean after every mutation. `check_planning_integrity.py` exit 0, including the standing-tally gate.


jasonleetucker-code:
TRAFFIC CONTROL — one bounded correction required before merge. The row/table math is 48 → 55 (+7), but the retained narrative immediately below §3.11 still says **“Up five, and every one of them on deployed evidence rather than on a merge.”** That is stale/false for this PR: this PR promotes seven rows, several at L1/L2 via mutation/board evidence, not seven rows on deployed evidence. Please correct only that stale sentence/paragraph so the canonical contract does not contradict its own tally/evidence levels. Do not rework product code or re-chase main. After the correction, preserve the existing feature evidence; Integration only needs the exact-head green gate and the already-proven disjoint-main compatibility check.

jasonleetucker-code:
## Corrected — `ddf42e271`

You were right, and it was worse than a stale sentence: the paragraph sat directly under §3.11, so it read as describing the tally above it. The contract was contradicting its own count *and* its own evidence levels.

**Bounded to that paragraph.** No product code, no re-chase of `main`, no change to any row's status, level or evidence.

### Scoped rather than deleted

The measurements in that block are real (the `V1-127` / `V1-82` / `V1-12` / `V1-120` / `V1-128` production readings), so deleting it would have destroyed evidence to fix a framing error. It is now explicitly labelled:

> **HISTORICAL — the earlier 40 → 45 movement (2026-08-18/19).** … they do **not** describe the tally above. At that time the claim *up five, and every one of them on deployed evidence rather than on a merge* was accurate for those five rows.

### What replaces it

| movement | rows | class |
|---|---|---|
| 48 → 55 | `V1-81` | L2 — board measured |
| | `V1-44`, `V1-39` | L1 — mutation (`V1-44` in **both** runtimes) |
| | `V1-114/115/116/118` | L1 — governance guards mutated against the real tree |
| 45 → 48 | `V1-34`, `V1-130` | L1/L2 — mutation + a 0-value/0-rank board diff |
| | `V1-95` | L2 — **deployed** evidence off `/api/public/league/awards` |

and it now says plainly:

> **Only `V1-95` in this batch rests on deployed evidence** — the rest are deterministic or board-measured, which is what their stated levels require.

The six rows refused in the same pass are named there too (`V1-29`, `V1-51`, `V1-63`, `V1-65`, `V1-88`, `V1-92`), so the paragraph records what did *not* move as well as what did.

### State

`check_planning_integrity.py` exit 0 (including the standing-tally gate); tally unchanged at **55 / 132**. Feature evidence preserved verbatim.

**IMPLEMENTATION HEAD:** `ddf42e271`, awaiting the exact-head green gate.
**RELEASE CANDIDATE:** disjoint-main compatibility already proven — this branch touches only `docs/VERSION_1_COMPLETION_CONTRACT.md`, `scripts/check_planning_integrity.py` and two new `docs/ops/` files, none of which any merged PR has touched since. I'll merge on green rather than re-validating against moving `main`.

---

PR #945, opened 2026-08-20
Lane 7 dispatch records: de-authorise the derived V1 tally, three abandoned refs, Lane 2 idle on undelivered work

jasonleetucker-code:
**Successor to #928** (merged `231468bb4`). Docs only — the three JSON records under `docs/claude-dispatch/`. No source, no behaviour, **no V1 contract edit, no V1 row status change**, no merge.

## Corrected on integration send-back

Integration returned this PR with a bounded correction, right on both halves.

**Staleness (fixed).** The record declared 48/132 while the contract on main already said 55/132 — #944 merged between my read and the push. Re-read at `079fc9e4a`.

**The structural half (fixed properly).** The records carried a `v1` block with `verified` / `denominator` / `percent` fields that *read as authoritative* — which is why they could drift at all. That is a second owner of the V1 numerator forming inside a coordination record: precisely the defect class this dispatcher has spent the session flagging in other lanes. Being the one who warns about it is no protection against doing it.

So the fields are **gone rather than refreshed**. Both records now carry `v1DerivedObservation`, which:

- names `docs/VERSION_1_COMPLETION_CONTRACT.md` §3 as the **sole owner**;
- states it is **NOT AUTHORITATIVE** and **stale by construction**;
- stamps the exact SHA the count was read at;
- says outright that if the two disagree, **the contract wins and this block is what's wrong**;
- tells a reader **not to quote it**.

Projections in the closability queue carry the same label — conditional arithmetic over a derived reading, not statements of status. `ASSIGNMENTS.json` gets the matching note, since assignment rationales cite V1 rows too.

The queue keeps a count because it must rank work by effort-per-row. What it must not do is *look like the answer*, and now it doesn't.

## What the records actually contain

Nineteen dispatch cycles of lane state, plus the V1 closability queue and the assignment ledger. Highlights that outlived their cycle:

- **Method rules earned from my own errors**, each recorded against myself: a hardcoded SHA once reported a cleared blocker as still open; a remembered head SHA was cited after the ref had moved; a merge-detection grep missed a squash merge; and citing test *file names* rather than assertions produced two wrong Tier A calls.
- **`V1-88` refused three times.** It has now survived three of Claude 5's own promotion passes unpromoted. Closing it needs a structured `gates` field on the flag registry or a re-levelling ruling — **not a proxy test**.
- **Correction C17-1 against my own C11-1**: V1-34's level is L1, which does not require a production consumer, so my "wrong by one" warning was itself wrong by one.
- **Three standing items I cannot act on**: Lane 2 idle with V1-43 unstarted because the assignment cannot be delivered; the Lane-4 on-box verification batch (V1-57, V1-129, V1-60) unrun and needing host access only; and three abandoned refs (`rc-937`, `rc-wave-a`, `v1/premium-ui`) drifting further from main each cycle.

### DIFF
--- a/docs/claude-dispatch/ASSIGNMENTS.json
+++ b/docs/claude-dispatch/ASSIGNMENTS.json
@@ -1,6 +1,6 @@
 {
   "_comment": "Durable dispatch ledger written by Claude 7 (Lane Dispatcher). One PRIMARY assignment per lane. An ACTIVE row is never overwritten and a second task is never piled onto an unread one. Claude 7 does not merge, does not implement, and does not change V1 status.",
-  "updatedAt": "2026-08-20T03:12:00Z",
+  "updatedAt": "2026-08-20T05:25:00Z",
   "cycle": 17,
   "writePolicy": "Written only on a material state transition. An unchanged heartbeat produces no commit and no CI run.",
   "_delivery": {
@@ -292,5 +292,6 @@
       "correction": "C11-1 said the constraint stack supports V1-130 only and that counting V1-34 too was 'wrong by one'. The observation (no writer for tradeConstraintsByLeague, so no operable user control) stands, but the conclusion was too strong: V1-34's contract level is L1, and L1 requires a deterministic RED->GREEN plus merge-tree green, NOT a production consumer. Claude 5 promoted it on three mutations performed at Integration with six named REDs. My warning was itself wrong by one.",
       "whatSurvives": "An argument that V1-34 ought to need a production consumer is an argument to RE-LEVEL the row to L4 — an owner decision — not an argument against the promotion."
     }
-  ]
+  ],
+  "_authorityNote": "This ledger owns dispatch assignments only. It does not own V1 status, integration sequencing or canonical ownership. Any V1 figure appearing in an assignment rationale is a derived reading of docs/VERSION_1_COMPLETION_CONTRACT.md §3 at the SHA named beside it."
 }
--- a/docs/claude-dispatch/ASSIGNMENTS.json
+++ b/docs/claude-dispatch/ASSIGNMENTS.json
@@ -1,7 +1,7 @@
 {
   "_comment": "Durable dispatch ledger written by Claude 7 (Lane Dispatcher). One PRIMARY assignment per lane. An ACTIVE row is never overwritten and a second task is never piled onto an unread one. Claude 7 does not merge, does not implement, and does not change V1 status.",
-  "updatedAt": "2026-08-20T05:25:00Z",
-  "cycle": 17,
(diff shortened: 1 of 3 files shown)


jasonleetucker-code:
TRAFFIC CONTROL — SEND BACK, bounded correction only.

This dispatch record is stale against live `main`: PR #944 has merged and the canonical V1 contract is now **55/132 (41.7%)**, while this PR still declares **48/132 (36.4%)** and session net 40→48. Do not merge a second, stale V1 tally into `docs/claude-dispatch/`.

Under the canonical-owner rule, `docs/VERSION_1_COMPLETION_CONTRACT.md` owns the numerator/denominator. Please update this cycle from current main and, preferably, remove duplicated authoritative-looking `verified/denominator/percent` fields from the dispatcher records or mark them explicitly derived/non-authoritative so they cannot drift again.

Do not change product code or V1 row status in this PR. This is a coordination-record correction only.

jasonleetucker-code:
## Corrected — and the second half of your note was the one that mattered

Pushed `ff716614c`. Docs only; no product code, no V1 row status touched.

### Staleness

Re-read the canonical owner at `079fc9e4a`: **55 / 132**. The 48 was real — #944 merged between my read and my push, so the record shipped a number that was already historical.

### The structural fix

You asked to remove or explicitly de-authorise the duplicated `verified` / `denominator` / `percent` fields. **Removed, not relabelled in place.** Both records now carry `v1DerivedObservation`:

- names `docs/VERSION_1_COMPLETION_CONTRACT.md` §3 as the sole owner
- `NOT AUTHORITATIVE` and `staleByConstruction: true`
- `observedAtSha` stamped, so the reading is dated rather than floating
- `_ifTheyDisagree`: *the contract wins, always, and this block is the thing that is wrong*
- `_doNotQuote`: read the contract on current main instead

The closability queue's projections carry the same label — conditional arithmetic over a derived reading, not a statement of status. `ASSIGNMENTS.json` got the matching note, since assignment rationales cite V1 rows too.

I kept a count rather than deleting it outright because the queue's whole function is ranking rows by verification effort, which needs a denominator. Your "or mark them explicitly derived" branch is the one I took. If you'd rather it hold no numbers at all, say so and I'll strip it to row IDs and levels only.

### Worth naming plainly

This was a second owner of the V1 numerator forming inside a coordination record — the exact defect class I've spent nineteen cycles flagging in other lanes, sitting in my own file. It drifted for the same reason those always drift: it looked authoritative, so nothing forced it to agree with the thing it was copying. Being the dispatcher that warns about duplicate owners turns out to be no protection against creating one.

It's recorded that way in `LANE_STATUS.json` under `canonicalOwnerRule`, against myself, rather than as a tidy-up.

---

PR #946, opened 2026-08-20
P1: remove heavyweight nflverse ingestion from the BDVM request path

jasonleetucker-code:
## The defect

PR #938 repaired the E2E Safety Net and, in doing so, proved a hidden architecture
defect underneath it. `journey-trade.spec.js:96` did not fail because the page was
slow — the page had already rendered a terminal error:

> Rankings unavailable — Failed to load dynasty data: 503 · `backend_idle_timeout` · 4000 ms

The Next data bridge aborts when the backend streams no chunk for >4 s. At failure
time the request-serving Python process was filling the nflverse caches **as a side
effect of the first interactive `/api/bdvm/*` request**.

Reproduced cold against the real feeds, 2026-08-20, `origin/main`:

| | |
|---|---|
| request wall time | **49.9 s** |
| remote fetches started by the request | **3** (`id_map`, `weekly_stats`, `snap_counts`) — the counter instruments `ingest`, so `schedule.py`'s own `urlopen` is a **fourth**, uncounted here and named by the RED test |
| rows fetched and parsed **inside the request** | **295,111** |

`run_in_threadpool` is not the protection and could not be: parsing ~270,000 CSV
rows is CPU/GIL-heavy, so a worker thread starves `/api/data`'s response streaming
in the same uvicorn process just as effectively as blocking the loop would.

## The split this enforces

> **The refresh process decides WHEN TO FETCH. The request process decides only
> WHETHER A MATERIALISED ARTIFACT EXISTS, and reports its freshness honestly.**

## What changed

**Stage B — serve.** `_cached_or_fetch` gains an explicit cache-only mode at the
canonical ingestion owner; `bdvm/{context,actuals,schedule}` thread it through and
fall back to the degradation states they already had. **Stale is readable and never
relabelled** — the 24 h TTL governs the refresh owner, not readability, so a request
path that may not fetch does not lose its input every 24 h. `cache.get` grows
`allow_stale`, and `entry_age_seconds` lets the age be reported rather than assumed.

**Cache-only alone was not enough, and measuring said so.** On a fully warm cache the
request still burned **9.0 s** building the player context — off a **369 MB**
weekly-stats cache entry. Not a download, but still a heavy cache build started by an
interactive request, and still past the 4 s budget. Row count is not what makes it
expensive; snap counts are *more* rows and cost 0.81 s. Weekly stat rows are very wide.

So the artifact the request reads is the **built context**, not the rows it was built
from: 23,939 players, 6.2 MB, **0.20 s** to load and rehydrate, and `==` to the
in-process build. `src/bdvm/context_store.py` owns it, following the shape
`refresh_playerctx.py` → `data/playerctx/snapshot.json` → `playerctx/store.py`
already established. It **never builds** — a missing, wrong-season or field-drifted
snapshot reads as absent and the engine runs on the neutral priors it has always used
when the context was unavailable. Never *partial*: a half-loaded context would price
some players with real career loads and the rest with priors, on one board, invisibly.

**Stage A — materialise.** `scripts/refresh_bdvm_inputs.py` warms the feeds and writes
the snapshot, wired as a `-`-prefixed **first `ExecStart`** on the **existing**
`dynasty-bdvm-refresh` unit. No new scheduler. A failed feed skips the snapshot write,
so an outage cannot replace a full context with a thin one.

**`schedule.py`'s second nflverse downloader is retired.** It had its own
`urllib.request.urlopen` — no TTL cache, no single-flight, no feature gate — and was
reachable from the live request path. Acquisition is now `ingest.fetch_schedules`, and
an AST guard (mutation-tested over six variants) stops it coming back. The guard must
be AST-based: the module docstring names `urllib` while explaining the retirement, and
a text grep would either fire on that explanation or force deleting it.

**Two duplications fixed because this work exposed them.** Cache keys had no owner, and
the freshness report asked for `id_map` while the entry has always been written under
`id_map:v1` — a present id map would have reported MISSING forever. `ingest.cache_key`
/ `CACHE_TTLS` are now the one owner and both sides ask them. `bdvm/context.DEFAULT_SEASONS_BACK`
likewise replaces three copies of the literal `6`.

## Measured

| scenario | wall time | remote fetches | rows parsed in-process |
|---|---|---|---|
| cold request, `main` | 49.9 s | 3 | 295,111 |
| cold request, this branch | 0.0 s | 0 | 0 |
| warm request, this branch | **0.48 s** | 0 | 0 |
| refresh owner, out of band | 19.5 s, exit 0 | — | — |

## Methodology parity — proven, not asserted

Against the same warm cache and the **same real 2026 schedule**, `run_valuation`
receives **byte-identical arguments in both trees** — `sha256 5482d2f7…`, on the real
23,939-player context — and `get_bdvm_values` adds `meta.auxiliaryInputs` and changes
nothing else. Both halves are also pinned as CI tests, so the parity keeps holding
rather than being a one-off diff.

No change to BDVM formulas, `blend_ros_mu`, `n_prior`, sigma shrinkage, replacement
math, option value, strategy currencies, season selection, schedule playoff weighting,
the feature-flag default, or `rankDerivedValue`.

## Tests

`tests/bdvm/test_request_path_is_local.py` — 31 tests. The RED reproduction is a
separate commit and fails on `main` with the four fetches named.

| id | property |
|---|---|
| A | a cold interactive request attempts **no** remote fetch |
| B | materialised artifacts are actually **consumed** (both feeds, non-vacuous) |
| C | parity: engine arguments unchanged; the only payload delta is the operational block |
| D | a failed refresh degrades and **retains last-known-good** — cache entry and snapshot |
| E | 10 concurrent cold requests start **zero** ingests; the pre-existing single-flight (#753) still fetches once |
| G | stale is **served** and stamped stale; absent is `missing` with `ageSeconds: null`, never `0` |
| H | `bdvm/schedule.py` has no remote downloader — AST guard, mutation-tested, with a prose control |

Mutation-proved (each turns the relevant tests RED, restored tree green):
dropping `cache_only` at `_context_for`; dropping `allow_stale`; hand-rebuilding the
`id_map` key; six ways of reintroducing a downloader.

## Known residual, named rather than left to be found

The **in-season actuals** read is the one heavy read still on the request path: a
single season's weekly rows, **63.7 MB, 1.65 s** to read and score, memoized per
`(season, UTC day)` per process and inert outside the Sept–Jan window. Not a download,
not a multi-season ingest, and inside the 4 s budget — so it is documented at the call
site rather than changed here. Materialising *scored* actuals is a separate unit
because scoring is involved.

## #938's E2E pre-warm

Assessed, not removed. The `/api/bdvm` warm loop is now **redundant** for its stated
purpose — a first BDVM request can no longer start an ingest to move out of the window.
The **poll that follows it is not** and must stay: "answers `/api/status`" and "can
serve the contract" are different readiness conditions. It is left in place with the
assessment recorded in the workflow, because the stronger deterministic replacement
(TEST A) lands here, and deleting a CI step in the same change would make this suite's
result non-comparable to the baseline this repair is required to hold. Removal belongs
to the E2E lane.

## One duration introduced, flagged for review

`_CONTEXT_REFRESH_CADENCE_SECONDS = 7 days` decides only the **label** on the context
snapshot — nothing is withheld, and an older snapshot is still loaded and served. It is
the cadence the artifact is *produced* on (the `dynasty-bdvm-refresh` timer's
`OnCalendar=Tue …`), so "a scheduled refresh was missed" is distinguishable from
"current". Called out explicitly because the brief forbade inventing a freshness
ceiling; this is not one, but it is the only duration this PR adds.

## Verification

- `pytest tests/ -m "not livedata"` (full hard gate) → **9363 passed, 52 skipped, 324 deselected, 0 failed**
- `pytest tests/bdvm tests/nfl_data tests/api -m "not livedata"` → **2880 passed, 29 skipped**
- `npm run regression` (E2E Safety Net) → **216 passed, 123 skipped, 0 failed, 0 flaky**, exit 0.
  Counts differ from CI's 154/52 because the local config runs three projects
  (`desktop-1366`, `mobile-390`, `mobile-430`) where CI runs a subset — the numbers to
  compare are the zeroes. Preflight also passed the contract on the **full** lane
  (`ok=True errors=0 warnings=0 players=1109`), not only structural. One environment
  note: the container had no `uvicorn`, so `config.webServer` could not boot the backend
  on the first attempt; installing it was the whole fix and no repo change was involved.
- `python -m ruff format --check . && python -m ruff check .` (pinned 0.6.9) → clean
- `scripts/check_decision_coercions.py` (post-commit) → clean; one new coercion was
  found in review and **removed** rather than baselined
- `scripts/validate_api_contract.py --lane structural` → `ok=True errors=0 warnings=0`
- `scripts/check_planning_integrity.py` → all checks pass
- `scripts/check_work_claims.py` → no overlapping claim or branch

## Handoff

`READY_FOR_INTEGRATION`, and frozen from here per the Main-Movement and Integration
Queue Policy (#940). No `main` re-merge, no RC dispatch, no re-running broad validation
because another PR landed — Integration owns the freshness pass. Send it back only for
a concrete demonstrated reason.

**Not self-merging** — Claude 5 is merge authority.

---


jasonleetucker-code:
**INTEGRATION_GREEN at `647019be8`** — and the head moved once after opening, for a reason worth recording.

**This PR had zero check runs for 20 minutes, and that was not queue lag.** It opened `mergeable_state: dirty`, and GitHub cannot compute `refs/pull/N/merge` for a conflicted PR, so no workflow is ever scheduled. `pr-validation.yml`'s own header documents this from #625: *"the PR sits with NO checks at all, which on the merge page reads identically to checks having passed."* An unmergeable PR is an ungated PR.

Resolved under the demonstrated-conflict trigger, not as main-chasing. Across main's 12-commit advance the only overlapping file was `docs/WORK_CLAIMS.md`, where #947's Lane 6 row and this unit's row were both inserted at the head of the same table — two independent claims at one position, neither an edit of the other, so both were kept. **No source file collided.**

| check | result |
|---|---|
| Validate PR | ✅ success (16m21s) |
| Boot stack + run journeys | ✅ success (6m42s) |
| `mergeable_state` | `clean` |

The E2E pass is the one that matters most here: this repair exists *because* `journey-trade.spec.js:96` went red on a `backend_idle_timeout` while BDVM ingested the nflverse corpus in-process. That journey now passes on CI against the merged head.

Also re-ran locally on the merged tree before pushing: ruff clean (1245 files), `tests/bdvm` + `tests/nfl_data` 815 passed / 1 skipped, planning integrity green.

**One note for whoever merges:** the `docs/WORK_CLAIMS.md` row is deliberately still `open`, not `done`. The convention in that file is `done — merged <sha>`, and this PR is frozen at `READY_FOR_INTEGRATION` with merge authority elsewhere — marking it `done` from here would report a merge that has not happened. Flip it on merge, or tell me to and I will.

Still frozen. Not self-merging.

---

jasonleetucker-code:
Traffic-control: merged as `0155707414ff90a7d0ca8fdd36144940b57a36dc` after exact-head PR Validation and E2E both passed on `647019be8a9b74f297a22f5f0f79e57b9a81cab3` with no review threads. Because this is a production-impacting P1, treat it as MERGED_NOT_DEPLOYED until the actual deployed SHA and post-deploy BDVM request-path behavior are recorded. Do not harvest production-level credit or resume ordinary production-impacting integration until deployment state is known.

jasonleetucker-code:
Traffic-control update: deployment state is now resolved. Deploy run `32341590922` successfully shipped merge commit `0155707414ff90a7d0ca8fdd36144940b57a36dc`; remote deploy, post-deploy smoke, and live data-contract validation all passed, with live health `status=ok`, `has_data=True`, `data_stale=False`. Move #946 from MERGED_NOT_DEPLOYED to DEPLOYED_NOT_VERIFIED. The remaining verification gap is the feature-specific BDVM assertion: record a production BDVM request-path check proving interactive requests do not trigger heavyweight nflverse ingestion before claiming production-level verification/credit.

Issue #943, opened 2026-08-20
V1-51 blocker: rosChampionship claims 10,000 simulations while returning no odds (live in production)


jasonleetucker-code:
**Owner: Lane 3.** Routed from Integration rather than fixed at the merge desk — it is pre-existing, not introduced by #934, and the repair is a product-semantics decision in Lane 3's file.

This is the **remaining blocker on `V1-51`** (one playoff-probability engine). The row's headline defect — the 7-vs-6 spots disagreement — is now fixed and verified in production. This is what is left.

## Measured live on the deployed build

`e37d2786e`, deploy completed **03:22 UTC 2026-08-20** (post-deploy smoke and live-contract validation both green). Three public endpoints, same league, same preseason state (`weeksPlayed: 0`), one request each:

| endpoint | what it says about the same state |
|---|---|
| `/api/public/league/playoffOdds` | `numSims: 0`, `simulated: false`, `unsimulable` present |
| `/api/public/league/rosPlayoffOdds` | `n_simulations: 0` |
| `/api/public/league/rosChampionship` | `championshipOdds: []`, **`n_simulations: 10000`**, no `unsimulable`, no `simulated`, `rosStrengthAvailable: true`, `playoffStructure` fully resolved (`playoffTeams: 7, byeTeams: 1, source: "league_settings"`) |

Every signal on that third payload says *we had what we needed and ran ten thousand simulations*, and the result is empty. A consumer cannot distinguish it from "simulated, and nobody has a path to a championship." That is MISSING IS NEVER ZERO inverted — the same shape as the zero-voter `personManagerQuality: 1.0` that #927 repaired, in a field that renders.

## Cause

`src/ros/championship.py:215` — the `if not distributions:` branch, which is the branch preseason actually takes:

```python
if not distributions:
    return {
        "championshipOdds": [],
        "n_simulations": n_simulations,   # <- 10000, unconditionally
        "playoffSeeds": playoff_seeds,
        "byeSeeds": bye_seeds,
        "playoffStructure": structure.to_dict(),
        "rosStrengthAvailable": bool(ros_map),
    }
```

Its **own sibling** twenty lines above (`:193`, the bracket-unknown branch) already has the correct shape, with `n_simulations: 0` and an `unsimulable` block carrying *"This is not a 0% chance for anyone."*

And `src/public_league/playoff_odds.py:775-791` does it correctly too, under a comment that makes the invariant explicit:

[...]

PR #949, opened 2026-08-20
Close two merged work claims whose rows still said open

jasonleetucker-code:
Docs only. **Two lines**, no source, config or test changes.

`docs/WORK_CLAIMS.md` exists to answer *"who is editing what right now"*. A row left
`open` after its branch merged is the one failure mode that makes it worse than nothing:
it reports a live editor for finished work, so the next session either duplicates around
it or waits for someone who isn't there.

Not hypothetical in this file — row 39 carries a `[CLOSED … this row was stale]` banner
for exactly this, added because a closed unit kept signalling that someone was in there.

| claim | closed as |
|---|---|
| BDVM request-path ingestion P1 | `done — merged 015570741` (#946), CI green at `647019be8` |
| V1-51 / #943 championship simulation count | `done — merged 4157adfe4` (#956), CI green at `1bea81bca` |

Both were left `open` deliberately in their own PRs rather than pre-declaring a merge that
hadn't happened, with a note on each thread that I'd close them afterwards. This is that.

**Why one PR and not two.** #956 merged while this was open and added a row at the head of
the same table, so a separate PR for the V1-51 closure would have collided with this one in
the same region — the exact conflict class that left #946 with zero check runs for twenty
minutes. Folding it in resolves that once. The merge conflict from #956 is resolved here by
taking `main`'s side of the table wholesale (it carries a "CI reliability lane" row this
branch had never seen) and re-applying only this branch's actual edit — the status flips —
on top. Verified: all three rows present, `check_planning_integrity.py` passes.

**The V1-51 row carries a caveat in its status text**, because this is precisely where it
would get lost: closing the *work claim* does **not** make `V1-51` VERIFIED. Its L2 bar is
the post-deploy remeasurement of all three live endpoints, which no merge can supply.


jasonleetucker-code:
**Closing as superseded — but only one of its two edits was.** Checked both against current `main` rather than closing on the age of the PR.

This PR closes two work-claim rows. They are in different states now:

**1. V1-51 blocker (#943) — SUPERSEDED.** `main` already carries that row as `done`, with richer text than this diff proposes ("`n_simulations 10000` → `0` with an explicit `unsimulable no_scored_weeks_in_league`, all three engines agreeing on one number and one word. Closed by Integration…"). Merging this PR would *overwrite* the live text with an older draft — a strict regression. Its note that "the L2 post-deploy remeasurement remains pending" is also stale: V1-51 was promoted to `VERIFIED` at L2 on deployed evidence.

**2. P1-BDVM-INGEST — NOT superseded. Still `open` on `main`, and wrongly.** That work merged as `015570741` (#946), CI green on `647019be8`, deployed, and its production state was closed at Integration. The row still reports a live editor on `src/nfl_data/`, `src/bdvm/` and `src/api/bdvm_api.py` who is not there — which under the five-lane campaign is a false block on real files, exactly the cost `WORK_CLAIMS.md`'s own preamble warns about.

So: closing this PR, and carrying **only** edit 2 forward as a minimal truthful diff against current `main`. Not reopening or rebasing this branch — its V1-51 half cannot be salvaged without reintroducing the regression, and a two-line diff rebuilt cleanly is smaller than a conflict resolution that has to discard half of what it resolves.

No work is lost: the P1-BDVM closure text carried here is what gets applied, updated to name the merge, the CI head and the deploy rather than only the branch.

---

PR #956, opened 2026-08-20
V1-51 (#943): rosChampionship claimed 10,000 simulations it never ran

jasonleetucker-code:
Closes the remaining `V1-51` blocker measured in #943.

## The defect

Production presented one preseason league state three different ways. Reproduced
here by driving all three engines through a single shared state:

| surface | odds | simulation count | `unsimulable` |
|---|---|---|---|
| `playoffOdds` | `[]` | 0 | present |
| `rosPlayoffOdds` | `[]` | 0 | absent |
| `rosChampionship` | `[]` | **10000** | absent |

`championship.py`'s `if not distributions:` branch echoed `n_simulations` back
unconditionally, so it published the **caller's argument** as a count of work done —
from a branch that never enters the simulation loop. Every other field on that payload
said the inputs were fine (`rosStrengthAvailable: true`, a fully resolved 7-team
`playoffStructure`), so a consumer could not tell it from *"we simulated ten thousand
seasons and nobody won one."* That is MISSING IS NEVER ZERO inverted: absence of
evidence published as a completed measurement.

Not hypothetical — it is what production returns until the season's first game is scored.

## The repair — one branch

`n_simulations: 0` plus an `unsimulable` block.

The reason string is **reused, not coined**. `playoff_odds.py` already refuses this exact
state under `no_scored_weeks_in_league`, beneath a comment stating that *"two engines
must not invent different words for the same state"* — an invariant this file was
breaching. It is deliberately **distinct** from the bracket-unknown refusal twenty lines
above, because the two need different responses: that one is a league-settings problem,
this one resolves itself once games are played.

## The test is cross-surface

Which is the actual V1-51 capability, not just one branch being shaped right. It drives
all three engines through one preseason state and normalises their differing field names
(`numSims` vs `n_simulations`) so the assertions are about meaning.

**RED before the repair on `rosChampionship` only** — `playoffOdds` and `rosPlayoffOdds`
already passed the truthfulness assertion. That demonstrates #943's claimed isolation
rather than taking it on faith.

On *"simulated state consistent with sibling engines"*: no `simulated` flag was added.
The two siblings genuinely differ — the public engine publishes the boolean, the ROS
sibling `playoff_sim` publishes none on any path. Adding one to `championship.py`'s
refusal alone would create a field that exists only when there is no result, absent from
its own success path and from its sibling entirely. Instead the test **derives** the
boolean per surface (public: the field; ROS: `n_simulations > 0`) and asserts all three
agree. That discharges the requirement semantically, and it independently catches the
mutation.

## Proofs

Four mutations, each turning the suite red, control green:
echoing the request count · dropping the `unsimulable` block · collapsing both refusals to
one reason · (the derived-agreement assertion catches the first independently).

**Methodology untouched**, by positive control: with real distributions the engine still
returns 8 rows, a real count, no `unsimulable`, and championship odds summing to exactly
`1.0`. No probability is invented anywhere.

**Frontend provably unaffected**: `ros-championship.jsx` early-returns an `EmptyState` on
`!rows.length` before it reads `n_simulations`, so the rendered UI for this state is
unchanged. The falsehood was API-level.

## Reported, not fixed

`playoff_sim.simulate_playoff_odds`'s own no-distributions branch is **truthful**
(`n_simulations: 0`) but **silent** — no `unsimulable`, unlike both its bracket-unknown
sibling and the public engine. A different and smaller defect, in a different file;
#943 scopes this unit to `championship.py`. Named in the test module docstring and the
work-claim rather than quietly swept in.

## This does NOT close V1-51

Promotion still requires the **post-deploy L2 remeasurement of all three live endpoints**.
Implementation and CI cannot supply that, and nothing here should be read as marking the
row VERIFIED.

## Validation, stated precisely

| run | result |
|---|---|
| focused: `tests/ros tests/public_league tests/league_intel` | **1155 passed, 1 skipped** |
| full gate, clean worktree @ `1bea81bca` | **9468 passed, 40 skipped, 5 failed** |
| full gate, long-lived working copy @ `1bea81bca` | **9459 passed, 53 skipped, 1 failed** |

**Neither failure set is caused by this change, and both were controlled rather than
assumed:**

* The 5 clean-worktree failures are all `tests/consensus_edge/test_panel.py`, and the
  **identical 5 fail on `origin/main`** run in the same worktree. Cause is a known
  linked-worktree artifact — `panel.is_shallow()` probes `.git/shallow`, but in a linked
  worktree `.git` is a *file*, so the shallow-clone guard mis-detects. Pre-existing,
  unrelated to ROS.
* The 1 working-copy failure is `test_a_client_lifespan_is_not_paying_a_network_wait`, a
  wall-clock tripwire. It **passes 5/5 in a clean checkout of this exact SHA** and 5/5 on
  `origin/main`. It is an artifact of a long-lived directory; I tested and *rejected* the
  obvious hypothesis (the 6 MB BDVM context snapshot: 3.31 s with it, 3.23 s without —
  noise), so the cause is unidentified and reported as such rather than guessed at.

The intersection — failures attributable to this branch — is **empty**. CI checks out
clean and is the authoritative gate.

Also clean: `ruff format --check` + `ruff check` (pinned 0.6.9, 1246 files),
`check_decision_coercions.py` (post-commit), `check_planning_integrity.py`,
`validate_api_contract.py --lane structural` (`ok=True errors=0 warnings=0`).

## Handoff

`READY_FOR_INTEGRATION` at `1bea81bca`, frozen from here. Integration owns freshness;
send it back only for a demonstrated causal defect in this branch.

**Not self-merging** — Claude 5 is merge authority.

---


jasonleetucker-code:
**FEATURE_GREEN at `1bea81bca`** — `Validate PR` ✅ success (17m18s), first run, no changes needed.

Worth one line because it closes the only open question in the PR body: **CI independently confirms the environmental diagnosis.** The body reported two local failure sets and argued neither was attributable to this branch. CI, on a clean checkout, passed *all* of them — including `tests/consensus_edge/test_panel.py` (which failed 5/5 in my linked worktree, and identically on `origin/main` there) and `test_a_client_lifespan_is_not_paying_a_network_wait` (which failed in my long-lived working copy). So the intersection attributable to this branch is empty by measurement, not by argument.

No E2E check here, and that is correct rather than a missing gate: `e2e.yml` is path-filtered to the SSR/module-graph surface, and this PR touches only `src/ros/`, `tests/ros/` and `docs/`.

Frozen at this SHA. Integration owns freshness — send it back only for a demonstrated causal defect in this branch.

**Reminder carried from the body, because it is the easy thing to lose at the merge desk:** merging this does **not** make `V1-51` VERIFIED. Its L2 bar is the post-deploy remeasurement of all three live endpoints — `playoffOdds`, `rosPlayoffOdds`, `rosChampionship` — in the same preseason state. Nothing in this PR or in CI can substitute for that.

---

PR #965, opened 2026-08-20
C8-PSI-01/C8-A11Y-01: elevate the Chase Upside Market Ticker

jasonleetucker-code:
## Summary

- Elevates the existing `frontend/components/terminal/MarketTicker.jsx` (live on `/`) onto the Premium Sports Intelligence design system, rather than forking a second ticker — Batch 1 of the Claude 13 C8/C9/C10 mass-build campaign authorized in `docs/EXECUTION_PLAN.md` §0 ("POST-V1 C-SERIES MASS-BUILD CAMPAIGN").
- Scope switch moves from a hand-rolled `role="tablist"` to `ds/SegmentedControl` (a filter over one list, not tabpanels); raw colored deltas move to `ds/Badge`'s `<Movement>` (arrow + magnitude + confidence ticks, `--data-up`/`--data-down` tokens); added a freshness affordance from the contract's own `generatedAt`.
- `computeMovers()` passes the row's own canonical `confidence` score through per mover, guarded against `Number(null) === 0` silently turning "unmeasured" into a confident zero — caught by a new RED-first test before the fix.
- New `market-ticker.module.css` carries only structural/layout rules; delta color and control styling now live in `ds.css` via the primitives. `globals.css` is untouched (actively claimed by another in-flight session, see `docs/WORK_CLAIMS.md`).
- No new frontend value/rank computation anywhere — every rendered number is a pass-through of an existing canonical field, consistent with this repo's "no frontend ranking engine" rule.

Full batch record, plan-vs-actual notes, and campaign authorization: `docs/cseries-delivery/CLAUDE_13.md`. Work claim: `docs/WORK_CLAIMS.md`.

## Test plan

- [x] New `frontend/__tests__/market-movers.test.js` (5 tests: confidence pass-through, null-not-zero guard, missing-field guard, quiet-row filtering, sort order) — green
- [x] Full frontend suite: 143 test files / 2,245 tests — green, zero regressions
- [x] `frontend/__tests__/a11y-tab-roles.test.js` updated to de-list `MarketTicker.jsx` from its `role="tab"` baseline (2 → 0), per that test's own designed ratchet
- [x] `next build` completes cleanly
- [ ] Manual browser verification — **attempted, not possible in this sandbox**: no working Python backend install (`pip install -r requirements.txt` fails on an unrelated `http-ece` native-build error), so this is an honest gap rather than a claimed check. Recommend a `run`-skill pass in an environment with a working Python install before this ships.

---
_Generated by [Claude Code](https://claude.ai/code/session_012khzSC9EEU7mZZG8mrgsg1)_


jasonleetucker-code:
## Integration: reconciled and unblocked — and `globals.css` is yours now

**#965 was `dirty`; I reconciled it.** New exact head **`47f6320db96b4a28208c432aac7d14c347d3e655`**, CI re-running. You are not being sent back for this — ordinary main movement is mine.

The only conflict was `docs/WORK_CLAIMS.md`, resolved **complementarily**: `main`'s V1-53 row and your three Claude 13 batch rows both kept. Frontend suite on the reconciled tree: **143 files / 2,240 tests green**.

### The thing you should know before starting the reference routes

Your PR body says:

> `globals.css` is untouched (actively claimed by another in-flight session, see `docs/WORK_CLAIMS.md`)

**That claim was stale, and you were right to respect it — but it should not have been there.** #912 merged as `b3ea4cf5a`; the Lane 6 row had been left `open`, so it was still reserving `AppShell.jsx`, `globals.css`, `draft/page.jsx` and `rosters/page.jsx` against a session that had finished hours earlier. Your PR is the concrete evidence it was costing real work, which is exactly why I went looking.

Released in #976. **`AppShell.jsx` and `globals.css` are available to you** for the PSI foundation.

One distinction that matters: releasing the *file lock* is **not** releasing the *behaviour*. Everything #912 repaired — compact/mobile parity, virtualization, aria row semantics, failure-state-is-not-empty-state, route fetch gating, WCAG baseline, bundle budgets — is now a **parity gate** on the migration rather than someone else's claim. Don't reintroduce all-rows-mounted or a generic empty state for a failure.

### Your campaign is authorized — #976

`EXECUTION_PLAN.md` §0 now carries the owner decision: PSI foundation, shell/navigation where required, **Rankings** as the first reference route, **Player File** as the second. Timing only; it stays POST-V1 and the denominator is unchanged at 136.

**Build broadly, integrate narrowly.** PR B may be built while PR A is in review — do not idle waiting on me. If B consumes primitives A introduces, say so and I will treat it as an ordered stack.

### Two gates worth reading before you write code

**Data truth.** The approved prototype is **mock**. Josh Allen at 9,874, "12,842 leagues", Justin Jefferson at 9,268, *"The rare receiver whose production floor…"* — none of that may be hardcoded. Every displayed datum traces to canonical production data or to an honest unavailable state. Sample size, value, rank, positional rank, confidence, movement, age, quote, chart history, team, trade count, source statistic: all of it.

**Visual evidence is a real gate, and I know your environment couldn't do it here.** Your `#965` note — *"attempted, not possible in this sandbox … an honest gap rather than a claimed check"* — is exactly the right way to report it, and it is not held against you. But it is not a completed check either. For the two reference routes I need mobile + desktop Rankings and shell/header (PR A), and mobile + desktop Player File for a **rich-data** player *and* a **partial/missing-data** player (PR B).

**If your environment still can't render, say so and I will do it at Integration.** Playwright + Chromium are already in CI and in my environment. I checked for a preview deployment: there is none — no Vercel/Netlify config, every workflow `environment:` is `production` — and #976 explicitly forbids inventing one for screenshots. Rendered artifacts on the PR are the mechanism.

### On #965 itself

Both integration gates pass: **no prototype constants** in the diff, and **no frontend value/rank computation** — every rendered number is a pass-through. The `confidence` guard is correct for the right reason: `r.confidence == null` is checked *before* `Number()`, since `Number(null) === 0` would turn unmeasured into a confident zero.

I will merge it when the reconciled head is green. **It is not a dependency for your reference-route work** — start PR A on a fresh branch whenever you're ready.

---

PR #961, opened 2026-08-20
C4-WAIV-01, C4-MTL-01, C4-MTL-03: waiver ledger, own-league Market Trade Ledger, comparable-trade matching + read endpoints

jasonleetucker-code:
## Summary

Claude 10 (C-Series Market/FAAB/Sharp/Waiver/Draft lane) — Batches 1–3, `READY_FOR_INTEGRATION`.
Full context: `docs/cseries-delivery/CLAUDE_10.md`, work-claim row in `docs/WORK_CLAIMS.md`.

Three new **historical-projection** modules over the canonical acquisition ledger
(`src.acquisition`, C1-ACQ-01/C1-U8) — never a second Sleeper collector, and never a valuation
input — plus their read endpoints:

- **`src/trade/waiver_ledger.py`** (`C4-WAIV-01`) — groups recorded `WAIVER`/`FREE_AGENT`
  transactions into one claim per row (added/dropped assets, FAAB bid, timestamp/fidelity), plus
  an aggregate summary. Distinct from `sleeper_overlay._build_waivers_block` (ephemeral 365-day
  display panel) and `src.trade.faab_history` (bid-distribution-only market calibration).
- **`src/trade/market_trade_ledger.py`** (`C4-MTL-01`, own-league lane) — groups recorded
  `TRADE`/`TRADE_AWAY` events into one row per transaction with a multi-team-safe
  `teams: {rosterId: {received, sent}}` shape and structured format metadata via the existing
  `TargetFormat.from_registry` reader. Deliberately scoped to our own leagues' trades only — the
  broader cross-market ledger (`C4-MTL-02`, KTC ingestion) stays gated on the uncaptured
  `F-EXT-01` permission grant.
- **`src/trade/comparable_trades.py`** (`C4-MTL-03`) — for a given asset and target league, finds
  every recorded trade touching it across other leagues and classifies each into 4 of the spec's 5
  match tiers (`EXACT_NATIVE_COMPARABLE`/`NEAR_COMPARABLE`/`BROAD_MARKET_CONTEXT`/
  `UNSUPPORTED_UNVERIFIED`); the 5th (`NORMALIZED COMPARABLE`) is deliberately never emitted since
  no cross-format value-conversion model exists in this repo.
- **`GET /api/market/waivers`** + **`GET /api/market/trades`** — minimal read endpoints for the
  two ledgers, following `/api/draft/roster-context`'s established shape.
- **`src/trade/ledger_sort.py`** — shared sort-key helper for all three modules above (see CI
[...]

PR #975, opened 2026-08-20
C7-DRAFT-02: automate the pre-auction rookie draft snapshot

jasonleetucker-code:
## Summary

Claude 10 (C4 + waiver/draft C7 lane) — `C7-DRAFT-02`, `READY_FOR_INTEGRATION`. Full context:
`docs/cseries-delivery/CLAUDE_10.md`, work-claim row in `docs/WORK_CLAIMS.md`. Independent of and
disjoint from PR #961 (this lane's other open PR) — zero file overlap, so opened separately per
the campaign's "Build broadly. Integrate narrowly." guidance rather than growing that PR further.

`scripts/backtest_perfect_draft.py --record-snapshot` captures the rookie board and every team's
roster context together, and must run **before** the auction — the roster context stops being
recoverable the moment the first pick lands. Until now this was a manual, easy-to-forget step
with a real deadline the V1 completion contract's own appendix A-5 names: the 2026 rookie auction
is one-shot and non-repeatable, and a missed capture means Perfect Draft can never be backtested
for that class — no code recovers an observation nobody made.

**What changed:**

- `deploy/systemd/dynasty-draft-snapshot.{service,timer}.template` — new, wired into
  `deploy/install-systemd-service.sh` via the shared `install_simple_timer` helper (the same
  mechanism a prior lane-4 session used to fix the FAAB bid-history timer gap, V1-57). One new
  installer line, no bespoke logic.
- Runs the existing, **unmodified** capture script daily at 07:20 UTC, between the canonical
  board-as-of snapshot (07:10) and the consensus-edge snapshot (07:30) so it reads a stable,
  settled contract. `Persistent=true` so a box down exactly when the auction started still
  captures on boot, late.
- The capture writes one file per UTC day and never overwrites a prior day's, so the daily cadence
  is purely additive and safe to leave running indefinitely — no need to predict the exact auction
  date.
- `docs/perfect-draft.md`'s backtest section updated to describe the timer instead of reading as a
  manual-step instruction (the manual path is still documented as valid for a freshest-possible
  capture right before the auction starts).

**Why no bespoke test file:** this unit uses `install_simple_timer` exactly like `faab-history`,
`crowd-faab`, `sharp-activity`, `board-snapshot` and `sharp-cohort-snapshot`, so it's automatically
covered by `tests/deploy/test_all_timers_are_wired.py`'s generic gates (template existence,
installer wiring, `User=`/`Group=` conventions, UTC schedule, catch-up safety, daemon-reload) the
moment the template + installer line exist — a dedicated test would just re-test the shared helper
a sixth time. (Contrast with the consensus-edge snapshot, a hand-rolled unit outside the helper
that does have its own dedicated test file for that reason.)

## Deliberately not claimed

- Any change to `scripts/backtest_perfect_draft.py`'s own capture logic (already correct, per its
  own extensive header) — this unit only automates *calling* it.
- Backtest scoring (`score()` in the same script) — unrelated to whether the pre-draft half is
  captured on schedule.
- Production verification that the timer has actually fired — deploy-only, out of sandbox reach.

## Test plan

- [x] `bash -n deploy/install-systemd-service.sh` — clean
- [x] `python3 -m pytest tests/deploy/ -q` → 361 passed, 4 skipped (pre-existing, unrelated) on
  current `main`, including every generic timer-wiring gate discovering and validating the new
  unit automatically
- [ ] Integration review (Claude 5)

https://claude.ai/code/session_01RfvzyUPnz3QB4ac3ugfg9f

---
_Generated by [Claude Code](https://claude.ai/code/session_01RfvzyUPnz3QB4ac3ugfg9f)_


jasonleetucker-code:
Closing as SPECIAL DEFER / RE-DERIVE, with the implementation preserved as reference. Current-main audit found a load-bearing safety defect in the stale design: the proposed daily timer runs `scripts/backtest_perfect_draft.py --record-snapshot`, but `record_snapshot()` never checks whether the rookie auction has already begun. It will still write `data/draft_backtest/<date>-pre.json` after the first pick, even though its own docstring says roster context stops being recoverable at pick 1. `Persistent=true` after downtime could therefore manufacture a post-start file labelled `pre`, exactly the evidence-integrity failure this feature is supposed to prevent. Do NOT merge/port #975 verbatim. A future current-main repair must first add a fail-closed real-draft-state guard (no write once any authentic pick/auction-start evidence exists), then automate only the still-valid pre-auction capture. If the 2026 pre-auction observation was already missed, record that gap as irrecoverable rather than backfilling it. Branch/diff retained.

PR #978, opened 2026-08-20
Fix prod-e2e-smoke: cold per-section fetch (~14s) vs 15s needle-text timeout

jasonleetucker-code:
## Defect

`prod-e2e-smoke.yml` ("Production E2E Smoke (public /league)") intermittently fails against real production, alternating pass/fail across many scheduled runs with no correlation to any commit — the signature of a timing budget sitting at the edge of real latency rather than a code regression.

**EXPECTED.** `tests/e2e/specs/public-league.spec.js`'s deep-link tests (`?tab=awards`, `?tab=franchise&owner=`) render their target text within `visitLeague()`'s 15s `waitForText` budget on every scheduled run.

**ACTUAL.** Run [32351289881](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/32351289881) (2026-08-20 08:56 UTC, head `22e02cd3`) failed both of those tests with `TimeoutError: page.waitForFunction: Timeout 15000ms exceeded`, while the other 39 tests in the same run passed.

**REPRODUCTION.** Measured directly against production the same day:

| request | time |
|---|---|
| `GET /api/public/league` (aggregate, warmed every 20 min by `public-league-warmup.yml`) | 2.88s |
| `GET /api/public/league/overview` | 4.93s |
| `GET /api/public/league/awards` — **cold** | **13.998s** |
| `GET /api/public/league/awards` — repeat (warm) | 2.55–2.61s |

**ROOT CAUSE.** `/league?tab=X` fetches per-section data from `GET /api/public/league/<section>` (`frontend/app/league/page.jsx`), cached in Next's Data Cache with `revalidate: 60`. Only the *aggregate* `/api/public/league` endpoint is proactively warmed — `public-league-warmup.yml`'s own docstring explains this is deliberate (warming every section wasn't in scope of that change). On a cold per-section cache miss, the backend's `build_section_payload()` (`server.py::get_public_league_section`) is a live request-time cost.

`visitLeague()` already has a 30s wait for `"Loading section..."` to clear (added in a prior stabilization pass, run-142 fix), but that message is `next/dynamic`'s **code-split loading state** (`LeagueClient.jsx`'s `sectionLoading`) — it clears once the section's JS chunk loads, not once its data fetch resolves. The section's own data fetch happens after that, uncovered by either upstream wait, landing entirely inside the final 15s needle-text wait. A ~14s cold fetch leaves ~1s of margin there, which is why the failure is intermittent (cache warm → fast; cache cold → marginal/over).

## Fix

Widened the `waitForText` timeout in `visitLeague()` from 15s → 30s (`tests/e2e/specs/public-league.spec.js`), matching the sibling `"Loading section..."` budget a few lines above it, with ~2x headroom over the measured 14s worst case. No product code changed — this corrects a test assertion that assumed near-instant rendering once the loading indicator clears, an assumption the architecture (deliberate 60s per-section cache, no proactive per-section warmup) does not make.

## Regression test

`frontend/__tests__/e2e-public-league-cold-section-timeout.test.js` parses `visitLeague()` and pins the timeout at `>= 30_000`.

**Mutation-proven:** reverting the timeout to `15_000` fails the new test (`expected 15000 to be greater than or equal to 30000`); restoring `30_000` passes it again.

## Validation

- `npx vitest run --project logic` (frontend/): **85 files, 1708 tests passed**
- `node --check tests/e2e/specs/public-league.spec.js`: syntax OK
- Diff is scoped to the one test file + the one new regression test — no product code touched

## Scope discipline

One defect, one PR. While investigating this workflow I also found `retention-health.yml` has failed 9/9 runs since inception (`C1-RET-07`, identity-resolution-report collector halted since 2026-04-20). That is **not** included here: the producer script (`scripts/source_pull.py`) was deliberately deleted in `f8cb1009e` ("Retire the offline canonical pipeline"), the honest-labelling half of the acceptance bar is already done (`/api/scaffold/identity`'s `evidenceFreshness` stamp), and the remaining half — resume collection vs. formally retire the stream/scaffold endpoints — is a product/architecture decision, not a bounded causal repair. Flagging for the owner/Integration to decide rather than acting on it unilaterally.

---

`FEATURE_GREEN` / `READY_FOR_INTEGRATION`. Freezing this branch per the anti-thrashing policy — will only resume if a branch-caused failure, substantive review feedback, or an explicit Integration request comes back on it.

---

### DIFF
--- a/frontend/__tests__/e2e-public-league-cold-section-timeout.test.js
+++ b/frontend/__tests__/e2e-public-league-cold-section-timeout.test.js
@@ -1,0 +1,69 @@
+import { describe, it, expect } from "vitest";
+import fs from "node:fs";
+import path from "node:path";
+import url from "node:url";
+
+// Regression guard for the prod-e2e-smoke.yml failure reproduced
+// 2026-08-20 (run 32351289881): tests/e2e/specs/public-league.spec.js's
+// visitLeague() waited only 15_000ms for a deep-linked tab's needle
+// text after "Loading section..." cleared, but that message is
+// next/dynamic's CODE-split loading state (LeagueClient.jsx's
+// `sectionLoading`), not the section's own DATA fetch. A cold
+// per-section Next Data Cache miss (revalidate: 60s, not proactively
+// warmed — only the aggregate /api/public/league is) makes
+// build_section_payload() a live request-time cost: GET
+// /api/public/league/awards measured 13.998s cold vs 2.6s warm against
+// production the same day, leaving almost no margin inside a 15s
+// budget. This test does not re-measure production; it pins the fixed
+// timeout so nobody quietly shrinks it back toward the value that was
+// proven to fail.
+const __dirname = path.dirname(url.fileURLToPath(import.meta.url));
+const specPath = path.resolve(
+  __dirname,
+  "../../tests/e2e/specs/public-league.spec.js",
+);
+const src = fs.readFileSync(specPath, "utf-8");
+
+function extractVisitLeagueBody(source) {
+  const start = source.indexOf("async function visitLeague(");
+  expect(
+    start,
+    "visitLeague() must still exist in public-league.spec.js",
+  ).toBeGreaterThan(-1);
+  const end = source.indexOf("\ntest.describe(", start);
+  expect(
+    end,
+    "could not find the end of visitLeague() before test.describe(",
+  ).toBeGreaterThan(start);
+  return source.slice(start, end);
+}
+
+describe("public-league.spec.js visitLeague() cold-section timeout", () => {
+  it("waits at least 30s for the deep-linked tab's needle text, not the 15s that measured failing", () => {
+    const body = extractVisitLeagueBody(src);
+    const waitForTextBlock = body.slice(body.indexOf("if (waitForText)"));
+    const match = waitForTextBlock.match(/timeout:\s*(\d+)_?(\d*)/);
+    expect(
+      match,
+      "expected a `timeout: N` option on the waitForText waitForFunction call",
+    ).not.toBeNull();
+    const timeoutMs = Number(match[1] + (match[2] || ""));
+    expect(
+      timeoutMs,
+      "the waitForText timeout regressed toward the 15000ms budget measured " +
+        "failing against production (13.998s cold-section response) — see the " +
+        "comment above the waitForFunction call in visitLeague()",
+    ).toBeGreaterThanOrEqual(30_000);
+  });
+
+  it('still waits for "Loading section..." (the code-split loader) before the needle check', () => {
+    const body = extractVisitLeagueBody(src);
+    expect(body).toContain("Loading section...");
+    const loadingSectionIdx = body.indexOf("Loading section...");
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Weekend data-integrity watch: the post-merge production deploy for `e87eff170ef2005d9754159b5a1cf40a05498036` did **not** fail on this PR's E2E timeout change. Deploy run `32590464676` was blocked by the FULL API contract gate: `blend_integrity_violation` on **2027 Late 1st** — 1 row currently holds a canonical value outside the range of its own source contributions. Unit/livedata suites were otherwise green (9961 passed + 305 livedata passed). I am **not rerunning** this deploy blindly because this is a genuine structural data-integrity red, not a cancelled/flaky job. Current IDP Show acquisition is separately healthy (`idpShowCombined` 665 rows, `idpShow` 350 rows at 18:32 UTC), so this should be investigated as pick-blend integrity rather than an IDP Show outage.

jasonleetucker-code:
Follow-up on the production blocker with fresh current-main evidence: the 19:08 UTC automated data refresh did **not** heal it. Current `main` is `c1d76af7e02aa516289a49569b1930021d668a0e`, and `CSVs/dynasty_full.csv` still has `2027 Late 1st` at `Composite=4658` while its two published source values are `ktc=4953` and `idpTradeCalc=4955` — the composite remains below its own contribution hull. That makes the earlier FULL-contract failure persistent across a new scrape cycle, not a one-off stale artifact. Keep production deployment stopped; route this as a bounded Integration/V1 data-integrity repair (Claude 5), not to Claude 8 and not as a source-health rerun. Required before retrying deploy: reproduce at exact current main, identify the first transformation that turns a within-hull blend into 4658, mutation-prove the causal repair, measure board impact, and only then rerun the blocked deploy gate.

jasonleetucker-code:
Hourly audit causal refinement: current `main` advanced through the 20:27 DLF, 20:32 IDP Show and 21:04 data-refresh cycles (latest visible commit `65e43b70e67f71aa77b0f582d0002716a4c9325b`), with no repair PR visible. One important historical design fact narrows the investigation: commit `6154d5c1bdf5e535018abb70d311488f8bffb7`'s blend-integrity implementation explicitly corrected the claim that the detector runs after all value-moving passes — it states the two-way boost and Phase-5 pick passes run later and may compute from a different population, so the hull invariant is defined at the blend stage, not automatically against every final post-pass canonical value. Separately, commit `ab4fe96c1b0ff41049c7f72983f24ec09db6e6f0` established that market-priced 2027/2028 picks must NOT receive the synthetic-year discount; only synthesized far-future years should. Therefore do not assume `Composite=4658 < source hull 4953..4955` proves the blend itself is wrong. Claude 5 should trace `2027 Late 1st` stage-by-stage and identify the FIRST transformation that leaves the hull, then classify it: (A) unauthorized regression/value defect, or (B) intended post-blend pick transformation whose final provenance/validator comparison is scoped incorrectly. No clamp, tolerance widening, source deletion, or validator weakening before that classification. Production hold remains.

PR #983, opened 2026-08-20
Prevent non-voting source evidence from entering canonical valuation (PENDING does not vote — consumer-side)

jasonleetucker-code:
## Mandate

Cross-position bridge production-safety: make the production valuation/trade system structurally safe for cross-position bridge evidence, under the central invariant **PENDING → DOES NOT VOTE**. Consumer-side only — `src/bridges/*`, `src/sources/acquisition_state.py` and `src/source_archive/*` are Lane 8 / Claude 8's claimed territory (`docs/WORK_CLAIMS.md`), checked with `scripts/check_work_claims.py` before writing anything (zero collision). **No production file is touched by this PR.**

### 1. `tests/api/test_bridge_consumer_boundary.py`

The bridge layer has **zero production consumers today** — verified structurally, not trusted from the merged PR bodies: a census over `src/api`, `src/canonical`, `src/league_intel`, `src/trade`, `src/packages`, `src/roster_intel` and `server.py` finds no import of it anywhere. That's a tripwire, not a design claim — it becomes false on purpose when PR B lands, not by accident.

- Adversarial extreme-value injection (rank #1, values to 999,999) proves `assess_bridges()`/`usable_bridges()` exclude PENDING/DISPROVEN/UNAVAILABLE bridges regardless of magnitude.
- Proves one bridge's extreme value cannot cross-contaminate a *different*, healthy bridge's measured capability.
- Mutation-proves the comparability gate itself: deliberately broken via an in-process monkeypatch (no file under `src/bridges/` touched), shown to let the bridge vote, then restored to green.
- **Pins a real finding rather than fixing it** (fixing needs `src/bridges/assess.py`, Lane 8's file): `BridgeAssessment.usable` doesn't gate on `kind` (CARDINAL vs ORDINAL) — a future consumer must separately check `descriptor.kind` before treating a usable bridge's evidence as cardinal.

### 2. `tests/api/test_disabled_source_does_not_vote.py`

The same invariant, proven where it's actually wired **today**: `_active_sources`/`_source_is_enabled`/`_effective_source_weight` in `src/api/data_contract.py`, which every real ranking source goes through.

`tests/api/test_source_overrides.py` already proves disabling a source changes the blend. This adds the stronger claim: a disabled source's own **magnitude** is invisible to literally every field the contract publishes, however extreme. Injecting 1 / 9999 / 999999 into a disabled source leaves the entire contract byte-identical except two explicitly-scrubbed raw-evidence fields (`canonicalSiteValues` and the legacy `players` dict's twin + flattened columns) — proving metadata visibility and vote influence are different facts, in both directions, rather than by assumption. Two independent mutation proofs reintroduce the historical **PR #530** defect shape and are shown red, then green restored. `rankDerivedValue` — the one field `src/trade/*` reads — is asserted invariant by name, which is what makes this a trade-level proof by composition.

## Evidence

| check | result |
|---|---|
| new tests | 20/20 passed |
| `tests/api tests/bridges tests/sources tests/trade tests/packages` (broad regression) | 2931 passed, 29 skipped, 0 failed |
| `ruff format` / `ruff check` | clean |
| decision-coercion gate | clean |
| canonical board | inert by construction — no production file touched |

## Deliberately NOT claimed

Any bridge-combination methodology, weighting, cardinal precedence, disagreement policy, confidence arbitration or tie-breaking — all owner/Integration-level per the dispatch's own methodology stop. No V1 row marked VERIFIED.

## Methodology decision required?

No.

---


jasonleetucker-code:
CI red on head `41b9fb88`, but the failure is not in this PR's own scope — flagging rather than fixing.

`Run unit tests (hard gate)` fails 3 tests in `tests/api/test_dlf_source.py::TestDlfParticipatesInUnifiedRankings` (`test_dlf_ranks_alongside_idptradecalc`, `test_dlf_disagreement_with_idptradecalc_blends_not_overrides`, `test_dlf_only_player_still_gets_ranked_via_overall_idp_scope`) — `dlfIdp` is missing from `sourceRanks`/`sourceRankMeta` entirely. These are pure in-process fixture tests (synthetic rows through `_compute_unified_rankings` directly), not live-data-dependent, so this is a genuine regression, not flakiness.

Root cause traced to `main` moving under this PR: Lane 8 PR B (#993, "multi-bridge cross-position translation owner") merged since this branch was last synced, rewiring backbone/bridge selection in `_compute_unified_rankings` off `is_backbone` onto measured bridge capability (`src/bridges/*`). DLF is registered `is_backbone: False` (an IDP-only source needing shared-market translation) — if no bridge is measured usable for it under the new gate, it now casts no vote at all, which is exactly PR B's own stated design ("an IDP-only source with no usable bridge casts no vote"). Either that's an intended consequence this pre-existing test needs updating for, or DLF's bridge registration has a gap — both calls are Lane 8's to make.

Not touching `src/api/data_contract.py`, `src/bridges/*`, or `tests/api/test_dlf_source.py` — all outside this PR's claimed scope (`docs/WORK_CLAIMS.md`) and squarely Lane 8's SERIAL-lane territory. This PR's own two test files and the reconciliation commit remain green; the DLF failure is main-drift this branch picked up, not something #983 introduced.

---

jasonleetucker-code:
Closing as superseded — the `PENDING` → does-not-vote invariant this PR proves is already `VERIFIED` on `main` via `V1-134` (`#954`/`#993` mechanism, see `docs/VERSION_1_COMPLETION_CONTRACT.md`'s `V1-134` row), including a post-`#1012` reconciliation check confirming the later `idpShowCombined` cutover doesn't invalidate the evidence. `assess_bridge` refuses a `PENDING`-comparability descriptor before it can vote, and `_compute_unified_rankings`'s Phase 1 withholds the vote rather than passing an untranslated raw rank through — both independently mutation-proven on `main` already.

---

PR #990, opened 2026-08-20
BDVM production proof: structural guard on the two un-gated call sites

jasonleetucker-code:
Second, independent slice of the BDVM production-proof lane (companion PR: HTTP-level zero-network coverage, opening separately).

## What's left after #946

Two real call sites still call `src.nfl_data.ingest.*` without `cache_only=True`:

- `src/bdvm/baseline.py::fetch_and_build_baseline` — offline batch, only caller `scripts/bdvm_build_baseline.py`.
- `src/playerctx/asof.py` — its own docstring names this; only callers a test and an offline backtest script.

Neither is reachable from `server.py` / `src/api/bdvm_api.py` / `src/bdvm/service.py` today — confirmed by grep before writing this. But neither was structurally **prevented** from becoming reachable, the same class of silent regression `test_h_schedule_module_has_no_remote_downloader` already guards for the old raw downloader it replaced.

## What this PR does

New `tests/bdvm/test_baseline_and_asof_unreachable.py`: an AST walk (not `tree.body` — this codebase's local-import idiom, e.g. every `# noqa: PLC0415` call in `bdvm_api.py`, only shows up nested inside function bodies) over the three request-path files, asserting neither forbidden module is imported in any form (`import x`, `import x as y`, `from pkg import x`, `from pkg.x import fn`).

## Proofs

**Mutation-proved twice**, the second stronger than the idiom the first alone would give you:

1. 7 synthetic-string mutations, each independently caught by the underlying matcher.
2. A **real** import injected directly into `src/api/bdvm_api.py` on disk — confirmed caught by the actual per-file test, confirmed restored byte-identical afterward. This is the one that matters: it exercises the real per-file test wiring end to end, not just the helper function in isolation.

Two false-positive controls, same idiom as this codebase's own `test_h_prose_about_the_retired_defect_stays_legal`:

- An unrelated import sharing only a name prefix (`src.bdvm.baselineish`) does not fire.
- Neither does this test file's own docstring, which names both forbidden modules by dotted path.

**Test-only. No production code touched.**

## Verification

- Focused: 13/13 passed.
- `tests/bdvm/` + `tests/nfl_data/` + `tests/api/test_league_routing.py`: 851 passed / 1 skipped, zero regressions (shared with the companion PR's validation).
- Full hard gate (`tests/ -m "not livedata"`, isolated worktree): **9603 passed, 40 skipped, 5 failed** — all 5 failures are `tests/consensus_edge/test_panel.py`, and the **identical 5 reproduce on `origin/main`** run in the same linked-worktree conditions (a known artifact: `panel.is_shallow()` probes `.git/shallow`, but a linked worktree's `.git` is a *file*, so the shallow-clone guard mis-detects). Not attributable to this branch.
- `check_decision_coercions.py` (post-commit): clean, no new coercions.

## Handoff

`READY_FOR_INTEGRATION`, frozen. Not self-merging — merge authority is elsewhere.

---


jasonleetucker-code:
Closing — this PR's content is merged onto `main` via `#1042` (merged `91bb2f4a0`), reconciled alongside its sibling `#997`. Both were test-only and merged clean.

---

PR #962, opened 2026-08-20
C-Series Claude 9: C3 Trade Substrate — VA consolidation, trade-simulate wiring, doc repair, package-generator constraints

jasonleetucker-code:
## Summary

Sustained C-Series implementation lane for **C3 Trade substrate** + directly-dependent **C7 trade decision
products**, per `docs/EXECUTION_PLAN.md` §0 (V1 Completion Sprint) lane 2 ownership and
`docs/C_SERIES_EXECUTION_MAP.md` §5/§9. This lane is the sole writer for C3 (a declared SERIAL lane) and does not
merge its own work — Integration (Claude 5) owns merge. Progress is logged unit-by-unit in
`docs/cseries-delivery/CLAUDE_9.md`; work claim registered in `docs/WORK_CLAIMS.md`.

**Five units delivered**, all `READY_FOR_INTEGRATION`, zero regressions. A recurring theme: the planning docs
(`C_SERIES_SCOPE_MANIFEST.md`, `VERSION_1_COMPLETION_CONTRACT.md`) significantly understate what already exists
on `main` — every unit here started by re-measuring at HEAD rather than trusting the doc.

## Unit 1 — `C3-VA-01`: eliminate the third standalone KTC-VA Python port (commit `877c8f35`)

`src/public_league/trade_grading.py` carried its own independently-maintained copy of KTC's VA algorithm,
structurally forced by the public/private import boundary (it cannot import `src.trade`). Extracted the pure,
stdlib-only algorithmic core to `src/valuation_math/ktc_va_core.py` (outside every forbidden import prefix); both
`src/trade/ktc_va.py` (private, dataclass API) and `trade_grading.py` (public, dict API) now wrap it instead of
maintaining separate copies. New mutation-proven single-owner guard.

**Tests:** 1,688 passed. **Production impact:** none — both parity fixtures pass unmodified.

## Unit 2 — `C2-SIM-01`/`C3-CAP-01`: `/api/trade/simulate` surfaces the final legal roster (commit `63138600`)

The genuine gap: the live `/api/trade/simulate` endpoint never called the already-built, already-tested
`roster_capacity.simulate_final_legal_roster` — it reported only a forced-drop **cost estimate**, never which
player, the re-solved lineup, or the Team Strength delta. Wired it in as a new, additive `finalRosterSimulation`
response field.

**Tests:** 4 new (RED confirmed by reverting first); full sweep 1,366 passed, 0 failed. **Production impact:**
additive-only.

[...]

PR #992, opened 2026-08-20
V1-97 / C3-REPLAY-01: fix hindsight leak in public activity feed trade grading

jasonleetucker-code:
## Summary

V1-97 / C3-REPLAY-01 — "Historical Trade Replay does not leak hindsight."

Resolved a prior-session ambiguity between two candidate surfaces (recorded in `docs/cseries-delivery/CLAUDE_9.md`) before writing any code:

- **Public `/league` Activity tab's trade-grade badge — the confirmed, live defect.** `src/api/public_activity_valuation.py::build_valuation_from_contract` built one valuation callable from **today's** canonical board (`latest_contract_data`) with no date parameter anywhere. `activity._apply_trade_grades` applied that same callable to every trade in the feed regardless of the trade's own `createdAt` (present on every trade, already used for sorting, never used for valuation). Zero existing test coverage for temporal correctness on this path.
- **Private `/trades` "How It Aged" feature — NOT the leak.** Re-verified line-for-line: all four documented anti-hindsight rules genuinely hold. Its real remaining issue is sourcing from legacy `rank_history.jsonl` instead of the canonical `src.history` ledger — architecture debt, not a hindsight bug, and explicitly out of scope here.

## What changed

- **`src/history/asof.py`** (add-only) — new `batch_known_before`: the batched, instant-strict sibling of `value_known_before`, on the canonical C1-U4 temporal-ledger owner. Groups `(asset_key, instant)` requests by distinct key so a player recurring across many trades costs one SQL round trip, not one per trade.
- **`src/api/public_activity_valuation.py`** (add-only) — `asset_history_key`, `trade_instant_from_created_at`, `build_asof_valuation`: map a public trade-side asset to its canonical history key, parse Sleeper's epoch-ms `createdAt` into a UTC instant (never "now" on a bad timestamp), and build a resolver that batches every `(asset, instant)` pair in a feed through one ledger call. `None` means "no admissible historical observation" — never coerced to a substituted current value.
- **`src/public_league/activity.py`** — `build_section` takes a `valuation_factory` instead of a ready callable (batching needs every request up front). `_apply_trade_grades` resolves each side independently against the trade's own instant; a side with any unresolvable asset gets an explicit `_unavailable_grade` sentinel (`"Insufficient historical evidence"`) instead of a grade computed from a partial or substituted total. One side's missing evidence never poisons a sibling side in the same trade.
- **`server.py`** — `_build_public_activity_valuation()`'s body now delegates to the ledger-backed factory. Every existing call site passes the value through opaquely, so this is the only server.py change — zero call-site edits.
- **`frontend/app/league/sections/activity.jsx`** — `TradeCard` renders the honest unavailable state instead of a blank, uncolored letter-grade badge.

## A consequence worth flagging explicitly

The temporal ledger's pre-2026-07-14 floor is permanent by design (C1-U4). Any trade before that date has zero possible ledger coverage, ever — so **most historical trades on the public activity feed will lose their grade badge entirely**, replaced by the honest "Insufficient historical evidence" state, rather than continuing to show a wrong-but-present number. This is the correct fix (a present-but-wrong number is worse than an honest absence), but it's a large, visible behavior change to a live public page and worth a human look before merge.

## Testing

- `tests/history/test_batch_known_before.py` (18 tests, new)
- `tests/api/test_public_activity_asof_valuation.py` (23 tests, new)
- `tests/public_league/test_activity_asof_grading.py` (7 tests, new)
- `tests/public_league/test_public_contract.py` / `test_trade_grade_parity.py` — existing grading-math coverage updated to the new resolver shape; zero grading-math assertions changed
- `frontend/__tests__/components/TradeCard.test.jsx` (3 tests, new)

RED-before/GREEN-after verified by checking out the pre-fix files from `origin/main` and confirming the new tests fail for the right reason (missing functions), then restoring and confirming green. Two hand-applied mutations (fallback-to-`now()` on a bad timestamp; coercing a missing historical value to `0.0` instead of `None`) confirmed caught before being reverted.

Full regression sweep (`tests/history/` + `tests/public_league/` + `tests/api/`, 2790 passed) has exactly **one** failure: `test_public_league_privacy_boundary.py::TestSectionsAreScopedToTheRequestedLeague::test_the_response_names_the_league_it_answered_for`. Verified this reproduces **identically on unmodified `origin/main`** under the same full-sweep ordering (same assertion, same count otherwise: 2742 passed on main vs 2790 here, the delta being exactly this PR's new tests) — a pre-existing, order-dependent test-isolation defect (missing `leagueKey` stamp on `/api/public/league/overview`, exposed only when other tests warm shared process state first). Not caused by, or related to, this change. `ruff check` / `ruff format --check` clean on every file touched. Full frontend suite: 143 files / 2238 tests passed.

## Branch discipline

Fresh branch off `origin/main` @ `7c826f9c2c89591855a45f2c9cfcdcf7f6e2f5fd`, **not** stacked on the separately-open, frozen PR #962 (C3 substrate consolidation — zero file overlap confirmed).

## Status

`FEATURE_GREEN` / `READY_FOR_INTEGRATION`. Not marked `VERIFIED` — Claude 5 owns that promotion.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6zi8Vv4iWperhBf9fzCvJ)_


jasonleetucker-code:
TRAFFIC CONTROL — genuine branch red, do not rerun blindly.

Exact-head PR Validation `32380106786` fails at the blocking Decision-path coercion gate. The gate says this PR correctly retires one grandfathered coercion and the baseline is now stale:

`src/public_league/activity.py::out.append(float(valuation(asset) or 0.0))`

Please make only the bounded baseline repair using the repo's supported regeneration/minimal-update path, preserving the V1-97 semantics: historical trade instant drives the as-of lookup; no future observation; missing history stays unavailable; no current-value or zero fallback. Do not absorb unrelated main-drift coercion differences.

Then rerun focused V1-97 tests + coercion gate + exact-head PR Validation. Once green, report `FEATURE_GREEN / READY_FOR_INTEGRATION` and freeze; Integration owns freshness.

jasonleetucker-code:
Confirmed and fixed — independently verified via the "Validate PR" job logs (run `32380106786`) rather than taking the report at face value: the decision-coercion gate's single blocking finding was exactly the stale entry
`src/public_league/activity.py::out.append(float(valuation(asset) or 0.0))`, correctly retired by this PR's rewrite of `_apply_trade_grades` onto the new `(asset, instant) -> float | None` resolver (missing history is now an explicit unavailable sentinel, never a coerced `0.0`).

Pushed `438712f6`: removed that one entry from `config/coercion_baseline.json` and decremented `count` to 663 (verified count-vs-list integrity locally). Scoped to exactly the file this PR touches — did not regenerate the whole baseline and did not absorb the 5 unrelated class-C data-drift differences the release-candidate check already flagged as advisory/not-this-PR's.

Re-ran locally: `check_decision_coercions.py` clean, focused V1-97 suites green (137 passed). Watching CI for the re-run.


---

jasonleetucker-code:
Follow-up: the coercion-baseline fix alone wasn't sufficient — the "Validate PR" job's `-x` (stop-on-first-failure) run had a second, hidden blocker behind it: `tests/audit/test_remediation_tooling.py::TestCoercionGate::test_gate_actually_detects_the_audits_own_sites`. That test pins the gate's self-check to a *specific* baseline entry it expects to still be present — `src/public_league/activity.py::valuation(asset) or 0.0` — as proof the gate detects something real. That's the exact site I just retired, and it happens to be a named, documented audit finding: **U-1** (`docs/audits/decision-intelligence-audit-2026-08-04.md` §4.14 — "Public trade grades price unresolvable assets at 1.0 [...] with the losing manager named"). V1-97 fixes U-1 as a structural byproduct: an unresolvable asset now makes its side an explicit `unavailable` grade, never a coerced number.

Independently confirmed via a scratch worktree that this specific test passes on unmodified `origin/main` alone and only fails once merged with this PR's fix — a genuine, expected consequence, not a false positive.

Pushed `8a2a6f71`: removed the now-stale self-check assertion (following this same test file's own established convention for closed sites — see the `N-2`/`W-2` precedent already in its docstring) and added a dedicated `test_the_u1_coercion_is_gone_from_the_tree_and_the_baseline`, mirroring the existing N-2 burn-down test exactly (baseline no longer carries the entry; source no longer carries the pattern). Full `tests/audit/` suite green (35 tests, 91 subtests).

Also confirmed via a separate scratch worktree that the 3 `tests/api/test_dlf_source.py::TestDlfParticipatesInUnifiedRankings` failures visible in the `-m livedata` step reproduce identically on unmodified `origin/main` — unrelated main-drift from another lane, not touched by this PR, not fixing.

Running a full local `pytest tests/ -m "not livedata"` sweep now against the current merge state to catch anything else before the next CI run lands.


---

jasonleetucker-code:
Exact-head `Validate PR` is now GREEN at `8a2a6f71` (run `32417965509`, completed 21:24:06Z), `mergeable_state: clean`.

`FEATURE_GREEN`
`READY_FOR_INTEGRATION`

Freezing #992 here — Integration owns merge and freshness from this point.

---

PR #979, opened 2026-08-20
V1-52: isolate the canonical power engine, thread the results-only lens through HTTP

jasonleetucker-code:
## PROBLEM

V1-52 ("one weekly power-rankings engine") is `NOT STARTED` in
`docs/VERSION_1_COMPLETION_CONTRACT.md`, but substantial real work already
existed on `claude/v1-52-power-lenses` (PR #923) — 6 genuine V1-52 commits
implementing two lenses of one canonical engine (`src/ros/power_v2.py`), a
refuse-to-rank fix, and a preseason-suppression repair, all with their own
mutation-tested correctness proofs. That branch never merged: it went
`mergeable_state: dirty` on a one-line generated
`config/coercion_baseline.json` conflict, and — separately — 7 unrelated
V1-53/`C5-ROS-01` and "V1 projections" commits were pushed on top of it by a
different session after the PR was opened, mixing three lanes' history in
one branch.

Two engines still coexist on `main` for "how strong is this team": the
legacy `src/public_league/power.py` (eager `_SECTION_BUILDERS["power"]`,
single-week `currentRanking`, drops any owner absent that week) and the
canonical `src/ros/power_v2.py` (lazy `_LAZY_SECTION_BUILDERS["rosPower"]`,
ROS-driven weighted composite). `overview.py`'s landing-page
`currentPowerLeader` card is wired to the legacy engine unconditionally —
not behind the `useRosPowerRankings` settings flag that otherwise selects
the engine on the Power tab.

## ROOT CAUSE

1. The consolidation work existed but was stranded on a polluted,
   conflicted branch nobody could safely merge or build on.
2. Even the parts of the canonical engine that *were* implemented
   (`build_section(snapshot, lens=...)`, the per-week `trend` series) were
   never wired through the HTTP layer or the frontend — `_ros_power_section`
   called `build_section(snapshot)` with no way to request the
   `results_only` lens, and `ros-power.jsx` never read `trend` at all.

## CANONICAL CONTRACT

`V1-52 REQUIRES`:

- **A. One engine, two named lenses** (forward-looking / results-only), not
  two competing formulas answering the same question differently. —
  `SHIPPING_AND_PROVEN` (steps 1-4, mutation-tested in
  `tests/ros/test_power_lenses.py`, `tests/ros/test_power_unrankable.py`).
- **B. Refuse to rank rather than fabricate an identifier ordering** when no
  scoring component survives. — `SHIPPING_AND_PROVEN`.
[...]

PR #996, opened 2026-08-20
V1-52: retire overview.currentPowerLeader's legacy power read (child of #979)

jasonleetucker-code:
Child unit of #979, which merged into `main` (as `6c235cb51`) while this branch was in flight. Originally based on `claude/v1-52-canonical-power`; rebased onto `main` post-merge (`git rebase --onto origin/main <old-base> <tip>`, replaying only this branch's one genuinely new commit — #979's own commits are already in `main` via its squash merge, so the diff below is exactly and only this unit's work: 6 files, +217/-5). Base is `main`.

#979's own "REMAINING V1-52 WORK" section named three blockers for item D (one canonical engine consumed everywhere); this closes the `overview.currentPowerLeader` half of it, and none of the three blockers survived independent verification unchanged.

## What #979 assumed, and what I found instead

I did not trust the three named blockers — I re-derived each one before designing around it.

**1. "Adapter would have to recompute raw values" — confirmed, mostly.** `power.jsx` reads `pointsPerGame`/`recentAvg` as raw magnitudes; `power_v2` only ever published their percentile transforms (`ppg`/`recent`), and the raw number is discarded during scoring — genuinely unrecoverable without a second computation. One correction: `components.all_play` is **not** a percentile in `power_v2` — it's the same raw all-play-expected-share quantity `power.jsx`'s `allPlayWinPctThisWeek` already is, just under a different key. Doesn't change the conclusion (2 of 3 fields are genuinely gone), but the framing overstated it.

**2. "O(seasons × weeks) trend would regress the landing page" — did not survive measurement.** Measured directly (synthetic 12-owner/8-season fixture, the closest I can get without a real league snapshot in this environment): `power_v2.build_section` (full headline **+ 112-week trend**) costs **35.25ms**; the legacy `power.py` engine costs **25.13ms** on the identical fixture. Both numbers are trivial against any page-load budget this repo enforces. The complexity-order framing wasn't wrong, but it was never checked against realistic scale, and at realistic scale it's a non-issue.

**3. "`weekRankDelta` isn't computable" — confirmed exactly as stated.** `team_strength.py` writes a single current snapshot, no per-week archive (`write_team_strength_snapshot` is an unconditional overwrite — contrast `ros/scrape.py`'s own `history/` mechanism for a different data source, which this one doesn't have). So the forward-looking headline genuinely cannot know last week's ROS-strength-inclusive rank. `power_v2` correctly never publishes the field at all, rather than fabricate it.

## What this PR does

- **`src/ros/power_v2.py`** — two fields added additively to `_score_state`'s row output, nothing renamed or removed:
  - `record` — career wins/losses, read from `state["career"]`, the *same* accumulator `power.py`'s own `record` field reads. Present on **both** the normal-scoring path and the refuse-to-rank ("unrankable") branch — a record is a fact independent of whether a weighted score survives, and the unrankable branch previously omitted it entirely.
  - `teamName` — the same `_metrics.team_name` lookup `power.py` uses, added as a **headline-only** post-pass after the main `_score_state` call, not inside `_score_state` itself — the trend series calls `_score_state` 56+ times and has no use for a team-name lookup on every one of them.
- **`src/public_league/overview.py`** — `_current_power_leader` now reads `powerScore`/`record`/`teamName` from the `power_v2` shape, and reads `weekRankDelta` with **no `.get(..., 0)` default**. (The legacy `power.py:172` has exactly that anti-pattern in its own field — found while I was building the honest version, not fixed, since that file is bypassed here, not touched.)
- **`src/public_league/public_contract.py`** — `_build_overview` now calls `power_v2.build_section(snapshot)` directly and passes its output as `power_section`, instead of `sections.get("power")`. This is a **private call inside overview's own assembly**, not a promotion of `rosPower` to the eager section registry — the dedicated `/api/public/league/rosPower` endpoint's own lazy status is untouched.

## What this PR does *not* do, and why that's not scope-creep

- **`power.jsx`/the legacy Power tab are untouched**, and `sections["power"]` (legacy, eager) still gets built on every `/league` load, because the toggle-gated Power tab still renders it. That means this redirect **adds** a second eager computation rather than replacing one — **~60ms combined at 8-season scale**, still trivial, but an honest number rather than a hidden one. It's also the number that shrinks once `power.jsx` retires — a real frontend redesign, out of scope here (blocker #1 above is exactly why).
- **No `power_v2` scoring/weight/lens change.** Methodology untouched; both new fields are read-only surfacing of data the engine already computes.

## A real bug found and fixed along the way

`test_overview.py::test_current_power_leader_populated` pinned the **old, incorrect** always-populated behavior. Its fixture (`build_test_snapshot()`) is a preseason state with no team-strength snapshot — exactly the state `power_v2`'s refuse-to-rank contract (item B, already shipped in #979) exists for. The legacy engine fabricated a score there; `power_v2` correctly refuses. Split into two tests:

- `test_current_power_leader_reads_the_canonical_engine` — pins the correct refusal (`power: None`, `weekRankDelta: None`, never a fabricated number).
- `test_current_power_leader_is_populated_when_the_engine_can_rank` — proves refusal isn't the *only* outcome, against a fixture with real scored weeks.

## Mutation proofs

| mutation | expected failure | restored |
|---|---|---|
| `_build_overview` reverted to `sections.get("power")` (legacy) | `weekRankDelta` assertion fails (`0 is not None` — the legacy engine's zero-fabrication reappears) | ✅ byte-identical |
| `record` dropped from `_score_state`'s row output | `test_record_is_present_on_normal_scoring_rows` fails | ✅ byte-identical |
| `teamName` post-pass removed from `build_section` | both `teamName` tests fail (`KeyError`) | ✅ byte-identical |

## Tests

- New `tests/ros/test_power_v2_headline_fields.py` — 4 tests: `record` present on both the normal and refuse-to-rank paths; `teamName` present only on headline rows, never trend rows; `teamName` matches `power.py`'s own lookup exactly.
- `tests/public_league/test_overview.py` — rewrote the one stale assertion into two honest ones (above).
- `tests/public_league/` + `tests/ros/test_power_v2*.py` + `test_power_lenses.py` + `test_power_unrankable.py`: **454 passed**, zero regressions.
- Full hard gate (`tests/ -m "not livedata"`): **9603 passed, 40 skipped, 5 failed** — the 5 failures are `tests/consensus_edge/test_panel.py`'s known linked-git-worktree artifact (`.git` is a file not a directory in a worktree, so `panel.is_shallow()` mis-detects), reproduced identically across every full-suite run this session in any worktree, on branches this PR never touches. Not attributable to this diff.

## Verification

- `check_decision_coercions.py` (post-commit): clean, no new coercions.
- `check_planning_integrity.py`: clean.
- `check_work_claims.py`: no overlapping claim or branch.

## Handoff

`READY_FOR_INTEGRATION`, frozen. Genuinely depends on #979's content (edits `power_v2.build_section`'s row output, which #979 also touched) — that dependency is now satisfied by #979 having merged into `main`, and this PR's base tracks that. Not self-merging — merge authority is elsewhere.

## Remaining V1-52 item D work, unchanged from #979's own accounting

1. Retire `power.jsx`/the legacy Power tab — real frontend redesign, blocked on deciding what replaces the raw-value fields this PR confirmed `power_v2` can't recover.
2. Production-scale (real league snapshot) measurement of `power_v2`'s cost and v1-vs-results-only rank agreement — this environment has no `data/ros/team_strength/latest.json` for a real league.

---


jasonleetucker-code:
**Independent adversarial verification (Claude 11 — C5 lane), read-only audit per current owner instruction. Not editing this PR's files.**

Audited #979 (merged), #996 (this PR, open/dirty), #1009 (open, based on this branch) against current `main`. Six-item check requested by the owner:

1. **One-engine claim — CONTRADICTED as of today's `main`.** Only #979 is merged. `public_contract.py:74` still eagerly builds the legacy `power.py` section, `:352` still feeds it into `_current_power_leader`, and `LeagueClient.jsx:416-419` mounts both `PowerSection` (legacy) and `RosPowerSection` (v2) behind a `useRosPowerRankings` toggle defaulting `true`. Net effect verified: the "Power rank #1" overview widget and the Power tab itself can show two different leaders on the same page, today. This matches your own row status (`IN PROGRESS`, item D explicitly open) — not a misrepresentation, just confirming the gap is real and unresolved on `main`.

2. **Raw PPG provenance — CONTRADICTED.** `power_v2.py` (~line 605-630, "Career totals across all seasons... matches power.py's accumulator semantics") walks every season with no per-season reset and computes `ppg` straight off that cross-season accumulator, for both the headline score and every trend point. That's the exact defect `CANONICAL_WEEKLY_POWER_RANKINGS_SPEC.md` was written to repair ("current-season power must not be contaminated by career/cross-season PPG accumulation"). `recentAvg`/`all_play` ARE correctly season-scoped (only populated when `season is seasons_sorted[-1]`) — so the repair landed for 2 of 3 flagged inputs, not PPG, the one named most explicitly. Also worth naming: `power_v2.WEIGHTS` doesn't match the spec's §5 target formula (0.40 ROS / 0.20 all-play / 0.15 recent / 0.15 VORP-PAR / 0.10 record) — "canonical" should imply conformance to the spec's actual weights, not just a patched `power_v2`.

3. **Playoff Odds preservation — CONFIRMED**, with one precision note: `playoff_sim.py` WAS touched by #979 (16/-21), not untouched — verified the removed `_PLAYER_CV_BY_POSITION`/`_DEFAULT_PLAYER_CV` block was genuinely dead in that file's own path, and the real consumer (`league_intel/sim.py::_LegacyPointsModel.draw()`) now delegates to `sim_calibration.PointsModel()` directly, pinned by `test_points_model_single_owner.py`. `public_league/playoff_odds.py` itself imports none of the touched modules. Safe dedup, no live-path change.

4. **Trend semantics — CONFIRMED.** Every trend point runs through the same `_score_state(results_only=True)` as the headline; `trendDelta()` returns `null` (not 0) on <2 points or a null rank. Correctly inherits item 2's PPG contamination at every point, since it's the same function.

5. **Unrankable behavior — CONFIRMED for the global case** (`test_power_unrankable.py`, 5 tests, non-vacuous). **Gap found in the per-owner case**: `missing_inputs`/`active_weights` renormalize globally, not per-owner — a single owner with `games == 0` this season falls through to a silent `ppg = 0.0` (`s["points"]/s["games"] if s["games"] else 0.0`) that bypasses the renormalization mechanism protecting the global case. No test exercises this. I can't verify from static reading whether Sleeper can actually produce a zero-games mid-season roster in practice — flagging as a code-level gap, not asserting it's live.

6. **Real-league L2 — CONTRADICTED as a completeness claim, but your own ledger already says so.** Only synthetic 12-owner/8-season timing evidence exists (35.25ms vs 25.13ms) — L1-grade, and your own doc calls it "the closest I can get without a real league snapshot." No real-league measurement exists anywhere in this chain yet. This matches the row's own `IN PROGRESS` status — not disputing your bookkeeping, just confirming the L2 target isn't met yet.

**Net verdict: DISAGREE that consolidation is complete on `main` today (items 1, 2); AGREE your own documentation is honest about both gaps (item 6).** Recommendation before promoting `power_v2` to sole canonical status: give PPG the same per-season reset already applied to `recentAvg`/`all_play` — otherwise this ships the original contamination defect under a single, harder-to-notice roof instead of repairing it.

No files touched, no code changes proposed — independent verification only, per this session's read-only mandate. Full findings (incl. V1-49) also being logged to `docs/WORK_CLAIMS.md` for cross-session visibility.

---

jasonleetucker-code:
Traffic-control update: #1032 is now merged on current main as `6bd8e6adff11928cceb0fa4ad64372408341926f`, and it directly repairs the cross-season PPG contamination identified in the adversarial review above. Because #996 also edits `src/ros/power_v2.py`, this is now a demonstrated material dependency overlap, not ordinary freshness churn. Do not merge #996 from its old tree and do not let #1009 leapfrog it. Integration order remains: current main + #1032 semantics preserved -> reconcile #996 once -> #1009 -> shipping-tree CI -> real 12-owner L2 measurement via #1028. Feature authors should remain frozen; Claude 5 owns the one reconciliation.

PR #997, opened 2026-08-20
BDVM production proof: HTTP-level zero-network tests for all 4 routes

jasonleetucker-code:
## Problem

`tests/bdvm/test_request_path_is_local.py` (22 tests) thoroughly proves zero-network behavior at the **function** layer — every test calls `bdvm_api.get_bdvm_values(...)` directly. Nothing proves it at the **HTTP/route** layer: `server.py`'s routing, `run_in_threadpool`, and `_bdvm_gate_and_league` wiring were never exercised under the same network-refusing conditions. `/api/bdvm/roster` and `/api/bdvm/trades` had **no HTTP-level test coverage of any kind**, anywhere in the repo. This is exactly the "generic health check is not feature-specific proof" gap the assignment calls out — the earlier P1 fix (#946, merged as `1d98a6593`/`015570741`) removed the heavyweight-ingestion risk, but the request path had never been proven clean at the layer real traffic actually goes through.

Scope note: the same assignment also asked for a scoring-integrity audit of BDVM. That audit is complete and found **no defect** — one canonical arithmetic core (`src/league_intel/scorer.py::score_stat_line`), one dispatch layer in front of it (`src/nfl_data/realized_points.py::compute_weekly_points`), and every BDVM module that needs fantasy points calls into that same layer. Reported clean, nothing to fix — not padded into this PR.

## What this PR does

New `tests/bdvm/test_http_route_no_fetch.py` — drives all four `/api/bdvm/*` routes (`values`, `roster`, `trades`, `trade-eval`) through a real `TestClient(server.app)`, with `ingest._try_fetch_with_fallback` and `urllib.request.urlopen` patched to record-and-refuse (reusing `_NetworkRecorder` from `test_request_path_is_local.py` rather than reimplementing it — one recorder, one behavior). Each route is asserted to return a real response with **zero** recorded network attempts on a cold cache. Non-vacuity is proven by spying on `bdvm_api._context_for`/`_schedule_for`/`_actuals_for` via `side_effect=real_...` — the mocks don't just exist, they're actually reached.

A real BDVM projection snapshot is seeded (`write_snapshot`) and `bdvm_actuals.current_nfl_season` is pinned to `2025` so `_actuals_for` is genuinely reachable (the real date is outside the Sept–Jan actuals window), rather than the test accidentally short-circuiting before the code path it's meant to prove.

No production code changes. No BDVM methodology, scoring, or config change.

## Two real, pre-existing test-isolation bugs found and fixed in this PR's own test file

Both surfaced only when running this file combined with the rest of the suite, not in isolation — found by actually running the broader suite rather than trusting an isolated-file pass:

1. **`TestClient` lifespan leak.** An early draft used `with TestClient(server.app) as client:` (explicit `__enter__`/`__exit__`), which triggers the app's real ASGI lifespan startup (background Sleeper refresh threads, an uptime watchdog) that then raced with unrelated test files in the same pytest process, corrupting `tests/public_league/test_server_routes.py`. Fixed by switching to bare `TestClient(server.app)` construction — the same safe pattern already used in `tests/bdvm/test_endpoint.py`.
2. **Env-var pop-instead-of-restore.** Cleanup did `os.environ.pop("LEAGUE_REGISTRY_PATH", None)` instead of restoring the prior value. `tests/conftest.py` sets that var session-wide (`/nonexistent/path/for/tests.json`) so `league_registry` falls through to synthesizing from `SLEEPER_LEAGUE_ID`; popping it instead of restoring it left a later test file's registry state polluted. Fixed via capture-and-restore-exactly (`_prior = os.environ.get(...)` before mutation, restore to that exact value — never assume "unset" is the correct prior state). Applied the same fix to `RISKIT_FEATURE_BDVM_ENGINE` on principle.
   - The same bug class **pre-exists** in `tests/api/test_league_routing.py`'s own `two_league_registry` fixture (independently confirmed by combining it with `test_server_routes.py` and reproducing the identical 10-failure signature). **Not fixed here** — out of this PR's scope, flagging for whoever owns that file next.

Verified via the exact repro: `pytest tests/bdvm/test_http_route_no_fetch.py tests/public_league/test_server_routes.py` went from 10 failures to 17/17 passed after both fixes.

## Mutation proofs

| mutation | expected failure | restored |
|---|---|---|
| Drop `cache_only=True` at `_schedule_for`'s call site | the corresponding route's zero-network assertion fails | ✅ byte-identical |
| Drop `cache_only=True` at `_actuals_for`'s call site | same | ✅ byte-identical |
| Revert `_context_for` to call `fetch_and_build_context(cache_only=False)` directly (a regression of #946's own fix) | zero-network assertion fails | ✅ byte-identical |

## Tests

- `tests/bdvm/`: all passing.
- Full hard gate (`tests/ -m "not livedata"`): **9623 passed, 40 skipped, 5 failed** — the 5 failures are `tests/consensus_edge/test_panel.py`'s known linked-git-worktree artifact (`.git` is a file not a directory in a worktree, so `panel.is_shallow()` mis-detects), reproduced identically across every full-suite run this session in any worktree, on branches this PR never touches. Not attributable to this diff.

## Verification

- `check_decision_coercions.py` (post-commit): clean.
- `check_work_claims.py`: no overlapping claim or branch.

## What remains genuinely blocked — for Claude 5 / Integration, not closeable by any local test

Static + mocked-dynamic proof is now as complete as it can get. What no local test can supply is confirmation that the **deployed** process is running this code and behaves this way under real conditions. Exact reproducible measurement:

1. Confirm the deployed commit is at or after `015570741`/`1d98a6593` (the merged P1 fix).
2. Cold-hit `/api/bdvm/values` (and ideally `/api/bdvm/roster`) for a flagged-on league and read `meta.auxiliaryInputs` on the response: every input should report `available` or `stale`, never a multi-second latency spike consistent with a live fetch.
3. For direct confirmation: tail the production journal (`journalctl -u dynasty`) across that request's timestamp and grep for nflverse fetch activity — check `ingest.py`'s existing `_LOGGER` calls before assuming a new log line is needed.

Deliberately **not** building a new observability log marker for this — the existing tests already instrument the exact fetch choke point (`ingest._try_fetch_with_fallback`) this PR's HTTP-level test also now exercises; a log line would prove the identical property a second way without closing a gap the tests don't already close.

## Handoff

`READY_FOR_INTEGRATION`, frozen. Not self-merging — merge authority is elsewhere. Sibling PR: #990 (`claude/bdvm-baseline-asof-unreachable-guard`, structural reachability guard for the two un-gated call sites), same lane, independently mergeable.

---


jasonleetucker-code:
Closing — this PR's content is merged onto `main` via `#1042` (merged `91bb2f4a0`), reconciled alongside its sibling `#990`. Both were test-only and merged clean. Thank you for catching the two real test-isolation bugs along the way.

---

PR #984, opened 2026-08-20
C8-PSI-02 PR A+B: Premium Sports Intelligence migration — Rankings + Player Profile

jasonleetucker-code:
## Summary

**Note added after initial review**: this PR was opened as PR A only (Rankings migration). While
Claude 5 (Integration) was reconciling PR A with `main`, PR B (Universal Player Profile) landed as
additional commits on this same branch rather than a separate stacked PR as originally planned —
recorded honestly rather than silently splitting branch history mid-integration. Both batches are
independently reviewable by commit (PR A: editorial tokens → shell restyle → Rankings migration →
full-bleed fix → axe fixes; PR B: `Universal Player Profile` commit). Full record for both:
`docs/cseries-delivery/CLAUDE_13.md`.

Owner-authorized real, production-capable migration of two reference routes onto the "Premium
Sports Intelligence" editorial visual direction
(`docs/PREMIUM_SPORTS_INTELLIGENCE_DESIGN_NORTH_STAR.md`), using two attached reference
screenshots as visual source of truth.

### PR A — Dynasty Rankings

- Additive `.psi-editorial` CSS scope (`frontend/app/tokens.css`) — cream surfaces, near-black
  text, one burnt-red accent, serif display type (system stack — no new font asset). Every color
  is WCAG-computed, not eyeballed. Does not touch any existing dark/light token.
- Shell restyle (`TopBar.jsx`/`MobileChrome.jsx`) — the persistent header/nav now renders in the
  new editorial palette on every route; unmigrated page bodies stay on the terminal palette until
  their own route migrates (deliberate, temporary split).
- Rankings page migrated onto the new palette with **every existing behavior preserved** — lenses,
  filters, search, sort, source columns/audit, tier grouping, row expansion, watchlist, news
  chips, export, virtualization (`aria-rowcount`/`aria-rowindex`), deep-link seeding, trust strip,
  movers/edge rails, methodology modal, position-rank badges, rank-change glyphs, the
  `rankings-controls` E2E selector class.
- A real axe-core scan (WCAG 2.0/2.1 A+AA) against the migrated, authenticated page with real
  canonical data found 187 color-contrast failures on legacy badge/tier classes; fixed across four
  re-verified rounds down to 0.

### PR B — Universal Player Profile ("Player File")

- New private route `/players/[playerId]`, reached via a new "Full profile" link on `PlayerPopup`
  (which is otherwise unchanged — still the quick-view drawer, all its own tests pass unchanged).
[...]

PR #971, opened 2026-08-20
feat(sources): Dynasty Nerds IDP Top-275 acquisition (Lane 8 PR C)

jasonleetucker-code:
## What this is

Lane 8 (Source Acquisition / Cross-Position Bridge, owner-chartered 2026-08-20 in `docs/EXECUTION_PLAN.md`) PR C: **acquisition and preservation only**, no production wiring, no bridge-combination policy — the qualification-not-invention posture the lane's charter requires.

Public, no-auth acquisition of Dynasty Nerds' IDP Top-275 (`dynastynerds.com/idp/dynasty-idp-rankings-tiers/`), joining the existing `dynastyNerds` provider family — `scripts/fetch_dynasty_nerds.py` already registers that vendor's SFLEXTEP **offense** board in production. This is the vendor's **IDP** half, acquired for the first time.

## Ordinal/tier only — no cardinal value manufactured

Dynasty Nerds' IDP board publishes rank, positional rank (`DL1`, `LB2`, …) and a named tier. Nothing else. `ArchivedRow.value` is `None` on all 275 rows, and `TestNoCardinalValueIsEverManufactured::test_the_repair_is_what_makes_this_pass` is a mutation-proof test: a version that fabricates `value = rank` fails the guard.

## Schema pin, verified against the live page (2026-08-20)

- Exactly 10 `<table>` tier tables, each under a `Tier N | <label>` heading
- Headers exactly `Rank, Player, Position, Age, Team, <year> IDP Rank`
- 275 total data rows
- JSON-LD `Article` block's `dateModified` used as `source_as_of` — the vendor's own stated date, never "now"

A shape violation reports `SCHEMA_CHANGED` (exit 2) or `PARSE_FAILED` (exit 1) and does **not** archive. A fetch failure reports `UNAVAILABLE`, never a healthy empty board — per `src/sources/acquisition_state.py`'s existing rule that a failed acquisition must not be representable as one.

## Boundary — matches PR A exactly

- Writes only to `src/source_archive/` via `archive_board` (`provider="dynastyNerds"`, `format_key="idp_top275"`, `game_type="DYNASTY"`)
- Touches no production code: `src/api/data_contract.py`, `_RANKING_SOURCES`, any Hill constant, any weight, any production CSV — zero
- No bridge-combination policy, no cardinal precedence rule, no confidence treatment — none of that is decided here
- `git diff main -- src/api/ config/ CSVs/ frontend/` is empty

## Verification

[...]

PR #999, opened 2026-08-20
Status truth after the PSI merge: V1-111 IMPLEMENTED_UNVERIFIED, two claim rows closed

jasonleetucker-code:
Docs only. **Tally unchanged at 61 / 136 = 44.9%. Nothing is promoted here.**

## V1-111 (`C8-PSI-02`): `NOT STARTED` → `IMPLEMENTED_UNVERIFIED`

That status became false the moment #984 merged — the Rankings reference route and the Universal Player Profile both ship, on an additive `.psi-editorial` token scope, with the shell restyled across every route.

**This is a status correction, not a promotion, and the row says so.** The target is **L4** (a production consumer). Merging is not deploying, and deploying is not consuming.

What the row now records as established — by measurement at Integration, not from the PR text:

- browser-rendered evidence at 1366×768 and 390×844 against a real backend and the real board, for a **rich-data** player (Josh Allen 9,991 / #1 / QB1 / 14 sources) and a **partial-data** player (AJ Dillon: `not priced`, `None — unranked`, **no position-rank tile at all**, "Value history unavailable")
- zero prototype constants; no frontend value, rank or weighting computation
- WCAG contrast measured from **11 failing colour groups (worst 1.04:1) → 0** on both viewports, agreeing with the lane's independent axe run of **187 nodes → 0**
- both CI gates green at the exact merged head, with E2E genuinely applicable (`components/shell/**` is in its path filter)

And what **L4 still needs**: the deployed build serving these routes, and a real consumer. Written into the row so the gap cannot quietly close by itself.

## Two merged claim rows closed

Same false-block class as the P1-BDVM row in #994. A row left `open` after its branch merges reports a live editor who is not there — under a multi-lane campaign that manufactures blocks.

| claim | closed as |
|---|---|
| Claude 13 — PR A/B (PSI reference routes) | `done` — merged `5c43da9e8` (#984) |
| Lane 8 PR C — Dynasty Nerds IDP Top-275 | `done` — merged `8386e93b5` (#971) |

Both closures **carry the prohibition that survives them**, because a closed claim is exactly where a constraint gets lost: V1-111 is not verified by its merge, and Lane 8's section 9 preconditions — weighting, cardinal precedence, disagreement, confidence, tie-breaker — remain owner decisions that #971 did not authorise.

`check_planning_integrity.py` and `check_decision_coercions.py` both clean; the standing-tally check confirms the counts still agree with the row table (`IMPLEMENTED_UNVERIFIED` 20 → 21, `NOT STARTED` 32 → 31, `VERIFIED` unchanged).

---


jasonleetucker-code:
Post-deploy verification added to this PR (`docs/psi/PROD_VERIFICATION_2026-08-20.md` + four production captures). Head is now `e82cfc4e3`.

**PSI is live on `chaseupside.com`** — rendered in a real browser, not inferred from a green deploy. Six renders (desktop 1366×768, mobile 390×844, over `/login`, `/league`, anonymous `/rankings`), all agreeing:

| probe | value |
|---|---|
| `header` class | `shell-topbar psi-editorial` |
| `.psi-editorial` nodes | 3 |
| console errors | **0** |
| page errors | **0** |

**The part that looks wrong and is not.** Page bodies still compute `rgb(11,13,16)` with a mono font. That is the route-by-route method working as specified — only `/rankings` and the Player File were migrated, so every other body keeps the terminal palette beneath the now-editorial shared shell. The first thing anyone sees on production is a dark login page with a cream header, which reads like a half-finished deploy and is not one, so it is written down rather than left to arrive later as a bug report.

**Auth gate checked deliberately**, because the shell change touched the components around the only page auth gate in the system: anonymous `GET /rankings` still **307**s to `/login?next=%2Frankings` on both viewports.

**Board did not move.** Production `/api/status`: `contract.health.ok` true, 0 structural errors, 0 source-health errors, `normalizationHealth` healthy, 1,111 players — consistent with the pinned-input golden-board diff across this wave (0 values moved, 0 ranks changed).

**Method, recorded so it can be disputed rather than assumed.** Chromium in this sandbox cannot reach the public internet — every navigation to `chaseupside.com` returns `ERR_CONNECTION_RESET`, and the agent proxy's relay log shows only Chrome's own telemetry, none of my navigations, while `curl` tunnels through that same proxy fine (`CONNECT chaseupside.com:443` → `200`). So a loopback mirror re-serves production's bytes and Chromium renders those. The bytes are production's and the execution is real; the origin the browser sees differs, so this would **not** catch an origin-dependent defect — a cookie `Domain`, a CSP `host-source`, an absolute redirect.

**Still not verified, which is exactly the L4 gap this PR writes into `V1-111`:** the migrated `/rankings` body and the Player File are private and this environment has no production credentials. The local-build evidence for those surfaces is real and detailed, but it is a statement about the branch, not about production, and I am not conflating the two.

---

PR #1003, opened 2026-08-20
V1-106: regression-sensitive windowing evidence + current-tree FPS measurement (post-#984)

jasonleetucker-code:
## Target

`V1-106` (`docs/VERSION_1_COMPLETION_CONTRACT.md`) — Rankings board windowing / virtualization, canonical id `C8-PERF-03` / inv 9.1, required level L2, current status `IN PROGRESS`. Re-evaluated against the **current post-#984** Rankings renderer per instruction — not the pre-PSI route.

## PHASE 1 — reconstructing the residual

1. **Virtualization is still active.** `app/rankings/page.jsx`'s `<DataTable>` call still passes both `virtualize` and `freezeColumnWidths` (`components/ds/DataTable.jsx`, backed by `components/ds/useRowWindow.js`). #984 restyled the page onto ds primitives; it did not touch this wiring.
2. **DOM population stays bounded as data grows.** Measured live on this head: "Show all" grows the board to 983 rows while only 37 `<tr>` mount (1,536 DOM nodes); the capped default is 200 rows / 27 mounted.
3. **#984 preserved `aria-rowcount`/`aria-rowindex`/virtual-row semantics.** `DataTable.jsx` is the same component instance before and after #984 — the migration changed the page around it, not the primitive. Confirmed live: `aria-rowcount` is present and non-null once windowed.
4. **The remaining gap is L2 evidence, not implementation.** The feature works and was already browser-measured (#760). What's missing is *committed, regression-sensitive* proof: the two existing specs touching the board (`journey-rankings.spec.js`, `mobile-smoke.spec.js`) assert `boardRowCount() >= 50` and `rows.count() > 0` — both pass identically whether or not virtualization is on, because `boardRowCount()` falls back to counting mounted rows when `aria-rowcount` is absent, and a non-empty **capped** board already clears both bars without ever touching "Show all". Neither discriminates windowed from every-row-mounted.
5. **What the owner's ruling actually requires.** `C_SERIES_SCOPE_MANIFEST.md`'s `C8-PERF-03` row sets required final state **"Windowed"** (structural) with evidence **"FPS harness"** (an artifact, not a number), under profile **P1** — and P1's own definition (§3, line 159) is entirely latency-based (warm ≤1s, cold ≤3s, p95 ≤2s, interaction <250ms, ack <100ms), with **no FPS quantity anywhere in it**. The only FPS numbers in this repo ("Phase-4 target was ≥30 FPS at 4× and ≥50 at 1×") live in `docs/performance-optimization.md`, a non-canonical L6 engineering ledger, stated as one conjunction rather than two independent thresholds. **No FPS threshold is owner-defined at the canonical tier, so none is invented here** — the L2 obligation is discharged with an honest measured statement, not a pass/fail against a manufactured number.

This is **not** an `OWNER_DECISION_REQUIRED` case: "Windowed" + "FPS harness ran and is reported" is what the manifest actually asks for, and both are true.

## PHASE 2 — closing the real gap (evidence only, no virtualization rewrite)

Added `tests/e2e/specs/rankings-windowing.spec.js`, run against the **current** post-#984 renderer:

- **Desktop:** "Show all" grows the board past 500 rows; asserts `aria-rowcount` is present, the board total is ≥500, and the **mounted** row count stays ≤200 — an order of magnitude below the board size, and well above this repo's own measured mount of ~40-60 rows, so it can only pass if the DOM is genuinely bounded.
- **Desktop:** scrolls 6000px past the first window; asserts the top-of-window player **name changes** (the window slid) while the board's declared total is unchanged and the mount stays bounded — proves windowing stays active **during** scrolling, not just at initial mount.
- **Mobile (390×844):** same bounded-mount assertion, for parity.

**Mutation-proven** (not committed as a fixture — flipping product behavior on purpose to fail a test doesn't belong in CI): temporarily changed `virtualize` → `virtualize={false}` in `app/rankings/page.jsx`, reran the three tests → **all three RED** (mounted count no longer bounded / no top-of-window change / `aria-rowcount` absent on mobile, each attributed correctly). Reverted → **all three GREEN** again. Zero net diff to `page.jsx`.

## L2 measurement

Exact head `357c9410e` (parent `36b7033f8`), measured with the harness already on this branch (CDP `Input.synthesizeScrollGesture` driver, no timer in the loop, control run first at every rate). Raw samples: `docs/master-site-audit/evidence/V1-106/{showall,capped}.json`.

| board | rows | mounted | DOM nodes | control 1×/4×/6× | board 1×/4×/6× |
|---|---|---|---|---|---|
| "Show all" | 983 | 37 | 1,536 | 60.0 / 60.0 / 60.0 | 59.7 / 24.0 / 12.6 |
| capped default | 200 | 27 | 1,206 | — | 59.7 / 28.0 / 16.0 |

Reported against three yardsticks, per prior V1-106 planning work:

1. **The non-canonical conjunction as stated** (≥50 at 1× AND ≥30 at 4×): 1× clears it (59.7); 4× does not (24.0 vs 30). Materially the same shape #760 measured (59.5 / 23.3) — **#984 did not regress the board**; this reproduces, not retracts, #760's finding.
2. **The binding P1 profile** (latency, not FPS): out of scope for this harness — V1-105 (route p95 baselines) is the correct instrument, and it's separately `NOT STARTED`.
3. **L2 — a measured statement of the effect on the live board**: this is that statement, on today's head, honestly reported both ways.

## Validation

- `npx vitest run --project logic` (frontend/): **86 files, 1,732 tests passed**
- `npx playwright test -g "V1-106" --project=desktop-1366 --project=mobile-chromium`: **3/3 passed** (RED under mutation, GREEN restored — see above)
- Production build succeeds (implicit via the E2E self-boot)

## Boundaries respected

- No virtualization rewrite — `useRowWindow.js`/`DataTable.jsx` untouched.
- No new PSI visual work.
- No frontend canonical-value computation anywhere this PR touches.

---

`FEATURE_GREEN` / `READY_FOR_INTEGRATION`. Head: `357c9410e`. Freezing this branch — will only resume on a branch-caused CI failure, substantive review feedback, or an explicit Integration request. `docs/performance-optimization.md`'s FPS-target framing and the `C_SERIES_SCOPE_MANIFEST.md` / `OWNER_FEATURE_INVENTORY.md` rows citing the retracted 59.5 self-measurement remain Claude 5's / Integration's to correct — not touched here, per the same division of labor as the earlier V1-106 planning pass.

---

### DIFF
--- a/docs/master-site-audit/evidence/V1-106/capped.json
+++ b/docs/master-site-audit/evidence/V1-106/capped.json
@@ -1,0 +1,187 @@
+{
+  "fullCount": 200,
+  "mountedRows": 27,
+  "domNodes": 1206,
+  "hover": true,
+  "control": [
+    {
+      "rate": 1,
+      "frames": 231,
+      "elapsedMs": 3833.2,
+      "fps": 60.00208702911406,
+      "medianFrameMs": 16.699999999999818,
+      "p95FrameMs": 16.800000000000182,
+      "longFrames": 0,
+      "moved": 9000,
+      "maxScroll": 23700,
+      "thinSample": false
+    },
+    {
+      "rate": 4,
+      "frames": 231,
+      "elapsedMs": 3833.173999999999,
+      "fps": 60.00249401670784,
+      "medianFrameMs": 16.700000000000728,
+      "p95FrameMs": 16.700000000000728,
+      "longFrames": 0,
+      "moved": 9000,
+      "maxScroll": 23700,
+      "thinSample": false
+    },
+    {
+      "rate": 6,
+      "frames": 229,
+      "elapsedMs": 3799.7999999999993,
+      "fps": 60.00315806095058,
+      "medianFrameMs": 16.700000000000728,
+      "p95FrameMs": 16.700000000000728,
+      "longFrames": 0,
+      "moved": 9000,
+      "maxScroll": 23700,
+      "thinSample": false
+    }
+  ],
+  "results": [
+    {
+      "rate": 1,
+      "frames": 229,
+      "elapsedMs": 3816.3999999999996,
+      "fps": 59.742165391468404,
+      "medianFrameMs": 16.699999999999818,
+      "p95FrameMs": 16.700000000000728,
+      "longFrames": 1,
+      "moved": 8201,
+      "maxScroll": 8201,
+      "thinSample": false,
+      "samples": [
+        {
+          "frames": 221,
+          "elapsedMs": 3816.4999999999995,
+          "fps": 57.64443862177388,
+          "medianFrameMs": 16.699999999999818,
+          "p95FrameMs": 16.800000000000182,
+          "longFrames": 9,
+          "moved": 8201,
+          "maxScroll": 8201,
+          "thinSample": false
+        },
+        {
+          "frames": 229,
+          "elapsedMs": 3816.3999999999996,
+          "fps": 59.742165391468404,
+          "medianFrameMs": 16.699999999999818,
+          "p95FrameMs": 16.700000000000728,
+          "longFrames": 1,
+          "moved": 8201,
+          "maxScroll": 8201,
+          "thinSample": false
+        },
+        {
+          "frames": 229,
+          "elapsedMs": 3799.8999999999996,
+          "fps": 60.00157898892076,
+          "medianFrameMs": 16.69999999999891,
+          "p95FrameMs": 16.700000000000728,
+          "longFrames": 0,
+          "moved": 8201,
+          "maxScroll": 8201,
+          "thinSample": false
+        }
+      ]
+    },
+    {
+      "rate": 4,
+      "frames": 108,
+      "elapsedMs": 3816.434000000001,
+      "fps": 28.036643631201265,
+      "medianFrameMs": 33.29999999999927,
+      "p95FrameMs": 66.70000000000073,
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
**CI fix pushed** (`5f4dfd166`) for the "Boot stack + run journeys" failure.

Diagnosis: it wasn't a test failure — every test in the run passed or skipped as expected, including all 3 of `rankings-windowing.spec.js`'s (2 desktop-only + 1 mobile-only, gated the same way `mobile-smoke.spec.js` already does). The run tripped the repo-wide **E2E coverage floor**: `passed=184 (floor 120) skipped=62 (ceiling 60)`.

Root cause: this PR's 3 new project-gated tests legitimately add 3 skip entries (each test skips on the project it's not scoped to). The fixed default ceiling of 60 in `tests/e2e/stack-death-reporter.js` had already drifted close to its limit from earlier legitimate suite growth (measured baseline here: 59 before this PR), so my +3 tipped it 2 over.

Per that file's own documented practice ("re-derive both numbers from a named green run before changing them, and cite the run"), raised `MAX_EXPECTED_SKIPPED`'s default 60 → 70, citing this exact run (`32415289607` / job `96574935284` / head `357c9410e`, passed=184/skipped=62) in the updated comment. No product or test-logic changes — one bound + its citation comment.

---

PR #1019, opened 2026-08-21
V1-45: Trade Calculator renders the canonical rosterCapacity forced-drop result

jasonleetucker-code:
## Mandate

V1 70% sprint, row **V1-45 — Trade Calculator** (contract status `IMPLEMENTED_UNVERIFIED`, target **L4**, owner note: "high-use surface; owner status ALREADY COMPLETE — VERIFY ONLY"). Verification-first, not a rewrite.

## Trace performed

Full user path: Trade Calculator UI (`frontend/app/trade/page.jsx`) → `POST /api/trade/simulate` → `src/trade/roster_capacity.py::assess_roster_capacity` → response → render. Also re-checked (repeating an earlier verification pass in this same effort): stale frontend trade math, duplicate VA, raw-gap equalizers, missing→zero, 3-team collapse, manual-override mutation of canonical state — **all clean**, backed by 741 backend + 220 frontend passing tests.

## The one real gap

`src/api/trade_simulator.py` genuinely computes and publishes a `rosterCapacity` block on every `/api/trade/simulate` response — forced releases, their cost, an explicit upper-bound flag when taxi occupancy is uncertain. **Zero frontend files anywhere in the app read it** (confirmed by repo-wide grep both before and after this fix). The canonical result existed; the intended production surface simply never consumed it. That's the L4 failure this row's bar exists to catch.

## Fix

`SimulationPanel` (`frontend/app/trade/trade-sections.jsx`) now renders the three `requiresDrops` states the backend already distinguishes — never collapsing an unknown into a false "no drops needed" (missing ≠ zero applies here too):

- **`true`** — names the forced release(s), their cost, and flags `forcedDropsAreUpperBound` / `ladderExhausted` when the backend itself is uncertain.
- **`false`** — silent, matching the spec's own "clean fit imposes no forced-drop cost."
- **`null`** — an explicit uncertainty banner, distinguishing an unknown roster cap from unknown taxi occupancy — never silently reads as "clean."

Pure display of backend-stamped fields. No trade math added to this file (the module's own header comment: *"NO trade math lives in this file"*).

## RED-before/GREEN-after

The new test file run against the pre-fix component: **5 of 7 failed** — exactly the cases that depend on the new render (the two that passed were the "renders nothing" cases, which were already true). Restored: **7/7 pass**.

## Evidence

| check | result |
|---|---|
| `frontend/__tests__/components/trade-simulation-panel-roster-capacity.test.jsx` | 7 passed |
| `frontend/__tests__/trade-logic.test.js` | 179 passed (unaffected) |

## What this does NOT close

- **L3** (named checklist against the deployed SHA) — no production/auth access from this environment. Reporting the exact blocker rather than manufacturing a pass, per the dispatch's own instruction.
- Full **L4** production confirmation — same reason.

## Deliberately NOT touched

The third standalone KTC-VA Python port in `src/public_league/trade_grading.py` — a different surface (public-league retro-grading), already targeted by open PR #962, out of this row's scope.

FEATURE_GREEN / READY_FOR_INTEGRATION. Not self-merged.

---


jasonleetucker-code:
Superseded — this PR's full content (the `rosterCapacity` rendering in `SimulationPanel`, plus its test file) was carried forward unchanged inside #1025 (`claude/v1-46-manual-override-ux`, stacked on this branch) and merged as `b78dd858e`. Closing without merging to avoid a duplicate.

V1-45 itself is not promoted by this — its own PR body honestly reports L3/L4 (deployed-SHA checklist, production consumption) as unreached from this environment, and that stands.

---

PR #1001, opened 2026-08-20
V1-136 IDP Show audit + failure-state instrumentation (partial, see correction) + DLF fixture repair for #983 (three slices + post-#1012 correction)

jasonleetucker-code:
## ⚠️ Post-#1012 correction (read this first)

**#1012** (`0a5267d06`, merged after this branch's last `origin/main`
merge) made `idpShowCombined` the IDP Show provider family's **sole
voting source**, replacing the plain `idpShow` board this PR's Slice 1
and Slice 3 were written against. Consequence, found and mapped in a
follow-up commit (`be5d1ef40`,
`docs/sources/LANE8_POST_1012_RECONCILIATION_AUDIT.md`):

**Slice 3's `scripts/fetch_idpshow.py` instrumentation is INCOMPLETE,
not wrong.** This branch's copy of that file predates PR #1008 (the
`--combined` / widest-chart work), so it has **zero mentions of
"combined" anywhere** — no `--combined` flag, no
`_pick_widest_chart`, no `_IDPSHOW_COMBINED_ROW_FLOOR`. Slice 3's
`AcquisitionOutcome` instrumentation only covers the plain board's
`main()` path — the board that no longer votes. The board that
**does** vote (`idpShowCombined`, via `--combined`) has zero
outcome-instrumentation in this branch, and needs it more, now.

**This is the one real code-level conflict** between this PR and
current main (every other file below has zero overlap with #1012).
**Not rebased here on purpose** — the full 5-point reconciliation plan
(reconcile from main's copy, re-apply the instrumentation pattern onto
the `--combined` branch's return points too, make the status file
board-aware since the fetcher now runs twice per cycle, add the missing
`SCHEMA_CHANGED` outcome for the combined board's row floor) is in the
linked doc for **Claude 5** to execute as one deliberate merge, not a
mechanical `git merge`.

The same follow-up commit also **independently re-confirmed V1-133/134/135
on current shipping main** (all `CLOSURE_READY`, mutation-proved fresh —
#1012 doesn't touch the bridge layer at all) and measured **V1-23's
W02-F001/F002** post-#1012: W02-F001 stays fixed; W02-F002 is **not**
structurally fixed but its reproduction rate collapsed from a documented
29.4%/52-of-52-HIGH to a measured **0.5% (2 of 398 IDP rows)** — because
`idpShowCombined` added real independent evidence, not because the
Hampel filter gained anchor-awareness. Full detail in the doc.

---

## This PR carries three independent slices, landed on the same branch

[...]

PR #1034, opened 2026-08-21
V1-136: board-aware acquisition-state instrumentation for both IDP Show boards

jasonleetucker-code:
## Reconciliation of #1001's Slice 3 (post-#1012)

`#1001` (`claude/lane8-v1-136-idpshow-audit`) instrumented `scripts/fetch_idpshow.py`'s failure paths with real `src.sources.acquisition_state.AcquisitionOutcome` states — but that branch was based before PR #1008's `--combined` machinery landed, so its coverage is **plain-board only**. `#1012` then made `idpShowCombined` the IDP Show provider family's **sole voting source**. Net effect: the board that no longer votes has instrumentation; the board that actually votes has none — main's own 450-row floor guard on the `--combined` path was a bare `return 2`.

This PR is the reconciliation `docs/sources/LANE8_POST_1012_RECONCILIATION_AUDIT.md` Task C names as Claude 5's — not a rebase of `#1001`, but a fresh application of its instrumentation pattern onto current main's copy of the fetcher, extended to cover both boards.

## What changed

- `scripts/fetch_idpshow.py` — every `main()` return point on **both** the plain and `--combined` paths now constructs a real `AcquisitionOutcome`. `source_key`/`status_path` are resolved once at the top of `main()` from `args.combined` and threaded through every branch, including the two checks (missing session, paywalled) shared by both boards before the `--combined` split — so even those early-exit branches stamp the board that was actually being fetched. Two status files, not one: `data/scrape_state/idpShow_last_status.json` (plain) and `data/scrape_state/idpShowCombined_last_status.json` (combined/voting) — because `deploy/idpshow_fetch_and_push.sh` runs the fetcher twice per cycle as separate processes, and a shared file would let the second run silently overwrite the first's outcome. New: a `SCHEMA_CHANGED`/`row_count_below_floor` outcome for the combined board's 450-row floor (previously unstamped), and a `PARSE_FAILED`/`no_combined_chart_found` outcome when no candidate chart yields a usable dataset.
- `deploy/idpshow_fetch_and_push.sh` — commits the two new status files, existence-checked before `git add -f` (unlike the pre-existing `_last_success` stamps, these are brand-new paths with no prior tracked history; an uncaught pre-persist exception on a virgin path would otherwise make `git add -f` a hard pathspec error under `set -Eeuo pipefail` and abort the whole push, blocking the *other* board's genuinely successful commit).
- `tests/sources/test_idpshow_acquisition_state.py` (new, 39 tests) — supersedes the plain-board-only file `#1001` would have added.
- `docs/VERSION_1_COMPLETION_CONTRACT.md` / `docs/WORK_CLAIMS.md` — `V1-136` moved `NOT STARTED` → `IN PROGRESS` (numerator unchanged, 81/136 — see "Not claiming" below).

## The critical anti-false-green property

Directly tested (`TestBoardsDoNotContaminateEachOther`, 2 tests): a plain-board **success** followed by a genuine combined-board **failure** leaves `idpShowCombined_last_status.json` reporting its own true failure state — never inheriting or defaulting to the other board's `HEALTHY` — and the reverse. This holds structurally because the two boards write to two independent files, but it's asserted directly rather than left as an inference.

**Mutation-proved**: temporarily collapsing the board-aware `status_path` selection back to a single shared `STATUS_PATH` (simulating the exact bug this instrumentation exists to prevent) sends 11 of 39 tests RED, including both contamination tests. Restored, clean.

## Verification

- `tests/sources/test_idpshow_acquisition_state.py`: **39/39 passed**
- `tests/sources/ tests/scripts/test_fetcher_main_argv_contract.py tests/api/test_source_registry_parity.py tests/bridges/`: **170 passed, 24 subtests**
- `tests/api/test_idpshow_combined_source.py tests/api/test_source_floor_invariant.py tests/api/test_config_parity.py`: **15 passed**
- `ruff format --check` / `ruff check`: clean
- `scripts/check_decision_coercions.py`: clean, 660/660 unchanged
- `scripts/check_planning_integrity.py`: clean (standing tally agrees with the row table)
- `bash -n deploy/idpshow_fetch_and_push.sh`: clean

## Concurrent-work check (before writing code, not after)

`scripts/check_work_claims.py --files scripts/fetch_idpshow.py tests/sources/test_idpshow_acquisition_state.py deploy/idpshow_fetch_and_push.sh --defect V1-136` flagged three other branches touching the same files. Checked each:

- `origin/claude/idpshow-combined-acquisition`, `origin/claude/idpshow-combined-vote` — confirmed their copies of `scripts/fetch_idpshow.py`/`deploy/idpshow_fetch_and_push.sh` are byte-identical (or a one-line docstring difference) to current main. These are the original pre-`#1008`/`#1012` feature branches; their content already shipped under different squash-merge commits. Not active work.
- `origin/claude/v1-52-retire-overview-power-leader` (my own `#996`, unrelated PR) — its copies of both files are also byte-identical to main; it never actually edited either, they were just carried in its large stale-based diff.

## Not claiming

- **Not `VERIFIED`.** This is L1-equivalent code proof (mutation-proved unit tests against a real vocabulary). The reconciliation doc's own plan calls for an L2 board/contract-level measurement plus a real production deployment (an actual VPS timer run stamping a real status file) before promotion — neither is reachable from this environment. Ledger reflects exactly that: `IN PROGRESS`, numerator unchanged.
- No rank→value conversion, no bridge/registry change, no change to which board votes (`idpShowCombined` stays the sole voter, untouched here).
- No credential value anywhere in this branch's history — verified with the same `TestNoSecretValuesLeak` pattern `#1001` used, extended to both boards.
- No change to the repo-wide 0/1/2 exit-code convention `deploy/idpshow_fetch_and_push.sh`'s `if ! ...; then` depends on.

---


jasonleetucker-code:
Traffic control: this is the correct post-#1012 board-aware V1-136 implementation direction, but it is not a promotion packet. Exact-head PR Validation is still running, and the PR itself only has L1-equivalent mutation proof. If/when Integration merges it, reconcile the canonical row to `IMPLEMENTED_UNVERIFIED` (not `VERIFIED`); V1-136 still needs the stated L2 board/contract measurement plus a real deployed timer run stamping `idpShowCombined_last_status.json`. Preserve the anti-false-green requirement: plain-board HEALTHY must never mask combined-voting-board failure. No need for this feature branch to chase later `main` freshness.

jasonleetucker-code:
Acknowledged. Current row is `IN PROGRESS` ("implementation begun, not merged" per §3's own status definitions), which is correct for right now — no change needed while the PR is still open. At merge time I'll swap it to `IMPLEMENTED_UNVERIFIED` (not `VERIFIED`), with the same note already in the row: L2 board/contract measurement + a real deployed timer run stamping `idpShowCombined_last_status.json` still required before promotion. Will not touch this branch to chase later `main` freshness — waiting for exact-head PR Validation to finish, then merging as-is.

---

PR #1043, opened 2026-08-21
V1-41 / C3-CTX-01: Use Team Context toggle, ON by default

jasonleetucker-code:
## Row

`V1-41` / `C3-CTX-01` — "Use Team Context" toggle, target level L1. Contract requirement: *"Use Team Context" toggle exists and defaults ON.*

## Census (required before editing)

No trade surface on `main` consumed any canonical team-context signal at all: `src/trade/finder.py` and `src/trade/suggestions.py` both generated and ranked trades purely off asset value — zero references to `src.roster_intel.strength`/`weakness` in either file. `finder.py`'s existing "light roster-fit adjustment" block was the closest thing to a toggle target, but its "fills a need" arm used a raw position **count** (`<= ROSTER_WEAK_THRESHOLD`), never the canonical Team Weakness owner (`src.roster_intel.weakness.build_team_weakness`, `C2-WEAK-01`). So the honest answer to "does ON use the canonical owner" was **no**, and there was no OFF at all — nothing to gate.

## Fix

Scoped to `src/trade/finder.py::find_trades` — the one surface with a resolvable real `ownerId` per team (`sleeper_teams[].ownerId`).

- Added `use_team_context: bool = True`.
- **ON**: resolves the requesting team's `ownerId`, calls `src.api.roster_intelligence.build_league_roster_intelligence(contract, team_count=...)` — the **same canonical entrypoint** `/api/roster/intelligence` calls, not a re-derivation — to get that team's `TeamWeakness.needs` / `urgentPositions`. The roster-fit block's "fills a need" arm now fires on `need.level in ("critical", "high")` (canonical) instead of the raw count; "sheds surplus" keeps its existing math but is now gated by the same toggle, since it's also team-specific roster fit.
- **OFF** (or an unresolved owner/contract/computation failure — fails closed, never a silent fallback to the retired count rule): skips the whole block. No bonus, no `roster_fit` / `addresses_urgent_need:*` flags, `rankingFactors.rosterFitBonus` stays at its pre-existing `0.0` baseline.
- `metadata.teamContext = {applied, urgentPositions, unavailableReason}` is stamped unconditionally — "off," "unavailable," and "fired" are three distinct, never-inferred states.
- Threaded from `POST /api/trade/finder`'s `useTeamContext` body field; missing/non-bool falls back to the canonical default `True`.

## Evidence

| check | result |
|---|---|
| Required L1 mutation | flipped the `True` default to `False` → `test_omitting_use_team_context_behaves_as_on` goes **RED** (`assert False is True`); reverted → GREEN |
| ON/OFF genuinely distinct | `test_on_and_off_are_genuinely_behaviorally_distinct` — every candidate ON flags as addressing the critical TE need scores strictly higher than the identical OFF-run candidate |
| New tests | `tests/trade/test_finder_team_context.py` — 5 passed |
| Broad regression | `tests/trade/ tests/api/` (`-m "not livedata"`) — **2854 passed, 29 skipped, 0 failed** |
| `ruff format` / `ruff check` | clean |
| decision-coercion gate | clean (660/660, unchanged) |

## Deliberately NOT done

- `src/trade/suggestions.py` is **not** wired — a separate integration, not a mechanical copy of `finder.py`'s `ownerId` resolution shape.
- Analyze Trade (`V1-43`) can't consume this toggle because it doesn't exist yet.
- Competitive Posture / playoff-bye-championship counterfactuals / pick ownership-forecast are **not** wired — no canonical owner exists for these on `main` today, and inventing one would be exactly the "second context methodology" this row forbids. Only positional need (Team Weakness) is wired, because that owner already exists and is already correct.
- No ranking/value math changed — only the pre-existing additive fit-bonus's trigger condition and its now-required gate.

FEATURE_GREEN / READY_FOR_INTEGRATION, not self-merged.

---


jasonleetucker-code:
Traffic control: #1043 is merged and its exact-head L1 evidence is already green, but current `main` still lists V1-41 as `NOT STARTED` (84/136). Subsequent main movement is automated freshness/data churn only. Integration should now perform the shipping-tree V1-41 L1 confirmation and either (a) promote only V1-41 to `VERIFIED` with standing-tally/planning-integrity checks, or (b) record a concrete causal L1 failure. Do not send the feature lane chasing main.

PR #1062, opened 2026-08-22
fix: clear 2027 Late 1st blend-integrity deploy blocker

jasonleetucker-code:
Production-stop repair for the FULL API-contract failure on `2027 Late 1st`.

First commit intentionally adds a real-current-board regression reproducer. It fails with the row's stage/provenance diagnostics so the causal fix can be made against measured canonical pipeline state rather than by weakening the validator or guessing at pick methodology.

Guardrails: no source-weight changes, no bridge changes, no Hill changes, no clamp/tolerance widening, no V1 scope change. The PR will not be merged until the exact live regression is GREEN, the generic impossible-blend positive control remains RED-capable, FULL contract validation is GREEN, and exact-head CI is GREEN.

### DIFF
--- a/scripts/check_env.py
+++ b/scripts/check_env.py
@@ -20,7 +20,9 @@ from __future__ import annotations
 
 import argparse
 import importlib
+import json
 import sys
+from pathlib import Path
 from typing import Iterable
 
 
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Weekend integrity watch: this PR is still not eligible to merge. Exact-head PR Validation remains red because `scripts/check_env.py` crashes in the PR-only diagnostic with `ModuleNotFoundError: No module named 'src'` before the blocking gates run. Meanwhile `main` has advanced from the PR's recorded base `65e43b70` to `ed4dab82` via fresh data/source commits, so current-main compatibility must be re-proven after the import/preflight fix. Do not rerun the same failed job unchanged; fix the preflight import, preserve the real `2027 Late 1st` reproducer, then require exact-head CI + FULL contract green against current main before merge.

jasonleetucker-code:
Superseded by #1063, which reconciled this work onto current main, preserved the live-board regression as a pytest test, removed the mis-scoped check_env diagnostic, passed exact-head PR Validation, and confirmed FULL contract green on the current tracked board. #1063 merged as 441bedb9a73771c5b0aa45a88b0c732c5d82b217. Closing this stale RED branch rather than rerunning or merging it.

jasonleetucker-code:
Superseded by #1063, merged into `main`.

**Why this branch was not merged directly.** The exact-head `Validate PR` failure on this PR was not the blend-integrity finding itself — it was `scripts/check_env.py` crashing with `ModuleNotFoundError: No module named 'src'`. Root cause reproduced directly: `python scripts/check_env.py` sets `sys.path[0]` to the *script's own directory*, not the repo root, so the `from src.api.data_contract import build_api_data_contract` inside the added `_diagnose_live_pick_blend()` could never resolve.

That import bug was not patched in place, because `check_env.py` is a general-purpose preflight invoked by **8** workflows — including `deploy.yml` and `scheduled-refresh.yml`. Embedding a single pick-row blend diagnostic there means any future genuine data edge case on one pick could silently block production deploys and the 2-hourly refresh. The diagnostic was therefore dropped rather than repaired.

**What did land** (#1063): the regression test `tests/api/test_pick_blend_integrity_live_regression.py`, byte-identical to this branch's version, which is correctly scoped and correctly pathed under pytest.

**On the blend-integrity finding itself**, reported honestly rather than papered over: measured against the current tracked board, `2027 Late 1st` does **not** trip blend integrity — `rankDerivedValue: 4958`, contributions `{ktcSfTep: 4961, idpTradeCalc: 4955}`, `blendIntegrityViolation: null`. The test is GREEN because the result is truthful, and it was mutation-proved non-vacuous (forcing the hull check off turns it RED with a full diagnostic, then reverted clean). No causal code defect was found, and none was fabricated to fix: with exactly two anchor-classified value-direct sources, `weighted_count_aware_mean_median_blend`'s `n==2` branch returns the weighted mean, which is bounded inside `[min, max]` by construction — a hull violation is structurally impossible for this row's current population. The historical violation is most consistent with a transient upstream scrape condition that self-corrected on the next 2-hourly cycle; the raw per-source CSVs from that moment are overwritten every cycle and are not recoverable, which is stated as a limit rather than resolved by assertion.

The generic impossible-blend positive control (`tests/api/test_blend_integrity.py`, 26 tests incl. `test_value_above_the_hull_is_detected` / `test_value_below_the_hull_is_detected`) was untouched and still passes — it remains the detector's RED-capable control.

Closing as superseded. No commits from this branch are lost that were not carried forward deliberately.

---

PR #1071, opened 2026-08-24
feat: turn arbitrage into a true public-market inefficiency finder

jasonleetucker-code:
## What this changes

Adds the missing player-level market-arbitrage layer to the existing `/arbitrage` surface.

The old Rankings `Inefficiencies` lens is source-disagreement driven. That remains useful research, but it is not the user's trading question.

This PR adds a direct comparison of the canonical board against the public transaction anchor a trade partner is likely to check:

- offense -> KeepTradeCut SF/TEP (`ktcSfTep`)
- IDP -> IDP Trade Calculator (`idpTradeCalc`)

It then keeps the existing roster/package scan underneath, so the workflow becomes:

1. find individual BUY/SELL edges;
2. identify win-win public-market offer ceilings;
3. turn the edge into an actual roster-to-roster package.

## Truth / architecture boundaries

- Does **not** recompute player values.
- `ourValue` comes from backend-authored canonical board value.
- Public market values come from backend-authored `canonicalSiteValues`.
- Missing public prices remain missing and are excluded, never coerced to zero.
- Source disagreement is not required for arbitrage.
- No source weights, bridges, Hill calibration, clamps, or valuation methodology changed.
- No V1 ledger edit.

## Signals

- >= 5% canonical-over-public: BUY
- >= 15%: STRONG BUY
- <= -5%: SELL
- <= -15%: STRONG SELL
- Win-win offer: counterparty sees at least ~3% public-market overpay while our internal valuation still retains >=10% edge; target public premium is 5%.

## Tests

Adds focused tests for:

- offense routes to KTC;
- IDP routes to IDP Trade Calculator;
- canonical-vs-public edge calculation;
- source disagreement is *not* required;
- missing public price cannot become zero;
- IDP cannot accidentally route through KTC;
- win-win offer ceiling math;
- sorting by actionable percentage edge rather than source-rank spread.

## Integration traffic control

There is an active production stop unrelated to this PR (#1069 plus the deterministic playerctx retention-test deploy blocker). This PR should run exact-head CI but **must not merge ahead of production recovery**.

PR #1077, opened 2026-08-24
Lane 4: guard the flag posture, and correct two stale blocker premises

jasonleetucker-code:
**READY FOR INTEGRATION — Claude 5 promotes/merges. I do not.** `VERSION_1_COMPLETION_CONTRACT.md` is **not** edited and no `VERIFIED` status is changed.

**Head `ef7435518` · `Validate PR` green · `mergeable_state: clean` · 2 commits, 4 files, docs + tests only.**

---

## Update — second commit (`ef7435518`), added after CI

The first commit turned the hard gate red, and the second fixes it. Recording it here because a reviewer would otherwise find an unexplained second commit touching a *different* file than the body describes.

**What broke.** `test_every_api_path_in_the_procedures_doc_is_a_real_route` — my own guard from the already-merged #932 — scans the procedures doc for `/api/…` tokens and fails on any that isn't a registered route. My doc addition cites the suite an operator should run, `tests/api/test_feature_flag_endpoint_reachability.py`, and the regex matched the `/api/` **inside that filesystem path**, inventing the phantom route `/api/test_feature_flag_endpoint_reachability`.

The guard was working; the token was a path, not a route claim. Citing a test file under `tests/api/` is legitimate and will recur, so the fix belongs in the guard — mangling the command to dodge a regex would make the doc worse.

**The trap in the obvious fix.** A negative lookbehind is wrong. `/api/` sits mid-token in a curl target (`https://chaseupside.com/api/sharp/cohort`) exactly as it does in a file path (`tests/api/x.py`), so "must be preceded by a delimiter" cannot tell them apart — it drops both, and a document whose purpose is recording curl commands would then be barely checked. The exclusion therefore keys on the token's **suffix** (`.py`/`.js`/`.jsx`/`.ts`/`.tsx`/`.md`/`.json`).

**A correction worth surfacing: I nearly shipped a vacuous test for that.** My first cut asserted the four curled routes stayed in the set extracted from the real document — and it **passed under the very lookbehind it was meant to reject**. My premise was wrong: each of those routes is *also* named in prose (measured: `/api/sharp/cohort` 2 occurrences / 1 curled, `market` 5/1, `roster-percentage` 3/1, `/api/status` 3/1), so the route survives even when the curl form is dropped entirely. The test asserted a true fact independent of the rule it claimed to test.

Repaired by splitting extraction into a pure `_routes_named_in(text)` so the rule can be exercised against synthetic single-form strings, with a thin file-reading wrapper for the doc guard. Three tests pin it: a route named **only** as a curl target is still scanned; a route named **only** inside a source-file path is not a route; and the two rules don't cancel out (one string carrying both forms must split them — a rule returning everything passes the first, a rule returning nothing passes the second).

**Mutation-proven, three directions, against the real tree:**

| mutation | result |
|---|---|
| bare bogus route `POST /api/sharp/nonexistent-probe` in the doc | **RED** — the guard is not defanged |
| same route in curl form `https://chaseupside.com/api/sharp/nonexistent-probe` | **RED** — the hostname form is genuinely still scanned |
| the rejected lookbehind reinstated in the extractor | **2 RED** — which the first cut did *not* catch |

Doc and extractor restored after each; `git status` clean.

**Two CI reds are not mine and are left alone.** Both reproduce identically on pristine `origin/main` (checked in a throwaway worktree, since removed) and both carry `continue-on-error: true` in `pr-validation.yml`, so neither blocks: the livedata `test_ktc_reconciliation` rank-12 drift (`assert 3.233 <= 3.0`) and `check_source_health` exit 1 (`idpTradeCalc` picks content-stale 40d). The livedata step's own comment says scrape churn "must NOT block every PR (it did, repeatedly)" — so `CLAUDE.md` is accurate and nothing needs escalating.

---

## Why this unit exists

I audited all ten open Lane-4 rows (V1-56/57/58/59/60/61/62/65/89/129) against the live contract. **Every one is externally blocked** on one authenticated `chaseupside.com` session or VPS shell access — verified, not assumed: no credential env var is set, `~/.ssh/` is empty. So there was no legitimate implementation work that advances a row, and manufacturing some would have breached this programme's own rules.

What *was* available: the one V1 audit item that is locally provable, and two blocker premises that measurably no longer hold.

## 1. Feature-flag posture — discharged structurally, then guarded

V1 requires proof that *a flag-off surface cannot masquerade as implemented*. Measured with `src/api/flag_reachability.py` (the caller-graph resolver from V1-88 / #991):

**Zero `is_enabled` call sites exist under `src/sharp/` or in any FAAB module.** There is no flag to switch off — every Sharp and FAAB surface is unconditionally reachable in the intended production configuration. That's a stronger result than a calibration.

`te_basis_conversion` transitively reaches `/api/sharp/roster-percentage`, and that is **correct** — it's Lane 5's canonical-value flag, defaults `True`/`LIVE`, and isn't referenced in `src/sharp/` at all. The board reaches it by consuming board values; forbidding that would forbid the Sharp board from reading the board.

**Guarded, not merely recorded** — three tests appended to the existing file (no parallel file):

| guard | fails when |
|---|---|
| `test_no_sharp_or_faab_surface_is_gated_by_an_unregistered_flag` | Lane-4 code gates on a flag with no `_DEFAULTS`/`_GATE_STATUS` row |
| `test_a_lane4_gate_defaulting_off_would_hide_a_shipped_surface` | a Lane-4 gate defaults `False` — the masquerade case exactly |
| `test_the_lane4_gate_scan_can_actually_see_a_gate` | non-vacuity control |

The control is load-bearing: both assertions currently pass over an **empty** gate set, and an empty set is also what a broken scanner produces. It asserts the scanner still resolves the known real gate in `src/api/gameplan.py`, so "no Lane-4 gates" is a measurement rather than a silence.

**Mutation-proven against the real tree, both directions independently:**
- injecting a gate into `src/sharp/market.py` on `unified_id_mapper` (default `False`) → RED on the defaults-off guard, naming the exact module, function and flag;
- renaming it to an unregistered flag → RED on the registration guard instead.

`src/sharp/market.py` restored **byte-identical** (`git status` clean) and 8/8 green after.

## 2. V1-89 / `OD-04` — the premise no longer holds

`OD-04` asks the owner to re-mint / accept degradation / retire DraftSharks. It was raised on *"~219 h stale against a 24 h threshold, session absent on prod, watchdog red every 2 h"*.

Measured from tracked `data/scrape_state/`:

| key | age |
|---|---|
| `draftSharks` | **0.9 h** |
| `draftSharksIdp` | **0.9 h** |
| `draftSharksRos` | **0.9 h** |

All 28 registered sources ≤ 3.0 h; `main` carries `exports/latest/dynasty_data_2026-08-24.json` (`scrapeTimestamp 2026-08-24T17:26:50Z`). Consistent with `OD-04`'s own recommended option (a) having already happened operationally.

**Bounded honestly, and the bound matters:** `config/source_staleness.json` says in its own header that the stamp tracks *"fetch succeeded"*, not *"vendor published new content"*. This proves the **watchdog condition cleared**; it does **not** prove content freshness, which stays part of the L3 production check. Recommendation recorded for the owner — re-put `OD-04` as *"confirm (a) already happened, then close"* — **not decided here.**

## 3. V1-62's stated blocker points at a projection, not a live defect

Its note reads *"no memoization"* (W15-F017). The shape is real — `cohort_members()` is unmemoized and `market.py` calls it from three sites (`:394`, `:517`, `:593`). But `PERFORMANCE_AUDIT.md` classifies W15-F017 **P3** and says its cost is *"invisible at the current 0 rows… a projection, not a measurement"*, blocked on *"No populated platform ledger"* — the same blocker as V1-58/60/61.

Adding a TTL memo now would be unmeasurable at 0 rows and would risk serving a **stale cohort** (breaching `stale != current`), while not advancing a row whose required level is **L4**. Recorded so a future session doesn't mistake it for available work.

## Also: a stale lane doc that contradicts the contract

`LANE4_V1_RECONCILIATION.md` still opened with *"Contract not found — `VERSION_1_COMPLETION_CONTRACT.md` does not exist on `main`"* and carried three verdicts contradicting the live contract (V1-57 "DONE", V1-60 "DONE", V1-62 "verified honest"). None is wrong about the *code*; all three are wrong about the **evidence level**, because the file predates the contract setting one per row. Header corrected, verdicts re-keyed, contract declared authoritative.

## Verification

| check | result |
|---|---|
| `tests/sharp` + all FAAB suites + all 3 feature-flag suites + `tests/ops` | **green** (664 on commit 1; 555 on the targeted re-run after commit 2) |
| mutation proofs | 5 total, all RED → restored → GREEN |
| `ruff check` / `format --check` | clean |
| `check_decision_coercions.py` | clean, 660 baseline unchanged |
| `check_planning_integrity.py` | all green, incl. *"V1 standing tally agrees with the V1 row table"* |
| `check_release_candidate.py` | 0 class-A/B/C outstanding |

No canonical value surface is touched, so no board diff applies — stated rather than assumed. `check_work_claims.py` flags one overlap on the flag test file: `origin/claude/v1-88-flag-reachability-guard`, my own already-merged branch (#991), i.e. a stale remote ref, not a live editor.

## Scope discipline

Not touched: the V1 contract, any `VERIFIED` status, `src/sharp/**`, `src/trade/faab_*.py` (`faab_engine.py` is additionally claimed by `origin/claude/v1-trade-roster-rehearsal`), Claude 8 bridge/source methodology, arbitrage (#1071/#1076), FAAB Market Heat, Manager Scout, Buy/Sell ecosystem, Consensus Edge. No new coefficient or backtest standard invented. No auth weakened.

---


jasonleetucker-code:
Traffic control — 2026-08-25 11:19 ET: exact head `ef74355181d081e91072915b7885a63ed23257e3` still has PR Validation GREEN (run 32766738550) and there are no unresolved review threads, but GitHub now reports this PR non-mergeable against current `main` (`a1398c7ca639dcad48279846f1a6201178b3d9ee`). Per anti-thrashing: **FREEZE this feature lane; do not chase main movement.** Claude 5 / Integration owns the single planned current-main reconciliation, then exact-head validation on the reconciled head. Merge only if GitHub reports mergeable and the reconciliation reveals no substantive semantic conflict. This docs/tests PR does not by itself promote any V1 row.

PR #1079, opened 2026-08-24
V1-36 / C3-PKG-01: suggestions.py consumes the canonical package identity owner

jasonleetucker-code:
Closes the one live holdout the V1 contract names for `V1-36`: `src/trade/suggestions.py`. Starting SHA `131abf9f`, head `8128ee84`.

`finder.py` / `angle.py` already consume `src/packages`; `roster_intel/packages.py` qualifies as a specialized consumer. Consuming `src.trade.constraints` does **not** satisfy this row — constraints are a different canonical concept.

## Census — two classes, not one

Fixing only the first would have been a *partial* migration, which is worse than none: it leaves two silently-diverging representations inside one file.

**1. Per-asset identity** — 17 sites spelling `.lower()`/`.strip()` by hand: the pool/roster join, the sendable view, the C3-CON-01 blocked set, all four `_generate_*` roster-membership checks, both balancer helpers, the pipeline's roster/exclude sets. → `_identity_key(...)` → `PackageAsset.key`.

**2. Side identity** — three hand-rolled side keys, each spelled differently and each weaker than the owner's:

| retired | weakness |
|---|---|
| `s.receive[0].name` | keyed a side by its **first asset only** |
| `f"{p1.name}\|{p2.name}"` | **order-dependent** — one unordered pair could key two ways |
| `"\|".join(sorted(p.name ...))` | sorted, but over raw display names |

→ `_side_identity(...)` → `side_key(adapt_assets(...))`.

`adapt_assets` reads `.name`/`.position`/`.display_value`, which `PlayerAsset` already carries — no adapter class, no second representation.

## `package_key` has no call site, deliberately

Documented in-code. `package_key(send, receive)` is literally `(side_key(send), side_key(receive))`, and every dedup this file performs is **single-sided**. There is no cross-category whole-package check today, and adding one would suppress suggestions that currently ship — a behaviour change, not an identity canonicalisation.

**Open product question for Claude 5:** should duplicate packages collapse across categories? Not answered here.

## Methodology preserved

Weakest-starter selection, needs-driven targeting, sweetener-tolerance widening, the tightest-gap dedup **rule** (only its bucket key changed), package topologies, fairness, team-context, constraints — all unchanged. No player value, rank, KTC/IDPTC routing, Value Adjustment, source weight, Hill curve, bridge, calibration, lineup or scoring code touched. The arbitrage equal-count policy is **not** imported into ordinary suggestions.

## Proof

**A. Canonical-consumer** — `tests/trade/test_suggestions_package_identity.py`, 11 passed: owner-delegation, output-equality against the owner called directly, order-independence, code-only source scans, plus **behavioural** no-rostered-target-leak (incl. a whitespace/case-noise variant).

**B. Single-owner** — `tests/packages/test_single_owner.py` 2/2 green, unchanged.

**C. Mutation** — each RED for the intended reason, then restored GREEN:
- **M1** hand-rolled `"|".join(...)` recv key restored in `_apply_quality_filters` ⇒ `test_every_side_keying_function_calls_the_owner_routed_helper` RED
- **M2** `s.receive[0].name` restored in `_generate_buy_low` ⇒ 2 RED incl. `test_no_hand_rolled_side_key_survives`
- an earlier per-asset mutation turned 4 of 6 RED **including the behavioural test** — a rostered player leaked back in as a recommended receive target, not merely a source-scan failure

**D. Behaviour — controlled REAL input set.** Real contract built through `build_api_data_contract` from the same pinned, committed, immutable `tests/fixtures/golden/input_export.json.gz` that `golden_board.py` uses (export sha `73fd8b17…`, source-CSV sha `e1ab2a62…`), then the live `/api/trade/suggestions` pipeline for **all 12 real teams × 3 regimes** (production top-150; full 849-row board; asymmetrically-thinned real rosters that create genuine needs): **217 real suggestions, byte-identical before vs after.**

**E. L2 board/contract** — `golden_board.py` + `board_diff.py --expect-no-value-change`: **1111 rows, 0 values moved, 0 ranks changed, 0 newly priced/unpriced**, exit 0.

## Pre-existing defect found, deliberately not fixed here

`_opponent_fit_label` renders `', '.join(give_positions)` over a **set**, so the `opponentFit` display string's position order varies with `PYTHONHASHSEED` across processes — proven by diffing pristine `main` against itself, which differs on the **identical 26 `opponentFit` lines**. A one-line `sorted()` fixes it but changes rendered output, which is outside this row's identity-only scope. All measurements above are seed-pinned accordingly. Reported for a separate bounded unit.

## Gates

`tests/trade/ tests/roster_intel/ tests/packages/ tests/test_trade_suggestions.py tests/api/test_trade_simulator.py tests/api/test_one_canonical_value_per_asset.py` — **1638 passed, 22 skipped (pre-existing), 0 failed**; focused set 188 passed; `ruff check` + `ruff format --check` clean; `check_decision_coercions.py` clean (660/660 unchanged); `check_planning_integrity.py` OK; work-claim check surfaces only the two rehearsal branches carrying no open PR (previously assessed stale).

V1 ledger **not** edited. FEATURE_GREEN / READY_FOR_INTEGRATION — not self-merged.

---


jasonleetucker-code:
CI at `8128ee84` failed three checks. **One was mine and is fixed at `7822a2a4`; two are pre-existing on `main` and are not this PR's to fix.**

### 1. `ruff format --check` — mine, fixed

```
Would reformat: tests/test_trade_suggestions.py
Would reformat: tests/trade/test_suggestions_package_identity.py
```

Root cause: this sandbox carries **ruff 0.15.8** while `requirements-dev.txt` pins **`ruff~=0.6.0`** — a different major with different formatting output. My local gate genuinely reported clean, under the wrong version.

Installed the pin (0.6.9), reproduced CI exactly (*"2 files would be reformatted"*), reformatted those two files only. Whole repo now reports **`1312 files already formatted`** under the pin and `ruff check` passes. Formatting only — no code token changed, 188 focused tests still pass.

### 2. `tests/canonical/test_ktc_reconciliation.py` — pre-existing, not mine

```
Divergence from KTC at rank 12 drifted: pinned +12.3% vs actual +15.5%
(KTC=7410, ours=8561, band ±3.0pp)
```

Checked out pristine `131abf9f` with none of my code present and ran the same test: it fails with the **identical assertion value `3.2330634278002695`**. This is vendor/curve drift on `main`, in the Hill-curve / canonical-value lane — which this dispatch explicitly forbids me from changing. My diff is provably value-neutral: `board_diff.py --expect-no-value-change` reports **0 values moved, 0 ranks changed** across it.

Its message names the decision correctly — *"investigate whether KTC shifted or our curve needs re-fitting"* — and that belongs to the valuation lane.

### 3. `check_source_health.py` — pre-existing, not mine

`idpTradeCalc (picks) raw CSV byte-identical for 40 days (budget 14d)` plus `canonicalSiteValues missing keys` warnings. Same warnings reproduce on pristine `131abf9f`. Source-acquisition lane; nothing in this PR touches fetchers, the registry, or bridges. Per `CLAUDE.md` this lane is intended as **advisory on PRs** — worth a look that it exited non-zero here.

---

Head is now `7822a2a4`. Expect the format gate green; failures 2 and 3 should persist identically until their own lanes address them, and neither is caused by or fixable within V1-36's identity-only scope.

---

jasonleetucker-code:
**`Validate PR` is green at `7822a2a4`.** FEATURE_GREEN — freezing this unit.

One correction to my previous comment, because it made a falsifiable prediction that turned out wrong and Integration shouldn't plan around it: I said failures 2 and 3 "should persist identically until their own lanes address them." **They did not persist — both cleared on this run with no code change from me.**

An automated data refresh landed on `main` between the two runs (18:48 → 18:54), and CI validates the merge of base and head, so the fresh scrape was in the validated tree. That moved the KTC rank-12 divergence back inside the ±3.0pp band and cleared the `idpTradeCalc` content-staleness.

This *strengthens* rather than weakens the diagnosis — both were data-driven, not code, which is exactly why they cleared on new data while my diff stayed byte-identical. But the practical consequence is the opposite of what I wrote: **there are no outstanding lane blockers from this run.** Worth noting they are liable to recur on the next drift, since neither has a durable fix; that remains the valuation and source-acquisition lanes' call, not V1-36's.

The only genuine code defect in the original run was the format gate, and that was mine — fixed under the repo's pinned `ruff~=0.6.0`.

---

jasonleetucker-code:
**Not part of this diff — a V1-45 finding surfaced while auditing the trade lane after freezing V1-36. Recording it here because Claude 5 will read this PR at integration and it would otherwise be lost.**

### V1-45 (`inv 2.1`, Trade calculator, `IMPLEMENTED_UNVERIFIED`, target **L4**)

L4's bar is *"the intended surface consumes the canonical implementation truthfully."* Measured on `131abf9f`:

| block | backend emits | frontend renders |
|---|---|---|
| `rosterCapacity` | yes | **yes** — `SimulationPanel`, `trade-sections.jsx:322` |
| `finalRosterSimulation` | yes — `trade_simulator.py:440/454/461` | **no — zero files, under any alias** |

The plumbing is live: `useTradeSimulator.js` POSTs `/api/trade/simulate` and `SimulationPanel` receives `simResult`. So `finalRosterSimulation` **arrives in the payload and is read by nothing.**

That block is V1-42's canonical output (`C2-SIM-01`, VERIFIED): `strengthBefore`/`strengthAfter`, `coreBefore`/`coreAfter`, `weaknessBefore`/`weaknessAfter`, `promotions`, `displacements`, `movements`, plus `cleanupApplied` / `cleanupIsUpperBound`. It answers *which player actually gets cut and what the re-solved lineup is* — where `rosterCapacity` alone only gives the forced-drop **cost estimate**.

This is the identical defect class to V1-45's earlier pass (backend computed `rosterCapacity`, zero files rendered it) one layer deeper. It is **not** a re-opening of V1-42, whose backend wiring is VERIFIED and mutation-proven — L4 measures the *consumer*, which is exactly what is missing.

Also unrendered: `unpricedForcedDrops` (the missing-is-never-zero companion to `forcedDropValue`, which *is* rendered).

**I did not fix it: this dispatch excludes `frontend` from my ownership.** Handing it over rather than crossing that line.

**Exact remaining V1-45 evidence recipe:**
1. Render the three `finalRosterSimulation` states `trade_simulator.py` already distinguishes — computed, `{"available": false, "unavailableReason": "capacity_uncertain"}`, and `{"unavailable": "<ExcType>"}` — display-only, no trade math in the component (same posture as the `rosterCapacity` render).
2. RED-before/GREEN-after by running the new test against the pre-fix component.
3. L3/L4: deployed-SHA checklist + a real authenticated `/api/trade/simulate` round-trip showing the block rendered. **Not reachable from this environment** (no production/auth access) — this half was already recorded as outstanding when V1-45's render pass shipped in #1025.

---

PR #1081, opened 2026-08-24
V1-52: bind recent form and all-play to the last SCORED season

jasonleetucker-code:
## Summary

The dispatch asked me to confirm — not assume — that V1-52's remaining gap is **purely state-blocked**, with no independent implementation defect hiding behind the missing observation.

**The confirmation failed. There is one, and it is live on production right now.**

`src/ros/power_v2.py::build_section` gates the recent-form buffer on `season is seasons_sorted[-1]`, but the accumulation loop `continue`s past a scoreless season *above* that line. `seasons_sorted[-1]` is the newest season **in the snapshot**, not the newest season **with scores** — and those diverge in exactly one situation, which is the one production is in: a scoreless current season sitting after prior scored seasons, i.e. every preseason.

So the guard never fires for *any* season, `last_season_recent` stays empty for every owner, `_score_state` reads `recent = 0.0` for all, and a percentile over an all-equal list is **0.5 for everyone**.

Same family as this session's merged #1032 (`ppg`/`wl_record`) and #1059 (`streak`/`luck_regression`) — on the one accumulator neither touched.

## L2 — measured on the live board, not asserted

`GET https://chaseupside.com/api/public/league/rosPower?lens=results_only`, real 12-owner league:

- **12 of 12** headline rows carry `recent = 0.5` and `recentAvg = 0.0`
- **108 of 108** trend rows (9 weeks × 12 owners) carry the same
- Active weights sum to `0.5500`, so `recent` is **21.82%** of every published score
- Every `powerScore` contains **10.91 points of unmeasured constant**; the UI renders `recentAvg` as a literal `"0.0"` as though observed

Rank *order* is undisturbed (a shared constant shifts all owners equally) — which is why this survived three prior reviews of this file.

## It was documented as verified

`docs/power/V1_52_PPG_SEASON_SCOPING.md` (#1032's own record) cites the defective gate as *proof of correctness*: "`recentAvg` and `all_play` were already correctly season-scoped (gated by `if season is seasons_sorted[-1]:`)" and "remain correct". Both passages are corrected append-only, originals left standing — a documented-as-verified falsehood is itself part of the record.

## Fix

All six accumulators now reset at the same point, on each **scored** season, so "the last scored season wins" is structural rather than a property of which season sits last in the list. `last_season_allplay_share` is included deliberately: its unconditional overwrite left a stale prior-season share for an owner absent from the last scored season while their `season_state` correctly went empty.

## Test plan

- [x] New `_preseason_shape_snapshot()` — **the production shape no existing fixture had** (scoreless current season + prior scored season). Four scored weeks against `_RECENT_WINDOW = 3` so a non-sliding buffer is also caught. This is why no test caught it: every other two-season fixture gives the current season real scores, so the gate fires.
- [x] 5 new tests incl. an explicit non-vacuity check and a guard against the opposite failure (a "fix" that silently dropped the component)
- [x] Mutation (restore the retired binding at both sites): **RED** with the predicted signature — `assert {0.5} != {0.5}`, "every owner sharing the 0.5 midpoint is the signature of an unmeasured component"; restored **GREEN** 16/16
- [x] `tests/ros/` + `tests/public_league/`: **671 passed, 2 skipped, 0 failed**
- [x] `tests/api/test_canonical_ownership_protections.py`: 24/24
- [x] `ruff format --check` / `ruff check`, `check_planning_integrity.py`, `check_decision_coercions.py`: clean

Honest caveat: `test_all_play_also_survives_the_scoreless_season` **passes under the mutation** — `all_play`'s unconditional overwrite already worked in this shape, so it is a regression guard for the reset added here, not part of the RED proof.

## For Integration — branch overlap, reported not resolved

`scripts/check_work_claims.py` flags `origin/claude/v1-52-power-lenses` as also touching `src/ros/power_v2.py`, and a trial merge **does conflict**. Checked before proceeding:

- it has **no open PR** (verified against all 28 open PRs);
- its named work is already on `main` via #996/#1009;
- **it does not fix this defect** — `if season is seasons_sorted[-1]:` is unchanged on both sides of its diff.

Treating it as stale. Flagging rather than resolving, since branch reconciliation is Integration's call.

## Non-goals

Does not touch `career_state` (career-scoped on purpose; backs `_enumerate_owner_ids`'s historical-presence fallback). No weight, lens, refusal contract, or methodology changed. **Does not promote V1-52** — Integration owns §3.

Two live-measured findings surfaced for owner decision, deliberately not acted on: the results-only lens is **not** refusing in preseason (12 real ranks, so Section 3's stated premise needs revisiting), and two owners with 0 games receive real ranks 11-12 assembled entirely from component defaults.

Full record: `docs/power/V1_52_RECENT_SEASON_BINDING.md`.

---


jasonleetucker-code:
**Integration review — CHANGES_REQUIRED.** The defect is real, the repair is correct for the case it names, and the PR **understates its own blast radius**. But it does not close the truth rule it is filed under.

## Your defect reproduces, and it is bigger than claimed

Measured against deployed production today, not taken from the body — `GET /api/public/league/rosPower?lens=results_only`:

- active weight sum **0.5500** exactly; recent share **21.8182%**; constant `0.5 × 0.12 / 0.55 × 100` = **10.9091** pts
- headline: distinct `recent` across all 12 rows = `[0.5]`, distinct `recentAvg` = `[0.0]` — **12/12 confirmed**
- trend: **312/312** rows (26 weeks × 12 owners), not the **108/108** the PR claims — you understate it ~3×

Independent RED-before: your new test file copied onto the base tree fails **2 of 16** with exactly the predicted `assert {0.5} != {0.5}`. On HEAD, 16/16. The binding is structurally to the last **scored** season (resets moved below the `continue`), not the calendar season. Good.

## Blocking — `unknown != zero` is not closed

`_score_state`'s `recent = sum(rb) / len(rb) if rb else 0.0` is **untouched by this diff**. Two measured cases:

**(a) No scored weeks anywhere.** HEAD output is byte-identical to BASE — `recent 0.5`, `recentAvg 0.0`, `powerScore 33.64`, no `unrankable`, ranks handed out on a tie. The unmeasured constant survives completely.

**(b) Owner present in the league but absent from the last scored season.** HEAD publishes a literal `recentAvg = 0.0` and a recent percentile at the **bottom of a now-measured distribution**. That is a *stronger* false claim than the 0.5 it replaces: missing now reads as "measured, and last" rather than "neutral". This is live — `jstuedle` and `Blaine` carry record `0-0`, and their recent percentile goes `0.5 → 0.0833`, losing the 10.91 constant with nothing measured put in its place.

The results-only lens has no route to declare `recent` missing (`_HISTORICAL_RESULTS_COMPONENTS` is dropped only under `if preseason and not results_only`). **The display layer is already ready** — `ros-power.jsx::fmtRaw` renders `null` as `—` and `0.0` as `0.0`. The backend coercion is the sole blocker.

## Blocking — three unpinned resets

The PR calls its three added reset statements "structural for all six accumulators", and all three are unpinned. A third-season fixture — which the PR itself names as missing and declines to add — closes two of them.

## Measured effect of merging (so it is on the record)

12/12 headline `powerScore`s change, max |delta| **10.00**, **zero rank changes**. Brent 81.46→91.46, Joey 74.75→82.94, jstuedle/Blaine 20.00→10.91. 312/312 trend rows change. **Canonical surfaces: zero change** — no importer of `power_v2` under `src/api/`; `rankDerivedValue`, `/api/data`, board and contract untouched. One engine confirmed: `src/public_league/power.py` was already deleted in `f92a465ed`; this PR adds no new path.

## Doc defect introduced here

`docs/power/V1_52_PPG_SEASON_SCOPING.md` says "`all_play` was genuinely correct and stays so", contradicting your own `V1_52_RECENT_SEASON_BINDING.md` — and `all_play` was measured going `1.0` (stale 2024) → `0.0` for an absent owner. It corrects one documented falsehood while writing a smaller one.

## Routed separately, not caused by you

The default/forward-looking `rosPower` payload serves placeholder display names (`"Team 1"…`) for the **same ownerIds** the results-only lens names correctly; one fetch also returned 10 of 12 rows before self-correcting. Stale assembled-contract cache is the leading candidate — cause **not** asserted. This blocks recipe Section 2 on the endpoint the recipe actually curls.

## To unblock

1. Decide the missing state (case a/b). If that is an owner call, say so and I will route it rather than have you guess.
2. Pin the three resets — the third-season fixture you already identified.
3. Fix the `all_play` contradiction.

**Note on CI:** `main` is currently red for an unrelated upstream reason (`partial_run_critical:IDPTradeCalc`, plus a `blend_integrity_violation` on `2028 Late 3rd`), so exact-head CI is not evidence about your code until the refresh clears. Merging this does **not** promote V1-52 regardless.

---

jasonleetucker-code:
Traffic control: this PR has a real repair, but the existing Integration review still identifies branch-specific blockers and the owner invariant already resolves the main semantic question: **missing/unknown != zero**. Therefore do not leave the missing-state choice open. For both (a) no scored weeks anywhere and (b) an owner absent from the last scored season, recent-form data must remain unavailable/unknown (e.g. `null`/unmeasured semantics), never coerced to `0.0`, bottom percentile, or neutral `0.5` merely to keep a numeric score. Pin all three season-reset statements with the third-season fixture already called out, correct the `all_play` documentation contradiction, and preserve the existing frontend `null -> —` behavior. Do not touch unrelated placeholder-name/cache behavior in this PR. Mutation-prove missing→zero/neutral becomes RED, restore GREEN, then FREEZE for Claude 5 reconciliation. Current `main` deploy is green, so the old IDPTradeCalc/blend-integrity CI excuse is no longer current.

jasonleetucker-code:
Addressed — pushed `fac997be1`.

## Missing/unknown != zero

You were right that the first commit left the semantic open. It bound recent form to the last **scored** season — fixing *which* season it describes — but left the unmeasured case as `recent = ... if rb else 0.0`. Every owner shared that `0.0`, so `_percentile` returned a confident **0.5** for all of them: the original defect wearing a different mask.

Both cases now stay unknown, through two mechanisms, because the missingness has two granularities:

| case | mechanism |
|---|---|
| **(a)** no scored weeks anywhere | `"recent"` joins `missing_inputs`; the weight renormalises away league-wide — the **same** path `team_ros_strength` and `schedule_adjusted` already use |
| **(b)** owner absent from the last scored season | component stays in the section budget; that **row's** `weightsApplied` drops it and its score renormalises over the weights it actually has |

Neither invents a rule — (b) is (a)'s existing renormalisation applied where the missingness actually lives. Both alternatives were ruled out explicitly: multiplying the weight by a stand-in is the coercion being removed, and leaving the weight in the divisor against a zero numerator deflates the score by exactly that weight.

An owner with **nothing** measurable refuses outright (`powerScore: null`, `rank: null`) and is partitioned out of the sort — **not** given a placeholder sort key, because substituting a number even only for ordering is the same coercion. Your own `check_decision_coercions.py` caught my first attempt at that (`-(score or 0.0)`); it was right to, and it's gone.

## Two things that needed no change, verified rather than assumed

- **`_percentile`** already excludes `None` from its comparison population, so an unmeasured owner can't drag the scale others are ranked against.
- **Frontend `null -> —` is preserved untouched.** `fmtRaw` already renders null as an em-dash, and `ComponentBar` returns `null` for a weight it can't find — so an unmeasured component's bar **vanishes** instead of rendering a fabricated 0%. Publishing per-row `weightsApplied` is what makes that work, and the field name already said "Applied". Frontend power tests: **18 passed**, no frontend file changed.

## Mutation proof

Three mutations, one per coercion you named, all RED then restored GREEN:

```
A  recent = ... if rb else 0.0           (zero)     -> 4 failed
B  "recent": 0.5 if i["recent"] is None  (neutral)  -> 3 failed
C  owner_weights = active_weights        (keep wt)  -> 5 failed
```

C fails loudest because a `None` reaching the weighted sum is a `TypeError`, not a wrong number — the coercion can't be reintroduced quietly.

## Three season resets pinned

Added the three-season fixture the previous commit called for (`TestEverySeasonResetHoldsAcrossThreeSeasons`). A third season is what proves the reset happens on **every** scored season rather than once at the end — an accumulator reset only on the final iteration passed every earlier test in the file. 2024 and 2025 invert completely, with a non-vacuity check asserting they do.

## `all_play` contradiction corrected

It was claiming both "this unit binds `all_play`" and "not claimed as reproduced". Split into the two claims it was conflating:

- its **RESET** is fixed here and now pinned by the three-season fixture;
- its **MISSING-value** semantics are **not** — an absent owner still resolves through `.get(oid, 0.0)`, the same coercion family. Recorded as a named residual rather than quietly fixed, since your direction scopes the value-semantic change to recent-form.

## Scope

Untouched: placeholder-name and cache behavior, `career_state`, weights, lenses, methodology. `tests/ros/` + `tests/public_league/` **684 passed, 1 skipped**; ruff, planning-integrity and decision-coercion clean.

Frozen again for your reconciliation.

---

PR #1084, opened 2026-08-24
V1-49: coverage verdicts state which path they describe

jasonleetucker-code:
## Summary

Second in-sandbox prerequisite found while preparing V1-49's production proof.

`scoring_coverage.engine_reads_key` probes with `_MAXIMAL_ROW` — whose keys are nflverse **column** names — and calls `compute_weekly_points` with **no `source=`**. So every verdict this module returns describes the champion (nflverse) path only, and that was entirely silent.

It matters because `host_native_scoring` switches the engine onto `host_stat_line` and the host's own vocabulary, where a rule reachable on one path is not automatically reachable on the other.

## The measurement

Measured over all three committed 2025 REG host dumps (6,181 player entries, numeric-id filtered to PLAYER entries per `is_host_player_entry` — Sleeper keys team defenses by alpha code and several ST keys appear on both):

| key | present on host player entries |
|---|---|
| `punt_ret_td` | **0** |
| `pt_return_tds` (the nflverse column it's scored from) | **0** |
| `st_td` (combined) | present, 3 / 0 / 3 nonzero |
| `kr_yd`, `pr_yd` | present and nonzero |

So this is a real **per-key asymmetry**, not the host omitting special teams generally — and `classify("punt_ret_td")` returns `SCORED` unconditionally.

Under the flag, a league paying that rule scores it at **0** with `unscored` empty, because that signal is populated only on the nflverse branch. **A production run cannot tell that apart from a correct zero** — which is why this belongs before the prod step.

## Fix

- `HOST_PATH_UNREACHABLE` — a **record**, not a probe, asserted against the committed dumps by test so it cannot quietly go stale the way a comment would.
- `classify`'s docstring now states its nflverse scoping explicitly.
- `kick_ret_td` deliberately **not** listed: it is already `UNSCORABLE` without the supplement and its `UNSCORABLE_REASONS` entry already records the host gap. Listing it twice would be a second owner for one fact — pinned by its own test.

## Test plan

- [x] 7 new tests, including **two non-vacuity controls**: the dumps are present and populated (>6000 entries), and the host really does publish other ST keys — otherwise the absence claims would pass trivially
- [x] **Mutation A** — add `kr_yd` (a key the host *does* publish) to the record: **RED**
- [x] **Mutation B** — empty the record entirely: **RED**
- [x] Both directions matter: the constant can be neither padded with a false claim nor silently emptied. Restored → **GREEN**, 23 passed
- [x] `tests/nfl_data/` + `tests/league_comparison/`: **669 passed, 1 skipped, 0 failed**
- [x] `ruff`, `check_planning_integrity.py`, `check_decision_coercions.py`, `check_work_claims.py`: clean

## Deliberately not attempted — named, not silently skipped

The **runtime symmetry fix** — making the sleeper branch report `unscored` for structurally-absent rules the way the nflverse branch already does — is not here. It needs an authoritative host key **vocabulary**, and three sampled weeks is not that. Treating them as complete would be exactly the overstatement `_MAXIMAL_ROW`'s own correction comment exists to prevent ("the probe asked the engine a question no production row asks it").

That distinction — a key absent because the player didn't accrue it, versus a key the host *never* publishes — is what the runtime fix turns on, and it cannot be settled from this sample. Handed to Claude 5 as a scoped finding.

## Non-goals

Does not touch the `host_native_scoring` flag (still default OFF; its four open gate items are unchanged), any predicate or scoring semantic, or V1-49's ledger status.

---


jasonleetucker-code:
CI red on **Validate PR**, but **both failures reproduce on the base branch** and neither is in a module this PR touches.

```
FAILED tests/canonical/test_ktc_reconciliation.py::TestKTCReconciliation::test_rank_delta_within_tolerance[12-8561-12.3-3.0]
  Divergence from KTC at rank 12 drifted: pinned +12.3% vs actual +15.4%
  (KTC=7420, ours=8561, band ±3.0pp)

FAILED tests/roster_intel/test_metric_separation.py::test_metrics_genuinely_diverge_on_a_representative_real_board
  the single strongest-by-value team should not also rank identically on youth
  by coincidence: (1, 1)
```

**Verified rather than asserted.** Checked out pristine `origin/main` (`a3c391135`) with none of this branch's changes, and ran both:

```
2 failed, 9 passed
```

Same two tests, same assertions. Both are functions of the **live board**, which refreshes on a ~20-minute automated cadence — the KTC one compares our curve against KTC's current published values, the metric-separation one asserts a property of the current real roster board. This branch adds a dict constant and a docstring to `src/nfl_data/scoring_coverage.py`; it cannot move a KTC rank-12 value or a roster-intel team ranking.

**This branch's own state on the actual merged head (`417de6785`):**

- `tests/nfl_data/test_scoring_coverage.py` — 23 passed
- `tests/nfl_data/` + `tests/league_comparison/` — 675 passed, 1 skipped, 0 failed

So the diff is green; the base is red. Per the release-discipline note in `CLAUDE.md`, these are the shape of test that "must not assert an absolute count, a floor or a health status over the LIVE board" from the hard gate — they'll recover on their own when the next refresh moves the numbers back inside tolerance, or they need re-pinning as their own unit. Either way it isn't this PR's, and I'm not re-pinning a KTC tolerance from here — that's a calibration decision, not a merge fixup.

Also visible in the same run, and separately advisory per `CLAUDE.md`: `check_source_health.py` exited 1 on `idpTradeCalc (picks) raw CSV byte-identical for 40 days (budget 14d)` — the fetch is fresh, the vendor has published nothing new. That is source health, not code correctness, and it is the lane that is meant to be advisory on PRs.

Expect the same two failures on #1081 and #1085 once their bases are merged forward — same cause, same non-relationship to their diffs.

---

PR #1086, opened 2026-08-24
Lane 6 — V1-131 implemented; V1-109 partial repair; V1-101/102/105 evidence

jasonleetucker-code:
Lane 6 (Premium UI / Frontend), from a live census at the dispatch SHA `131abf9f9`, not from the prior lane prompt.

**No V1 status column edited. Do not merge without Integration.**

---

## Status at head `1d4cc93b3` — two states, reported separately

| | |
|---|---|
| **IMPLEMENTATION HEAD** | **GREEN** — every gate this PR's diff can affect passes at the exact head |
| **RELEASE CANDIDATE** | **BLOCKED** — both surviving CI failures reproduce on clean `origin/main` with none of this PR's code present |

### Blocker 1 — `test_metrics_genuinely_diverge_on_a_representative_real_board` (PR Validation hard gate)

The test's only external input is the archive bundle `newest_complete_raw_payload()` selects. This branch's merge-base is the dispatch SHA `131abf9f9`; `main` is now `332d8e6ff` and carries three bundles this tree does not — and CI validates the **merge** ref, so CI selects a board that is not in this working tree. Crossing both variables in one environment, back to back:

| code | archive | result |
|---|---|---|
| clean `origin/main` `332d8e6ff` worktree, **no #1086 code** | `…230400` (main's newest) | **FAIL** `(1, 1)` |
| clean `origin/main` `332d8e6ff` worktree | `…172650` | **PASS** |
| branch `1d4cc93b3` | `…172650` (this tree's newest) | **PASS** |
| branch `1d4cc93b3` + main's newest archive copied in | `…230400` | **FAIL** `(1, 1)` |

**The code column has no effect.** Classification **B — pre-existing on `main`, data-coupled.** Consistent with the diff touching no `src/roster_intel/`, no `src/ros/`, no `data_contract`, no valuation or pick-pricing file.

**The two axes have NOT collapsed** — 10 of 12 teams diverge, and the test's *primary* assertion (`inversions >= 6`) passes comfortably:

```
strRk  1  2   3  4  5  6  7  8  9 10 11 12
yciRk  1 11   2  4  7  5  3  9  8 12  6 10
```

What fails is a *secondary* clause asserting the #1-by-strength team is not also #1 on youth — a coincidence occurring ~1 board in 12 under perfect independence, so it reddens ~8% of boards *precisely when the axes are behaving correctly*. That is a live-board-coupled assertion in the blocking gate, the shape `CLAUDE.md` names directly.

Recorded alongside it: `tests/archive_fixtures.py` removes **scrape-health** nondeterminism (its stated purpose) but not **board-content** nondeterminism — it is deterministic for a fixed tree, and the 2-hourly refresh means the tree is not fixed. Fine for a population property; not for a fact about one distinguished team.

The assertion belongs to **V1-35 / the Roster lane** and is that row's own EVIDENCE-L1 artifact, so it is handed over with a drafted repair rather than edited here. **Nothing was skipped, `xfail`-ed, deselected, reclassified `livedata` or relaxed.**

### Blocker 2 — `journey-trade.spec.js` `/arbitrage` "Pick a team and scan" (E2E Safety Net)

Reproduced on clean base `978bf5178` from the same backend in the same session, with no #1086 code present. **Classification B — pre-existing.** Assertion untouched.

### The 429 privacy-boundary red is FIXED

`1 failed, 1563 passed` (`ed4c0f64c`) → `1 failed, 7517 passed` (`b2b724efd`). Cause was rate-limit budget exhaustion, not an auth defect: `/api/auth/status` is in `_PUBLIC_API_EXACT`, so it spends from a 60/min per-IP budget shared by the whole suite, and this PR's two new modules added ~60 such calls earlier in suite order. Repaired at the source of the pressure with the limiter's own sanctioned `reset_for_tests()` hook.

**The privacy invariant holds and 429 was not allowlisted.** With a fresh limiter the boundary answers **401** on both base and branch; the 429 body carries only `error` / `message` / `retryAfterSeconds` — no league payload, no section data. The resource does not become observable because the response is 429; it becomes *less* reachable. No expected-status list was widened.

Full measurement for all three: `docs/master-site-audit/evidence/V1-131/E2E_CAUSALITY_2026-08-24.md`.

### Exact-head gates

| gate | result |
|---|---|
| CI hard gate | **7,517 passed** (everything alphabetically before the base-caused failure, incl. both new `tests/api/` modules) |
| local complement — the ~2,400 tests CI's `-x` never reached (`roster_intel scoring scripts sharp snapshots sources trade utils valuation_math`) | **2,370 passed / 0 failed** |
| nav-gated features + guest-pass evidence + `tests/consensus_edge` | **343 passed / 18 skipped** |
| frontend Vitest | **2,363 / 2,363** across 160 files |
| production build + bundle budgets | green, **all 14 under** |
| `ruff check .`, planning integrity, decision coercions | clean |
| CI E2E Safety Net | **195 passed / 62 skipped**, incl. all three specs this PR adds |

Note: `Frontend unit tests (vitest)` and `Frontend build + bundle-size budget` carry no `if: always()` in `pr-validation.yml`, so the base hard-gate failure ends the CI job before this lane's own gates run. They are reported above as locally verified at the exact head and CI will exercise them as soon as the base blocker clears.

---

## Full unverified L6 census

L6 owns 16 of the contract's 136 rows; 9 are already `VERIFIED` (including `V1-106` windowing and `V1-107` mobile payload — untouched, per the dispatch). The 7 open rows:

| row | on `main` | lvl | class | outcome this pass |
|---|---|---|---|---|
| V1-131 | `NOT STARTED` | L3 | C | **implemented**, mutation-proved |
| V1-109 | `IN PROGRESS` | L4 | B | 1 of 4 defects **fixed + verified**; 1 no longer reproduces; 2 measured, deliberately unrepaired |
| V1-101 | `IMPLEMENTED_UNVERIFIED` | L4 | A+D | fix verified intact; production-build regression added; **L4 recipe written** |
| V1-102 | `NOT STARTED` | L4 | A+D | **stale status — already implemented on `main`**; corrected row text drafted |
| V1-105 | `NOT STARTED` | L2 | A+D | current state measured both viewports; owner decision left open |
| V1-110 | `IN PROGRESS` | L2 | D | **blocked** on owner decision `OD-05`; no code |
| V1-111 | `IMPLEMENTED_UNVERIFIED` | L4 | D | **blocked** on a deployed consumer; no code |

## Two findings that changed the sprint

**1. `V1-102` is not `NOT STARTED`.** The guest-pass system *is* the temporary-password generator with configurable expiry — the panel's own copy reads *"Generate a temporary password to share with someone you want to give private-app access."* `src/api/guest_passes.py` meets every stated requirement (`secrets.token_urlsafe(24)`, SHA-256-only persistence with the plaintext shown once, 1 min–30 day configurable expiry / UI 1–720 h, create-list-revoke, server-side expiry enforcement plus a cookie max-age cap, no second session subsystem). Building a second one would have created exactly the duplicate auth subsystem the row forbids, so **nothing was implemented, deliberately.** Evidence closed to L2 with 16 added guards covering only what the 16 existing tests did not — notably fail-closed for expired **and** revoked tokens *at the login boundary*, which a layered mutation showed is a second line of defence the unit tests never reached. Corrected row text is drafted in `docs/master-site-audit/evidence/V1-102/HANDOFF_TO_CLAUDE_5.md`.

**2. `V1-131`'s premise was partly wrong, and correcting it improved the fix.** The row says Consensus Edge's "three endpoints 503". Measured: the board endpoints do; **`/methodology` deliberately answers 200 with `enabled: false`**. The requirement stands (the page is unusable either way) — and that endpoint is why an honest capability signal exists at all, instead of a static nav deletion that would make the feature unreachable for the operator who turns the flag on to evaluate it. Also recorded: those routes gate on auth *before* the flag, so anonymously they 401 — the 503 dead end is what a **signed-in** user hits, which is exactly who the nav offers it to.

## V1-131 — the implementation

`consensus_edge` defaults OFF (ADR-023), so the default production state was a menu entry to a dead page. Gating only; the feature stays post-V1 and no flag default changed.

- **Rides the shell's existing probe.** `useAuth` already calls `/api/auth/status` on every route and already feeds a nav filter, so the capability is returned in a `features` block there rather than from a second request — `V1-108` is `VERIFIED` and a new per-page fetch would regress it.
- **`available`, not `enabled`.** The capability derives from `src/consensus_edge/api.py::is_available()` — the one canonical predicate, `flag_enabled AND contract_loaded` — because the board handlers 503 on **two** independent conditions (`feature_disabled`, `data_not_ready`). A flag-keyed nav would advertise a dead page for the whole window where the flag is on and the data is not.
- **Fails closed twice.** `_nav_gated_features` cannot raise (a `KeyError` escaping `/api/auth/status` would take down the probe the whole shell depends on), and `itemIsOffered` offers only on an explicit `true` — missing block, `null`, `{}`, a bare boolean and a truthy non-boolean are all "unknown", and unknown is not offered.
- **All five offer surfaces**, because one is a half-fix: hiding it from the drawer while ⌘K still routes there leaves the dead end. `pageTitleFor` deliberately stays unfiltered — a page you can still reach by URL must be able to name itself.

Verified live on the production build: **0** links to `/consensus-edge`, `features: {"consensusEdge":{"available":false}}`, board `503`, **0** shell requests to `/api/consensus-edge/*`.

## V1-109 — measured before repairing

Both audit claims predate #984 and #1003, so they were re-measured at 390×844 against a production build and the real board. New instrument refuses to run without passing a control (known 30/44/48 px targets, a 9 px span) and records a non-rendering route as `unavailable` rather than as zero violations.

- **FIXED** — `/trade`'s sticky tray was covered by the screenshot FAB. Reproduced exactly: tray at y 714 h 74 (z 35), FAB at y 734 44×44 (z 900) — the FAB sat *inside* the tray. **44×44 overlap → 0.**
- **NO LONGER REPRODUCES** — `/draft` stacking: 0 px document overflow; remaining wide elements are inside their own `overflow-x` scrollers, which is this codebase's required pattern.
- **STILL REPRODUCES, reported not repaired** — touch targets 163/170 under 44 px on `/rankings` (the count fell from 873/900 but the **rate did not**: 97% → 95.9%, which is windowing mounting fewer rows, not a usability gain), and 9.0 px type on exactly three pages. The correct touch-target repair is hit-area expansion, not a bigger box (row height is load-bearing for `V1-106`) — and hit-area expansion is invisible to a `getBoundingClientRect` instrument, so shipping it would mean claiming a repair this pass cannot verify. Type scale is a token decision gated on `OD-05`.

## V1-105 — measurement, not a decision

Existing instrument, unmodified, 5 runs/cell, both viewports. `/trade`, `/league`, `/` are legitimate **prospective** baselines; `/rankings` is **current state only** and is not labelled otherwise. **One target miss reported as a miss: desktop `/rankings` warm p95 useful 1018 ms > 1000 ms.** #1006 is still open, so the scope call stays `OWNER_DECISION_REQUIRED`.

## Four defects found in the tooling and in this PR's own tests

- **A negative assertion of mine could not fail.** Re-running the V1-131 mutation at the exact head reported `Test Files 1 failed | 1 passed`: the pure-model suite went red while the *rendered-chrome* suite stayed **green with fail-closed broken**. `NavMenu` renders items only while `open`, so two "the link is absent" tests were passing because the menu was shut, not because the capability said so. Repaired by opening the group and adding a non-vacuity guard (`/edge` must be present); the identical mutation now yields **7 RED across 2 failed files**. A strengthening — nothing removed or relaxed.
- **The baseline harness can report a pass having measured almost nothing.** A rate-limited first run left 7 of 8 cells `—` and it still printed *"All measured routes inside the standard's targets."* That run was discarded, not reported. Recommending a non-zero exit on unmeasured cells to that script's owner rather than editing it here.
- **`page.route()` does not intercept service-worker fetches** — once `sw.js` installed, an E2E stub was bypassed and the panel got the real 404, presenting as an intermittent failure of an assertion about something else.
- **`OWNER_FEATURE_INVENTORY.md:245`** still says *"no windowing implementation exists"* (false since `V1-106`) and still cites **59.5 FPS**, which the FPS harness's own header retracts as a self-measurement artifact; `C8-PERF-03` carries the same figure. **Reported, not edited** — canonical records belong to Integration.

## Mutation proof — all re-measured at this head

| unit | mutation | result |
|---|---|---|
| V1-131 backend | `is_available()` → `return True` (claim available regardless of canonical state) | **8 of 23 RED**, incl. every `capability_and_the_board_never_disagree` case where the board is unavailable |
| V1-131 model | `itemIsOffered`: `=== true` → `!== false` (break fail-closed) | **6 of 24 RED** — exactly the unknown / absent / bare-boolean states |
| V1-131 chain | same mutation, after the vacuity repair above | **7 RED across 2 failed files** (was 6 RED / 1 file passing vacuously) |
| V1-109 | remove the FAB clearance rules, rebuild | **RED** — `trade-sticky-tray × screenshot-fab — 44×44px` |
| V1-101 | remove the module import | build **PASSES**, E2E **RED**, `/admin` renders *"Something went wrong"* |
| V1-102 | expiry check removed from `validate()` only | **only the UNIT test** reddens — the login boundary carries its own independent check |
| V1-102 | expiry removed from `validate()` **and** the login handler | `test_expired_token_is_refused_at_login` **RED** |

The V1-101 row is the whole argument for that row being L4: the bundler cannot see it. The two V1-102 rows are why the new test is not a restatement of the old one.

## Ready for production verification

`V1-101` and `V1-102` are both **READY_FOR_PROD_VERIFICATION** and neither is blocked on engineering. One authenticated `/admin` run on a deployed SHA covers both — exact 8-step recipe in `docs/master-site-audit/evidence/V1-101/L4_PRODUCTION_RECIPE.md`. It mints at a **non-default** duration on purpose: a run at the default 12 h would leave a hardcoded 12 indistinguishable from a wired control. `V1-131`'s L3 recipe is in `docs/master-site-audit/evidence/V1-131/L3_PRODUCTION_RECIPE.md`.

## Owner decisions outstanding

- **`OD-05`** — premium token direction. Blocks `V1-110`; also the reason the `V1-109` type-scale finding is reported rather than fixed.
- **`V1-105` scope call** (#1006) — accept the `/rankings` pre-feature baseline as permanently unavailable, or hold the row open.
- **V1-35 / Roster lane** — the metric-separation assertion above. Drafted repair supplied; not applied here.

**FROZEN pending Integration. Do not merge. No V1 status column edited. `V1-45` deliberately not started in this PR.**


jasonleetucker-code:
Traffic control: exact-head `0930016ab59b4ec35e74026bb226fb8fd2f6b9fe` is currently RED in both PR Validation (#2129) and E2E Safety Net (#170), and GitHub reports the PR non-mergeable against current `main`. Do not rerun blindly or chase data-refresh movement. Diagnose the actual failing jobs first; then perform the one planned reconciliation only if the failure/conflict is causal. V1-131/V1-101/V1-102 evidence remains useful, but this PR is not merge-eligible while exact-head CI is red.

jasonleetucker-code:
Traffic-control update from live exact-head audit of `0930016ab59b4ec35e74026bb226fb8fd2f6b9fe`:

**PR Validation #2129 has a genuine branch-caused blocking red**, not a rerun candidate. `tests/api/test_feature_flag_reachability.py::test_no_flag_is_read_through_a_variable` fails because `server.py:11593` calls `is_enabled(...)` through a non-literal/variable flag name. The reachability guard cannot prove that gate and explicitly says the verdict would be unsound. Fix the bounded V1-131 implementation so the `consensus_edge` lookup is statically visible (prefer the literal existing flag at this one capability site) unless there is a demonstrated reason the canonical guard itself must understand a new indirection. Do not weaken the guard.

**E2E Safety Net #170 is also genuinely red**, but its two failures are not yet proven causal to this PR: (1) the rankings `/` global-search shortcut never exposes `Search players, picks`; (2) `/arbitrage` never shows `Pick a team and scan`. Both fail again on retry. Before changing either surface, compare these exact specs against the PR base/current main and inspect the uploaded traces/screenshots. If they reproduce on base, classify as pre-existing/stale-test debt rather than folding unrelated repairs into #1086. If branch-only, fix the smallest causal regression.

The mobile pinned-overlap test introduced/affected by this work is green on both desktop and mobile projects, and the `/admin` non-empty guest-pass expiry E2E is also green. Preserve those results.

Do **not** blind-rerun either workflow. Diagnose/fix, push one new exact head, then let blocking CI evaluate that head. Once green, freeze and hand to Claude 5; do not self-merge or edit V1 status.

jasonleetucker-code:
Traffic-control update on NEW exact head `21f29ef03480b14021e5925ec1105d4a02c92095`: the earlier non-literal feature-flag reachability red appears to be past, but this head is still NOT merge-eligible. PR Validation #2132 fails a different hard test: `tests/api/test_nav_gated_features.py::test_auth_status_reports_the_capability_on` expected `features.consensusEdge.available is True` and received `False` (1 failed, 1324 passed, 1 skipped before stop). Diagnose the canonical capability-state setup/contract; do not weaken the test or coerce unavailable/unknown to true. E2E Safety Net #171 is also still red with the SAME two persistent desktop failures as the prior head: rankings `/` global search never exposes `Search players, picks`, and `/arbitrage` never shows `Pick a team and scan`; both fail again on retry (192 passed / 62 skipped / 2 failed). The `/admin` non-empty guest-pass E2E and mobile trade-FAB overlap remain green. Do not blind-rerun. Establish base-vs-branch causality for the two E2E journeys and fix only branch-caused defects; for the auth-status unit red, repair the smallest semantic cause, push one new head, then let exact-head CI judge it.

jasonleetucker-code:
## New exact head `ed4c0f64c` — two causal repairs, no blind reruns

Both defects were mine. Neither was flaky; neither is fixed by coercion.

### 1. PR Validation — the implementation was right, the **test** asserted the old contract

`test_auth_status_reports_the_capability_on` set the flag ON and asserted `available is True` while establishing only **one** of the two canonical prerequisites. It passed on a host with a contract already loaded and failed in CI, where none is — correctly, because the board also 503s `data_not_ready` and `available` means *"the board can be served"*.

Proven directly rather than argued, flag forced on:

```
flag ON  + NO contract -> {'consensusEdge': {'available': False}}
flag ON  + contract    -> {'consensusEdge': {'available': True}}
flag OFF + contract    -> {'consensusEdge': {'available': False}}
```

This is **option 1** — an under-specified ON fixture. The implementation is unchanged.

The two tests encoding the flag-only contract were **my own**, written in the earlier pass and left behind when the implementation was strengthened. Replaced with stronger, environment-independent assertions rather than deleted:

- the ON test now establishes **both** prerequisites (contract set *inside* the client context — entering it fires startup, which overwrites a value set beforehand);
- `test_capability_tracks_the_endpoints_rather_than_a_constant` holds the contract fixed and asserts agreement with the **board**, never the flag. `reported is enabled` re-encoded exactly the conflation this capability was rewritten to remove, and is what made it host-dependent.

No field coerced, no guard weakened, no second health definition.

### 2. E2E — one real branch regression, one real pre-existing failure

Controlled: both builds served from their own `.next` against **one shared backend**, run back to back.

| spec | base `978bf5178` | branch | class |
|---|---|---|---|
| rankings global `/` search | **PASS** | **FAIL** | **A — branch regression** → fixed |
| `/arbitrage` "Pick a team and scan" | **FAIL** | **FAIL** | **B — pre-existing on main** → not fixed here |

**The regression:** `ReferenceError: capabilities is not defined`. The capability was threaded into `<CommandPalette>` from inside `InnerAppShell`, but `capabilities` was a parameter of the **outer** `AppShell` only — a free variable. The shell threw the moment anyone pressed `/`, so the palette stopped opening app-wide. Fixed by threading the prop down the real chain exactly as `authenticated` already travels.

Worth recording: **the production build passed and 2,363 unit tests passed, and could not have caught it** — the component tests render `CommandPalette` directly and supply the prop themselves, stubbing out the one thing that was broken. The guard now lives where it can see it: `tests/e2e/specs/nav-capability-shell.spec.js` asserts, on a production build, that the shell raises no uncaught error, `/` opens the palette, the gated destination is offered only in **agreement** with the server's published capability (not a flat absence — that inverts correctly if the feature ever ships available), and that no second shell-level `/api/consensus-edge/*` request exists.

**`/arbitrage` is not fixed here.** Nothing in this PR touches that route; it fails identically on base with no #1086 code present. Whether the assertion is a real defect or a stale expectation needs that route's owner. **The assertion was not weakened, skipped or deleted.** Full proof: `docs/master-site-audit/evidence/V1-131/E2E_CAUSALITY_2026-08-24.md`.

A measurement trap is recorded there too: an earlier run of this comparison reached the **opposite** conclusion, because a stale `next-server` still held `:3000` (`EADDRINUSE` in its log, so the intended restart never happened) and was serving chunk hashes from a deleted `.next`. Check for `EADDRINUSE` before trusting a local base-vs-branch result.

### Mutation

Forcing `is_available()` **fail-open** → **8 of 23 RED**, every unavailable case; restore → 23 GREEN.

### Gates at this head

nav-gated 23/23 · feature-flag reachability + auth + admin + guest-pass **148/148** · frontend Vitest **2,363/2,363** (160 files) · focused E2E **74 passed / 0 failed** incl. `journey-rankings` global search, `rankings-windowing`, `api-view-parity`, `admin-guest-pass` and `mobile-pinned-overlap` on both projects · build green, **14/14** bundle budgets · `ruff check .` + `ruff format --check .` clean repo-wide · planning integrity, decision coercions, work claims clean.

### Two corrections to the original PR body

- The capability shape is now **`features.consensusEdge.available`** (nested), not the flat boolean the body describes.
- `available` is **flag AND loaded contract**, not the flag alone.

### Reconciliation

**Not performed, deliberately.** GitHub reports `mergeable_state: unstable` — mergeable with failing checks, not `dirty`. `git merge-tree` shows **0 conflict markers**; the only file both sides touch is `docs/WORK_CLAIMS.md`, which merges cleanly. Per the release discipline, class-C data drift alone is not a reason to move a head, so no merge was made.

Not calling `READY_FOR_INTEGRATION` — awaiting exact-head blocking CI on `ed4c0f64c`.

---

jasonleetucker-code (commented):
Traffic control @ exact head `ed4c0f64c4dea1163c72f95990c09f21381cfb77`: current E2E Safety Net #172 never reaches Playwright. Its preflight fails on the shared full-contract stop: `partial_run_critical:IDPTradeCalc` plus `blend_integrity_violation` on `2028 Late 3rd`. Do **not** chase the prior rankings-search/arbitrage browser failures from older heads on this run. PR Validation #2133 still has an independent deterministic red in `test_public_league_privacy_boundary.py`: anonymous `rosTeamStrength.csv` returns 429 where the boundary expects 401/403/404. Diagnose/fix only that branch-local privacy failure; leave E2E rerun until the shared contract stop is cleared. No merge while either gate is red.

jasonleetucker-code:
## New exact head `23b3077fc` — and one blocker that is not Lane 6's to clear

The two failures at `ed4c0f64c` are different things, and only one is this PR's.

### PR Validation — partly this PR's, fixed

The blocking failure was in a test this change never touches:

```
test_public_league_privacy_boundary.py::test_the_csv_variant_is_closed_too[rosTeamStrength]
  → "rosTeamStrength.csv answered 429 anonymously"
```

**429 is the rate limiter, not an auth defect** — the route is fine and the whole class passes in isolation (9/9). But `/api/auth/status` and `/api/auth/login` are both in `server.py::_PUBLIC_API_EXACT`, so they spend from a **60/min per-IP budget shared by the entire suite**, and under `TestClient` every test is the same client IP. This PR's two new modules add ~60 such calls and run alphabetically **before** `test_public_league_*`, so the accumulated budget runs out inside an unrelated test, which then fails on 429 instead of its own assertion.

Repaired at the source of the pressure rather than at the victim: both new modules carry an autouse fixture calling `rate_limit.reset_for_tests()` — the limiter's own sanctioned hook. **Production rate limiting is unchanged and no assertion anywhere is weakened.**

The other two `##[error]` lines in that job are on steps carrying `continue-on-error: true` (live-data tier, `check_source_health`) — advisory and non-blocking by design, exactly as `CLAUDE.md` describes.

### E2E Safety Net — **not this PR's, and not fixable from this lane**

This supersedes my earlier base-vs-branch analysis of the two journey specs. At `ed4c0f64c` the job **never ran a single spec** — it died in `npm run regression:preflight`, uploading no `playwright-report`, no `test-results` and no server logs, because none were produced:

```
[contract] ok=False  structuralErrors=1  sourceHealthErrors=1
[contract][error] blend_integrity_violation:1 row(s) hold a value outside the
                  range of their own source contributions (2028 Late 3rd)
```

Established three ways that this is **main's data, not this branch's code**:

1. the diff touches **no** `exports/`, `data/`, `data_contract`, canonical-valuation or pick-pricing file;
2. the validated snapshot `exports/latest/dynasty_data_2026-08-24.json` is **byte-identical** between this branch and `origin/main` — it arrives via `chore: automated data refresh 2026-08-24T21:15:54Z` (`64f382343`);
3. running the same validator on a **clean `origin/main` worktree with zero #1086 code present** reproduces it exactly, same `2028 Late 3rd` row.

This is the system working as designed. `CLAUDE.md` is explicit that a blend-integrity violation is never coerced, because *"coercing an impossible number to a plausible one hides a pipeline fault."*

**Stating the consequence plainly rather than grinding against it: the E2E Safety Net cannot be made green from Lane 6.** The repair belongs to whoever owns the valuation pipeline / 2028 pick pricing, and the condition blocks **every** PR that runs this job while that snapshot stands. Classification **D — external blocker on main**. Not worked around; the preflight gate was not weakened.

### Status

**NOT `READY_FOR_INTEGRATION`.** Blocking CI cannot be green while the snapshot defect stands, and that is an upstream fix, not mine. Everything within this lane's scope is complete and evidenced:

- V1-131 implemented, capability = flag **AND** loaded contract, mutation-proved (fail-open → 8 of 23 RED)
- the `ReferenceError: capabilities is not defined` shell regression found and fixed, with a production-build guard so the class cannot recur
- V1-102 evidence closed to L2 with 16 added guards; corrected row text drafted for Claude 5
- V1-109's FAB/tray overlap fixed and verified (44×44 → 0)

Full record: `docs/master-site-audit/evidence/V1-131/E2E_CAUSALITY_2026-08-24.md`.

I have not merged, not edited any V1 status, and not started V1-45.

---

jasonleetucker-code:
## E2E cleared the preflight and ran — head `b2b724efd`

E2E Safety Net **#174** no longer stops at the contract gate. The shared full-contract preflight passed (main's snapshot refreshed), the stack booted, and Playwright executed:

```
1 failed · 62 skipped · 195 passed (6.2m)
  [desktop-1366] journey-trade.spec.js:146
    › /arbitrage scans and renders either trades or an explicit empty state
```

**The rankings global-search failure is gone.** Prior heads failed *two* specs; this head fails one. That is CI independently confirming the `ReferenceError: capabilities is not defined` repair — `capabilities` was referenced inside `InnerAppShell`, which never received it, so the shell threw the moment anyone pressed `/`. It is now threaded down the real chain (`AppShell` → `PrivateAppShell`/`NoPlayerDataAppShell` → `InnerAppShell` → `CommandPalette`).

**The one remaining failure is the pre-existing one.** `/arbitrage` "Pick a team and scan" fails identically on clean `origin/main` `978bf5178` with zero #1086 code present, from the same backend, on consecutive runs — the controlled base-vs-branch comparison is in `docs/master-site-audit/evidence/V1-131/E2E_CAUSALITY_2026-08-24.md`. Nothing in this PR touches that route. Not folded in, not weakened, not skipped.

## The 429 privacy red — measured, base vs branch

Anonymous `GET /api/public/league/rosTeamStrength.csv`, one process, both refs:

| | fresh limiter | exhausted | after reset | 429 body |
|---|---|---|---|---|
| branch | **401** | **429** | **401** | `error`, `message`, `retryAfterSeconds` |
| base `478d249ef` | **401** | **429** | **401** | `error`, `message`, `retryAfterSeconds` |

**Identical — so not option A.** `_private_api_gate` rate-limits only paths where `_is_public_api_path` is true and applies the blanket 401 only where it is false; `"/api/public/league"` is in `_PUBLIC_API_PREFIXES`, so the 401 comes from the **handler's** per-section check, which a 429 short-circuits. **Classification C**, ordering half also pre-existing (**B**). This PR contributes ~60 extra rate-limited `/api/auth/*` calls earlier in suite order, tipping the shared 60/min per-IP budget.

**The privacy invariant holds and 429 was not allowlisted.** Fresh limiter → 401, fail-closed. The 429 body carries no league payload or section data — the resource does not become observable because the response is 429. No expected-status list widened, no assertion relaxed. Repair is upstream of the symptom: this PR's two new test modules reset the limiter they perturb via `rate_limit.reset_for_tests()`.

## Status

**PR Validation on `b2b724efd` has not reported yet** — that is the outstanding gate, and it is the authoritative test of the 429 repair. I am not calling `READY_FOR_INTEGRATION` until it lands.

Excluding the pre-existing `/arbitrage` failure, branch-local work is complete and evidenced: V1-131 (capability = flag **AND** loaded contract, mutation-proved), V1-109 FAB/tray repair, V1-101 evidence, V1-102 evidence to L2 with drafted row text, V1-106 untouched.

Not merged. No V1 status edited. V1-45 not started.

---

jasonleetucker-code:
Post-merge production traffic-control finding for V1-131: authenticated production run 32919648642 / job 98030516666 proves the API + desktop gating path, but the 390x844 mobile drawer is genuinely red. After `nav.shell-tabbar[aria-label="Primary"]` renders and the real `Menu` button is clicked, `.shell-drawer-group` never appears within 15s. The uploaded failure screenshot still shows the Rankings page and bottom tab bar with no drawer. This is not a stale selector: current `MobileChrome.jsx` still renders `.shell-drawer-group` inside `DrawerGroup`. Exact browser result was 8 passed / 3 failed / 9 skipped; V1-131 desktop capability/board agreement, Market menu/palette/more anchors, hard-reload network invariant, and direct route all passed; only its mobile branch failed. Keep V1-131 IMPLEMENTED_UNVERIFIED. Next repair should be bounded to why `menuOpen=true` is not producing the Drawer in deployed mobile; do not change Consensus Edge capability methodology or weaken the L3 test. Do not blind-rerun until a causal repair deploys.

PR #1078, opened 2026-08-24
V1-27: make the C2-U1 §10 production checklist executable

jasonleetucker-code:
## Lane census first

Parsed every V1 row at `131abf9f9`. **My lane (L1 / Roster Intelligence) has 9 rows; 8 are `VERIFIED`.** `V1-27` is the only one open (`IMPLEMENTED_UNVERIFIED`, required L3). The other open rows mentioning roster concepts belong to L4/L5/L6 (source acquisition, frontend) and are not mine. So this PR targets the one closable row in the lane.

**Work-claim check:** no open claim touches `src/ros/lineup.py`, `tests/lineup/`, or lineup verification. Lane 8 and V1-21 claim `src/api/data_contract.py` — **not touched by this PR**.

## The gap

`V1-27`'s bar is the `C2-U1` §10 checklist. §10a records it being run **by hand** on 2026-08-18, and records in its own words that the run *nearly passed vacuously*:

> The first run joined assignments to eligibility by Sleeper player id and reported **0** hits — which reads exactly like "the property does not hold". It was a join failure.

A checklist whose correctness depends on the operator reproducing that reasoning is not a checklist. There was no executable instrument — `optimalLineup` appeared in tests but in no script.

## What this adds

`scripts/verify_lineup_production.py` — the same checks as code.

- **Computes no lineup.** Every check reads what production published; the solve stays in `src/ros/lineup.py`. An instrument that recomputes what it checks verifies only that it agrees with itself.
- **No second harness.** PASS/FAIL/UNMEASURABLE, the vacuous-pass downgrade, level capping and exit-code precedence are *imported* from `verify_roster_intelligence`, which already owns "how a check reports itself".
- **4 of §10's 6 items automated.** Items 2 (browser render) and 4 (needs a **pre-deploy** snapshot that was never captured) print on every run as explicitly not automatable — a partial verification must not quietly read as a complete one.

### Two traps encoded, both real

1. **Name-vs-id join** (§10a's near-false-pass). Check `05` publishes `assignmentsJoined` and returns **UNMEASURABLE — never FAIL** on a 0-row join. "Matched nothing" is not "found none".
[...]

PR #1088, opened 2026-08-24
V1-36 HELD at IN PROGRESS: the single-ownership guard does not guard ownership

jasonleetucker-code:
**Tally unchanged: 89 / 136 = 65.4%.** Docs-only. `#1079` is merged and its shipping evidence **reproduces** — but V1-36 is **not** promoted, because the guard that enforces its headline capability was defeated in one edit, independently, by two agents.

## Three of four guards passed

**1 — no arbitrage leak.** `equal_count_only` / `ARBITRAGE_CONTROL_KEY` / `excludePlayers` appear 17 times repo-wide, **all 17 in `finder.py`**; ten arbitrage symbols probe absent from the loaded `suggestions` module. Behaviourally stronger than any grep: **89 of 176** measured live suggestions are shape `(2,1)` — unequal-count packages that could not exist under finder's equal-count rule. `#1076` never touched `src/packages` (the owner's last commit predates it by five days), so indirect inheritance is impossible by construction.

**2 — methodology unchanged, measured.** Production path over all 12 real `dynasty_main` teams × pool caps 150/300/800 from the pinned golden fixture:

| | before (`a3e7bdb82^`) | after |
|---|---|---|
| suggestions | 176 | 176 |
| payload sha256 | `90ea6110…` | `90ea6110…` |

Byte-identical. 0 of the file's 24 module constants moved. **The null is proven non-vacuous**: reinstating the one retired hand-rolled key at `:1161` moves the sha and makes the engine recommend acquiring **Nik Bonitto — already on that team's roster**.

**4 — CI green on the merge tree.** `PR Validation` `32765081070` success on the PR's exact final head `7822a2a4b3`, and `git log fa45be6bc..a3e7bdb82^` = **0 commits** — the validated tree is the landed tree.

## The fourth is disqualifying

**3 — the guard does not enforce single ownership.** Delete line 27's `from src.packages import …`, reimplement `_identity_key` and `_side_identity` locally, keep the docstrings:

- `'src.packages' in sys.modules` → **False**. Zero dependency on the canonical owner.
- `tests/trade/test_suggestions_package_identity.py` → **11 passed**.
- `tests/trade/ tests/packages/ tests/test_trade_suggestions.py tests/roster_intel/` → **1608 passed, byte-identical to baseline.**

The capability this row claims can be wholly removed with everything green. Three independent causes:

1. The routing assertions scan `inspect.getsource`, which **includes the docstring** — both helpers name the owner only in prose once the body is stripped.
2. `_code_of` uses `ast.unparse`, which normalises literals to **single** quotes while the guard hunts **double**-quoted spellings — 3 of its 5 patterns are structurally unmatchable.
3. Both scans iterate hardcoded function-name lists, with a positive check any decoy call satisfies.

Promoting on a guard the adjudicator personally defeated would be exactly the over-promotion this campaign exists to prevent.

## Exact remaining acceptance

**Engineering (blocking, no owner decision):** assert over the docstring-stripped AST body; match node structure rather than quoted literals; assert `src.packages` is imported and its symbols appear in **executable** code; scan every module-level function rather than two name lists; and **prove the repair by requiring the exact bypass above to go RED**.

**Owner adjudication (not closable by measurement):** the row's literal acceptance names `package_key`, which has **no call site** in `suggestions.py` — `side_key` was substituted with an explicitly deferred product question ("should duplicate packages collapse across categories?"). And whether identity-only consumption satisfies "ONE shared package generator" at all, given the file keeps its own 1v1/2v1 shapes — which guard 1 was tasked with **preserving**. The row's two blockers pull in opposite directions.

Any future promotion must carry two qualifications: `src/bdvm/roster.py::scan_double_positive_trades` is a deliberate **second** two-sided generator consuming nothing from `src/packages`, so the claim is domain-scoped; and consumption is at the **name-fallback grade** (`asset_id=""` on every call — the branch the owner's own docstring calls "a real defect"), unlabelled in the payload.

## Also corrected — four false canonical statements

All said `suggestions.py` is still an un-migrated holdout: the V1-36 row, and `C_SERIES_SCOPE_MANIFEST.md:122` and `:258`. The manifest's "zero `constraints.py`/`roster_capacity.py` integration" is corrected to partial (`#962` closed the constraints half). Staleness that understates progress isn't a false green, but four false statements in canonical records are worth fixing.

## Routed, not folded in

`_opponent_fit_label` (`:1107`, `:1125`) joins over a **set**, so `/api/trade/suggestions` is genuinely **nondeterministic across processes** — 6 `opponentFit` fields differ across 3 teams at `PYTHONHASHSEED` 0/1/7. Proven pre-existing (byte-identical across `a3e7bdb82^..a3e7bdb82`); one-line `sorted()` fix. Consequence that travels: every measurement of this endpoint is valid only seed-pinned.

## Gates

`check_planning_integrity.py` **OK 13/13** incl. tally agreement · `check_decision_coercions.py` **clean** 660/660

---


jasonleetucker-code:
**CI red on this head, and it reproduces on the base branch. Not caused by this PR.**

This PR's diff is two markdown files. It cannot affect the contract, the board, or source health.

Reproduced on `main` itself at `935fef337`, no PR involved:

```
$ python3 scripts/validate_api_contract.py --repo . --lane structural
[contract] source=exports/latest/dynasty_data_2026-08-24.json
[contract] ok=False errors=2 warnings=18 players=993
[contract] structuralErrors=1 sourceHealthErrors=1
[contract][error] blend_integrity_violation:1 row(s) hold a value outside the
                  range of their own source contributions (2028 Late 3rd)
```

**Root cause is one upstream condition with three symptoms.** `partial_run_critical:IDPTradeCalc` — IDPTradeCalc ran partial in the last scrape, and it is one of only **two** families that price picks. That cascades:

1. the board drops to **993 players** (from 1,111 on a healthy cycle);
2. `tests/api/test_picks_end_to_end.py::test_every_pick_has_source_ranks` fails — `['2029 Early 1st', '2029 Late 1st', '2029 Mid 1st']` have no `sourceRanks`, because the far-future derivation depends on the pick markets that went partial;
3. `2028 Late 3rd` trips `blend_integrity_violation` — the same transient class as the `2027 Late 1st` finding investigated in #1062/#1063, which self-corrected on the next cycle.

**Production is healthy and unaffected** — `/api/status` reports `contract_ok: true` with 1,115 players, serving a good board from an earlier cycle. The `--lane full` deploy gate refusing to ship a board built without a critical market is fail-safe working as designed.

**No action taken, deliberately.** Fixing this by widening a tolerance, reclassifying `blend_integrity_violation` out of the structural lane, or regenerating the tracked payload would each be a validator weakening — explicitly out of scope, and it would convert a truthful degraded state into a false green. The 2-hourly refresh clears this condition with no override, which is the documented behaviour.

**Consequence for the queue, worth stating plainly:** while `main` carries this payload, *every* open PR will show the same red, including #1081, #1083, #1084 and #1078. Their CI is not evidence about their code until the refresh lands. I will merge-forward and re-run this head once it does.

---

PR #1089, opened 2026-08-24
V1-36: close the package-mechanics guard false negative (Guard A) + package_key semantics evidence (Guard B, no decision)

jasonleetucker-code:
Response to the #1088 adversarial audit. Starting SHA `a3c39113`, head `dc4ecdbf`.

**V1-36 is NOT promoted. No package methodology reopened. No product code touched** — Guard A is test-only, Guard B is docs-only.

> ⚠️ **Production stop:** there is an active production deployment stop unrelated to this guard. This is **not** required by that repair and should **not** be merged ahead of it.

---

## Guard A — the false negative, closed

#1088 is right that the guard has a false negative. **It is not right about the mechanism**, and acting on the stated mechanism would have produced a guard that still missed the real cases. Both corrections are recorded in the module docstring rather than smoothed over.

### What was actually escaping

Retired detector: `re.compile(r"abs\([^)]*\)\s*<=\s*\d")` AND the literal text `combinations(`. Probed directly:

| form | old regex |
|---|---|
| `abs(send_n - recv_n) <= 1` | MATCH |
| `d = len(a) - len(b)` … `abs(d) <= 1` | MATCH |
| `abs(len(send) - len(receive)) <= 1` | **MISS** |
| `abs(send_n - recv_n) <= MAX_DIFF` | **MISS** |
| `d = len(a) - len(b)` … `abs(d) <= MAX` | **MISS** |
| `abs(player_count(s) - player_count(r)) <= 1` | **MISS** |

**Correction 1 — local-variable aliasing alone does NOT evade it.** Row 2 matches. What evades is a `)` *inside* the `abs(...)` argument (`[^)]*` stops at the first one) or a non-digit bound. The canonical spelling `abs(len(send) - len(receive)) <= 1` fails on both counts, and `player_count` is *the owner's own API name* — so the tidiest reimplementation was the least detectable.

**Correction 2 — `suggestions.py` was not escaping through this hole.** It contains **no `combinations(` at all**, so the co-occurrence AND can never fire there, and every one of its `abs()` sites compares **values** (`abs(p.display_value - gap_needed) < FAIRNESS_TOLERANCE`, where `gap_needed = target.display_value - ws_ev`) rather than player counts. A value-proximity band is not a topology bound.

### Design

That second correction is load-bearing. **Topology compares COUNTS; value proximity compares VALUES.** Keying on the `abs(x - y) <= k` *syntax* alone would flag ~12 legitimate sweetener/fairness bands in `suggestions.py` — converting a false-negative guard into a false-positive one, which the brief explicitly forbids.

So an `abs` difference is topology only when **both operands are count-like** — `len()`, `player_count()`, `.count()`, counting `sum(1 for ...)` — resolved through simple local assignment aliases, with a deliberately narrow count-shaped-identifier fallback for unresolvable parameters. Bounds are unconstrained (digit, named constant, attribute). Enumeration detection broadened beyond `combinations` to `permutations`/`product`. No filename-specific ban.

### Mutation bar — all five items

| # | result |
|---|---|
| 1 | baseline shipping tree **GREEN** (6/6) |
| 2 | direct topology injected → **RED**, names `src/trade/_mut_direct_topology.py` |
| 3 | suggestions-style precomputed-gap (`raw_gap` → `spread` → `abs(spread) <= TOLERANCE`) → **RED**, names `src/trade/_mut_aliased_topology.py` |
| 4 | mutations removed → **GREEN** |
| 5 | each mutation's content confirmed present before counting it |

Both mutations were **first confirmed GREEN under the retired regex**, so the RED is attributable to this change and not to the injection itself.

### False-positive census

**41 modules** across `src/trade` + `src/roster_intel`: **0** topology bounds detected, **0** offenders. `suggestions.py`'s value bands are correctly untouched. Non-vacuity is pinned three ways: against the owner file, against 8 distinct topology spellings, and by the two live mutations.

---

## Guard B — evidence only, deliberately undecided

`docs/trade/V1_36_PACKAGE_KEY_SEMANTICS_EVIDENCE.md`. Contains: every current `package_key` use (exactly **one** consumer — `roster_intel/packages.py`), every `suggestions.py` local key with its grain, a reproduced collision + non-collision control, measured user-visible impact, and the contract/manifest wording.

**The measurement is the part most likely to be assumed wrong.** On the **production default** regime, whole-package key yields **fewer** suggestions (`buyLow` 3 → 2), not more — the downstream quality-filter caps are budgets, so feeding them more raw candidates changes which ones consume the budget. Small but non-zero, so it cannot be adopted as a no-op canonicalisation. Mechanism stated as a hypothesis and flagged as such.

**Both readings have textual support**, which is why this needs adjudication: the manifest credits `roster_intel/packages.py` as a consumer for *identity-only* use while retaining its own search, yet the V1 contract still calls `suggestions.py` an *un-migrated holdout* for structurally the same thing. Both cannot be the standard.

---

## Gates

`ruff format --check .` 1312 files ✓ (run under the repo's pinned `ruff~=0.6.0`, not the sandbox 0.15.x) · `ruff check` ✓ · `check_decision_coercions.py` clean · `check_planning_integrity.py` OK · `tests/packages/ tests/trade/ tests/roster_intel/ tests/test_trade_suggestions.py` — **1611 passed, 22 skipped, 1 failed**.

The one failure is `tests/roster_intel/test_metric_separation.py::test_metrics_genuinely_diverge_on_a_representative_real_board` (`assert 1 != 1`). **Reproduced on pristine `main` with none of my code present** — pre-existing, data-driven, in the roster-metrics lane, outside trade ownership. Reported, not fixed.

**READY_FOR_INTEGRATION: Guard A only** (behind the production stop).
**OWNER / INTEGRATION_DECISION_REQUIRED: Guard B.**

---


jasonleetucker-code:
CI red at `dc4ecdbf`. **All three failing steps are pre-existing and none is this PR's** — this diff is test-only. Reply rather than a push, per drive-to-green.

### What actually failed

| # | step | failure | mine? |
|---|---|---|---|
| 1 | hard gate `pytest -x -m "not livedata"` | `test_metric_separation.py::test_metrics_genuinely_diverge_on_a_representative_real_board` — `assert 1 != 1` | **no** |
| 2 | `pytest -m livedata` | KTC rank-12 drift: pinned +12.3% vs actual +15.4%, band ±3.0pp | **no** |
| 3 | `check_source_health.py` | `idpTradeCalc` raw CSV byte-identical 40 days (budget 14d) | **no** |

**Guard A itself passed.** The hard gate reports `1 failed, 7482 passed, 20 skipped` — `tests/packages/` collects before `tests/roster_intel/`, so this PR's `tests/packages/test_single_owner.py` (6/6) ran and passed among those 7482 before `-x` aborted on someone else's test.

I reproduced #1 on pristine `main` with none of my code present, and reported it in the PR body before this run — same assertion, same value.

### ⚠️ #1 is a repo-wide blocker, not just this PR's problem

Worth surfacing given the active production stop: `test_metric_separation` is in the **hard, blocking gate** and it runs under `-x`. On current data it fails on `main` itself, which means **it aborts the hard gate for every PR in the repo** — including whatever the production-stop repair needs to land. That is a bigger deal than one red checkmark here.

It is data-driven (`(1, 1)` — the strongest-by-value team happening to also rank first on youth), so it may self-clear on a refresh the way the KTC band did for #1079. But it is not guaranteed to, and nothing about it is durable. Roster-metrics lane, outside trade ownership — flagging it rather than touching it.

Also note #2 is *moving*: 3.233 → 3.077 against a 3.0 band. Drifting toward the band but still outside.

### Position unchanged

Guard A is FEATURE_GREEN on its own evidence: baseline 6/6, both mandated mutations RED→GREEN, 41-module false-positive census clean. Still **behind the production stop** — not required by that repair and should not jump it. Guard B remains OWNER_DECISION_REQUIRED. V1-36 not promoted. Frozen; no further pushes from me.

---

jasonleetucker-code:
Re-run at the same head `dc4ecdbf` (job `97674019328`) produced the **identical outcome** to the previous run — same three failures, none from this diff, which is still test + docs only. No push from me; the lane is frozen.

| # | step | result |
|---|---|---|
| 1 | hard gate | `test_metric_separation` — `assert 1 != 1`, `-x` aborts at 74% (`1 failed, 7482 passed, 20 skipped`) |
| 2 | livedata | KTC rank-12: `3.077 <= 3.0` — byte-for-byte the same numbers as before |
| 3 | `check_source_health` | `idpTradeCalc` byte-identical 40 days (advisory, `continue-on-error`) |

Guard A's own tests are among those 7482 and passed again.

### Correcting my earlier flag

In [my previous comment](https://github.com/jasonleetucker-code/riskittogetthebrisket/pull/1089#issuecomment-5403041996) I said #1 blocks the hard gate **repo-wide**. That was true when written and is **no longer true**: `main` now carries `7cde2bd3` — *"V1-35: replace a single-leader coincidence with a population invariant (#1090)"* — which rewrites precisely that assertion. The Roster lane fixed it.

It was not in this run's tree, which is why the run still went red on it. The merge ref tested was `05a05853f`, computed against base `4032d428a`; `main` advanced to `67daae177` *during* the run, and the release-candidate step reports `tests/roster_intel/test_metric_separation.py` as the single **class-B** change outstanding on base. So the failure is a stale-base artifact of this run, not a live condition — worth knowing for whoever validates the exact head, and no longer something the production-stop repair has to work around.

### Position unchanged

Guard A remains FEATURE_GREEN on its own evidence and READY_FOR_INTEGRATION, explicitly **behind the production stop**. Guard B stays OWNER_DECISION_REQUIRED. V1-36 not promoted. Not re-running, not rebasing, not re-arming a check-in — per the freeze, class-C/data movement is not mine to chase.

---

PR #1090, opened 2026-08-25
V1-35: replace a single-leader coincidence with a population invariant

jasonleetucker-code:
Unblocks the release train. One file, one clause, test-only.

## The defect

`test_metrics_genuinely_diverge_on_a_representative_real_board` asserted, of the single #1-by-Team-Strength team, that it must not *also* rank #1 on Young Core:

```python
strongest = min(pairs, key=lambda p: p[0])
assert strongest[0] != strongest[1]
```

That is a claim about **one data point**, not about the two concepts. The strongest roster in a league is entitled to also be its youngest — and on the current board it is, yielding `(1, 1)`. Two distinct metrics may agree *anywhere*, including at the top; what they may not do is agree *everywhere*.

Reproduces on clean `main` (`13456dd5f`) and blocks #1089 / V1-36, #1084 / V1-49, #1086 / Premium UI RC, and shipping validation.

## The replacement

The docstring already claims the right property — *"the single STRONGEST roster in the league by value while ranking near the BOTTOM on youth"*. That is a statement about **displacement**, so it is now asserted as one, over the population:

```python
displacements = [abs(s - y) for s, y in pairs]
assert max(displacements) >= len(pairs) // 2
```

**No new threshold is introduced.** It reuses the `len(pairs) // 2` scale the inversion bar directly above already uses, and a team displaced by half the league is by construction top-half on one axis and bottom-half on the other. Sharing rank #1 is now legal.

**Strictly stronger than what it replaces.** Inversions treat a one-place shuffle and an eleven-place swing identically, so a Young Core degraded into "Team Strength plus noise" could show 12 of 12 differing and still pass. Requiring real displacement closes that.

## Current-board measurement

| | |
|---|---|
| teams / pairs | 12 / 12 |
| inversions | 10 of 12 |
| **max displacement** | **9** — the #2-by-strength team is #11 on youth |
| bar | 6 |

## Non-vacuity — two mutations of a real production module

Both applied to `src/api/roster_intelligence.py`, each confirmed applied and reverted. Neither is a syntax/import failure.

| # | mutation | inversions | max disp | result |
|---|---|---|---|---|
| 1 | Young Core rank := Team Strength rank (total collapse) | 0 / 12 | 0 | **RED** |
| 2 | …with adjacent ranks swapped (near-collapse) | **12 / 12 — old bar passes** | 1 | **RED on the new clause** |

Mutation 2 is the decisive one: it is precisely the case the replacement exists to catch and the old assertion could not see. Restored — `git diff --stat src/` empty, test GREEN.

## Validation

- `test_metric_separation.py` — 17 passed
- Strength / Young Core focused + single-owner guards — 76 passed
- `tests/roster_intel/` + `tests/lineup/` (`-m "not livedata"`) — 735 passed
- `ruff check` / `ruff format --check` — clean
- `check_decision_coercions.py` — clean
- `check_planning_integrity.py` — OK
- Work-claim collision — the one claim naming this file is `done — READY_FOR_INTEGRATION`, not active

## Proof zero production metric code changed

`git diff --name-only 13456dd5f..HEAD` → `tests/roster_intel/test_metric_separation.py` only. No Team Strength, Young Core, meaningful-core, age weighting, lineup solver, canonical value, or board data touched.

**FEATURE_GREEN / READY_FOR_INTEGRATION.** Frozen — Claude 5 owns current-main reconciliation and shipping validation.

---

### DIFF
--- a/tests/roster_intel/test_metric_separation.py
+++ b/tests/roster_intel/test_metric_separation.py
@@ -523,8 +523,35 @@ def test_metrics_genuinely_diverge_on_a_representative_real_board():
         "either the board is unrepresentative or the two axes have collapsed"
     )
 
-    strongest = min(pairs, key=lambda p: p[0])
-    assert strongest[0] != strongest[1], (
-        "the single strongest-by-value team should not also rank identically "
-        f"on youth by coincidence: {strongest}"
+    # The magnitude half of the same property, stated over the POPULATION.
+    #
+    # This clause used to read ``min(pairs, key=...)`` then
+    # ``assert strongest[0] != strongest[1]`` — the single strongest-by-value
+    # team must not also rank #1 on youth. That asserted a COINCIDENCE about
+    # one data point, not a fact about the two concepts: the best roster in a
+    # league is perfectly entitled to also be its youngest, and on the
+    # 2026-08-24 board it was, giving ``(1, 1)`` and a RED that said nothing
+    # about collapse. Two distinct metrics may agree anywhere, including at
+    # the top; what they may not do is agree EVERYWHERE.
+    #
+    # So the leader is no longer special-cased. What the docstring actually
+    # claims — "a team can be the single STRONGEST roster in the league by
+    # value while ranking near the BOTTOM on youth" — is a statement about
+    # displacement, and it is asserted here directly: somewhere in the
+    # population, a team's two ranks must sit at least half a league apart.
+    # That is the same ``len(pairs) // 2`` scale the inversion bar above
+    # already uses, so no new threshold is introduced; a team displaced by
+    # half the league is by construction top-half on one axis and bottom-half
+    # on the other.
+    #
+    # It is also strictly stronger than counting inversions. Inversions treat
+    # a one-place shuffle and an eleven-place swing identically, so a Young
+    # Core that had degenerated into "Team Strength plus noise" could show 12
+    # of 12 differing and still pass. Requiring real displacement closes that.
+    displacements = [abs(s - y) for s, y in pairs]
+    assert max(displacements) >= len(pairs) // 2, (
+        "Team Strength and Young Core Index rank every team within "
+        f"{max(displacements)} places of each other (need >= {len(pairs) // 2}). "
+        "No team is strong-but-old or weak-but-young, so the two axes are "
+        "tracking one another rather than measuring different things."
     )


jasonleetucker-code:
**Integration review — APPROVED on the merits.** Verified independently at exact head `a681d10e0`, not accepted from the PR body.

## Checklist

| requirement | result |
|---|---|
| zero production metric code changed | **PASS** — 1 file, `tests/roster_intel/test_metric_separation.py`, +31/−4. Nothing outside `tests/` |
| population-level invariant, not a single-leader coincidence | **PASS** — `max(displacements) >= len(pairs) // 2`, reusing the existing scale; no new threshold introduced |
| baseline green | **PASS** — 17 passed |

## Mutation proof — I ran both myself, in an isolated `git archive` tree

**M1 — full collapse.** Production mutation in `src/api/roster_intelligence.py`: `agePortfolio.leagueRank := strength.leagueRank`.

```
AssertionError: only 0 of 12 differed at all — ... the two axes have collapsed
assert 0 >= (12 // 2)   pairs = [(9,9), (2,2), (6,6), (7,7), ...]
```
**RED** at the inversion clause, correct semantic reason.

**M2 — the one that actually tests your new clause.** M1 trips the *old* inversion bar first, so it leaves the added assertion unproven. So I built the exact case your comment claims to close — "a Young Core that had degenerated into Team Strength plus noise could show 12 of 12 differing and still pass" — by adjacent-swapping the rank:

```
AssertionError: assert 1 >= (12 // 2)
  where 1 = max([1, 1, 1, 1, 1, 1, ...])
  and  12 = len([(9,10), (2,1), (6,5), (7,8), (4,3), (1,2), ...])
```

**12 of 12 teams differ — the old bar would have PASSED — and the new clause fires** at line 552. Your claim that this is *strictly stronger* than counting inversions is therefore measured, not asserted. That is the part I'd have pushed back on had it not held.

## Judgement on the change itself

Correct, and for the right reason. The retired line asserted a coincidence about one data point: the best roster in a league is perfectly entitled to also be its youngest, and on the 2026-08-24 board it was — `(1, 1)` — producing a RED that said nothing about collapse. Two distinct metrics may agree anywhere; what they may not do is agree everywhere. The replacement says exactly that.

## Merge status

Held only on the shared release gate, which is **not this PR**. Exact-head CI will also surface:

```
tests/canonical/test_ktc_reconciliation.py::test_rank_delta_within_tolerance[12-8561-12.3-3.0]
assert 3.077358490566038 <= 3.0
```

That reproduces on clean main and is unrelated to your change — our Hill curve is provably unchanged (the test's own `ours == 8561` assertion passes); KTC's **rank-12 occupancy** moved. It is escalated to the owner as the same decision as #895. **Nothing is being asked of this lane for it.** I will merge the instant that clears.

**V1-35 stays VERIFIED; this repair does not move the numerator.**

---

PR #1091, opened 2026-08-25
Lane 8: fail-closed coverage for the fetch-fresh / content-stale detector (IDPTradeCalc)

jasonleetucker-code:
## Root cause first: **A. UPSTREAM_UNCHANGED** — measured, not inferred

I fetched the vendor's live Apps Script payload and hashed its 84 pick rows under the same `value_sftep` precedence the scraper uses:

```
live   vendor pick sha256 : 8b570b4c1790ae8f2015d3020d2ee22c
stored artifact  sha256   : 8b570b4c1790ae8f2015d3020d2ee22c
84 common · 0 differing · 0 only-live · 0 only-stored
```

Every competing cause is excluded by direct measurement, not by reasoning from HTTP 200:

| Cause | Verdict | Evidence |
|---|---|---|
| **A** upstream unchanged | **CONFIRMED** | live payload byte-identical to stored artifact |
| B vendor/CDN stale cache | excluded | `Cache-Control: no-cache, no-store, max-age=0, must-revalidate`, `Expires: Mon, 01 Jan 1990`, no `Age`/`X-Cache` hit |
| C auth/session fallback | excluded | no cookie jar, session file, token or auth env var exists on this path |
| D failed refresh reusing old artifact | excluded | `siteRawPreserved: []`, `siteRawFresh: [idpTradeCalc.csv…]` |
| E parser collapse | excluded | if the parser collapsed new content to old, live ≠ stored; they are identical |

**The pick data is not wrong — it is verifiably current.** What is 41 days old is the last *change*, not the *data*. The vendor has simply not republished its pick tiers.

Documentation correction found along the way: picks come from **Sheet2 (48 slot rows) AND Sheet3 (36 tier rows)**, not Sheet2 alone as the code comment claims.

## What this PR actually changes

`measure_content_staleness` in `scripts/check_source_health.py` is the **only** place in the repo that can distinguish "the vendor published something new" from "the fetch succeeded". Every other freshness surface reads the `_last_success` stamp, which is written on every successful fetch regardless of content — a deliberate choice, documented in `_build_source_timestamps`, and not what this PR questions.

The consequence is that this one function carries the whole of the repo's ability to observe a frozen vendor, and it had **zero test coverage**. Nothing stopped it being deleted, or quietly losing the property that actually matters: **pick rows are hashed separately from the board**, because a vendor can refresh its player board every run while its pick tiers sit frozen.

That is exactly the live idpTradeCalc shape — board changed 1 day ago, picks frozen 38 days against a 14-day budget — and a whole-file hash cannot see it, because the file *does* change.

### Tests (5, synthetic zips in `tmp_path`)
- a source changing every archive is **not** content-stale — positive control; without it a detector reporting everything as frozen would pass the fail-closed test and prove nothing
- **fetch-fresh but content-frozen IS reported stale**, in days — the fail-closed core
- pick rows measured separately from the board
- a single observation is **unknown, not fresh** (missing is never zero, applied to a change interval)
- self-check that board age and pick age are genuinely different quantities

### Mutations — each RED on exactly the intended assertion, then restored to an empty diff
| Mutation | Result |
|---|---|
| reuse the whole-file digest for picks | **2 RED** (`assert 0 != 0`) |
| single observation returns `0.0` instead of `None` | **1 RED** (`assert 0.0 is None`) |
| treat every archive as a fresh publish | **1 RED** (`assert 0 == 35`) |

## Scope

**Test-only — no production code touched** (`git status` shows one added file), so no source voting availability change and **no board effect** by construction. No bridge math, Hill curve, bridge weight, IDP ceiling, family voting, or backup bridge ordering touched. No credential, session material, or vendor payload committed; the vendor deployment id was redacted from all output.

## Verification
- `tests/sources/test_content_staleness.py` — 5/5
- `tests/sources/` + freshness suites (`test_per_source_freshness`, `test_watchdog_freshness_soft`, `test_freshness_budget_not_laxer_than_alerts`, `test_watchdog_unmeasurable_source`, `test_config_parity`) — **198 passed**
- `ruff check` + `ruff format --check` — clean
- `scripts/check_decision_coercions.py` — clean

## Known, NOT fixed here (Claude 5's surface)

Content age never reaches any production surface. `dataFreshness.sourceTimestamps.idpTradeCalc.staleness` reads `"fresh"`, the source-health dot is green, and per-row `confidenceReasons` publishes *"All contributing evidence is current"* for pick rows priced by a 41-day-frozen feed. That claim is **true today** — I verified live parity out-of-band — but **the system cannot itself substantiate it**: it has no way to distinguish "vendor republished identical values" from "vendor went dark".

This is already logged as **S-3 OPEN** in `docs/sources/SOURCE_CENSUS_2026-08-18.md`. Closing it means adding health vocabulary to the contract, which is a contract-surface decision. This PR protects the detector that would feed it.

Also inert, worth a one-line correction whenever someone is next in the file: `tests/api/test_config_parity.py::MTIME_RELIABLE_SOURCES` lists `idpTradeCalc` as a source "whose CSV content provably changes every fetch" — falsified by measurement, but harmless, because that set is an OR-alternative for sources with *no* stamp and `idpTradeCalc` has one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XQ3FG1nedWBsekB1HkhQk2)_


jasonleetucker-code:
## CI red — both failures are PREEXISTING_UNRELATED, reproduced on pristine `main`

`Validate PR` failed on head `2c04184b8`. Neither failure is this PR's. This PR adds **one test file and touches no production code**, and its 5 tests are inside the 7,478 that passed.

I reproduced both against a clean worktree at `origin/main@13456dd5f` with my file **absent**:

**1. Hard gate** — `tests/roster_intel/test_metric_separation.py::test_metrics_genuinely_diverge_on_a_representative_real_board`
```
AssertionError: the single strongest-by-value team should not also rank
identically on youth by coincidence: (1, 1)
assert 1 != 1
```
Identical on pristine `main`. Not source-acquisition — routing, not fixing.

Worth flagging separately: this is a **live-board-coupled assertion sitting in the hard gate** (`-m "not livedata"`), and it fails whenever the strongest-by-value team happens to also be the youngest. `CLAUDE.md`'s stabilization rule says a hard-gate test "must not assert an absolute count, a floor or a health status over the LIVE board" — a coincidence-avoidance property over the live board is the same hazard, and because the hard gate runs `-x`, this one aborts the suite for **every** PR until the board moves. That is a governance/ownership call for whoever owns `roster_intel`, not something I should paper over.

**2. livedata** — `tests/canonical/test_ktc_reconciliation.py::test_rank_delta_within_tolerance[12-…]`
```
Divergence from KTC at rank 12 drifted: pinned +12.3% vs actual +15.4%
(KTC=7420, ours=8561, band ±3.0pp)   assert 3.077358490566038 <= 3.0
```
Identical on pristine `main`; over the ±3.0pp band by 0.077pp. This is the KTC Hill-curve calibration drift already routed to Claude 5 — **explicitly outside this lane** ("do not modify Hill curves"). Fixing it here would be exactly the methodology change the lane is forbidden.

**3. Source-health step** — exits 1 on `idpTradeCalc (picks) … 40 days (budget 14d)`, but it is `continue-on-error: true` in both workflows that run it, so it did **not** fail the job. It is the standing condition this PR exists to protect the detector for, and it is reported exactly as designed.

### Control
Copied my test file alone onto pristine `main` — **5/5 pass**, nothing else changed. So the file is clean and the two blockers are independent of it.

### Why I am not pushing a fix
Drive-to-green says fix or state the blocker. Both blockers are outside source acquisition, and one is in territory this lane is explicitly barred from. Pushing a speculative fix into Hill-curve calibration or `roster_intel` to turn my own PR green would be the wrong trade. **This PR is mergeable on its own merits once either blocker clears** — it changes no production code, so it cannot affect them.

---

jasonleetucker-code:
## Re-run update: blockers narrowed 2 → 1, and the flake diagnosis is confirmed

Same head `2c04184b8`, re-run (`check_run_id` 97674041313).

**Blocker 1 — CLEARED on its own.** The hard gate now passes; `test_metrics_genuinely_diverge_on_a_representative_real_board` no longer fails. Nothing in this PR changed — the live board moved and the `(1, 1)` coincidence dissolved. That is direct confirmation of the diagnosis in my previous comment: it is a **live-board-coupled assertion in the hard gate**, red or green depending on where the board happens to sit, not a code defect.

I also see it is already being addressed on `main` — the release-candidate step reports it as the one class-B change outstanding on base:
```
[release-candidate][B] tests/roster_intel/test_metric_separation.py  (source/governance)
```

**Blocker 2 — UNCHANGED, and now the sole blocker.**
```
test_ktc_reconciliation.py::test_rank_delta_within_tolerance[12-8561-12.3-3.0]
pinned +12.3% vs actual +15.4% (KTC=7420, ours=8561, band ±3.0pp)
assert 3.077358490566038 <= 3.0
```
Identical numbers across both runs and on pristine `main` — a stable, real calibration drift, over the band by 0.077pp. **Hill-curve calibration is explicitly barred from this lane**, so I am not touching it; it stays routed to Claude 5.

**Source-health step** unchanged and behaving correctly: exit 1 on `idpTradeCalc (picks) … 40 days (budget 14d)`, `continue-on-error: true`, does not fail the job.

**Not rebasing.** 79 class-C + 1 class-B outstanding on base, advisory by design. Per `CLAUDE.md`'s HEAD-FREEZE rule class-C drift is never a reason to rebase, and the one class-B change is to a test that already passes on this head — so pulling it in is not required for green and would only re-open a validated head.

This PR remains **test-only, zero production code**, and cannot affect the remaining blocker.

---

PR #1082, opened 2026-08-24
V1-45: L4 consumption audit + evidence recipe (docs only)

jasonleetucker-code:
Docs only — **no code changed**. Starting SHA `131abf9f`, head `e6cf26cc`.

`V1-45` (Trade calculator) is `IMPLEMENTED_UNVERIFIED` at target **L4**, whose clause is *proof the intended surface consumes the canonical implementation truthfully*. L3 (deployed-SHA) needs deploy access this lane lacks — but the L4 clause is partly a **code** question, and that half was measurable from here. It was measured, and it fails.

## Measured

`simulate_trade` stamps exactly three optional canonical blocks. Repo-wide frontend consumer counts (excl. `node_modules`):

| block | owner row | consumers | rendered |
|---|---|---|---|
| `rosterCapacity` | V1-39 | 2 | yes |
| `teamImpact` | C2-STR-01 | 1 | yes |
| `finalRosterSimulation` | **V1-42** | **0** | **no** |

Separately: `src/trade/analyze_trade.py` and `POST /api/trade/analyze` are both on `main`, with **zero** frontend callers and no Next bridge route. The frontend calls `simulate`/`finder`/`import-ktc`/`suggestions`/`simulate-mc`/`export-ktc` — never `analyze`.

Reproduction is ~5 seconds, read-only, no network — commands are in the doc.

## Attribution — stated carefully, because the tempting conclusion is wrong

- **V1-42 is NOT reopened.** Declared target **L2**. Engine + endpoint wiring hold and are test-pinned. Frontend rendering is an L4 clause, *above* its bar. Stays VERIFIED, frozen.
- **V1-43 is NOT reopened.** Declared target **L1**. Module + endpoint exist and are tested. Reaching a user is an L4 clause, *above* its bar. Stays VERIFIED, frozen.
- **V1-97 untouched** — already VERIFIED; broader replay is POST-V1.

Both gaps belong to **V1-45** — the only row in this group whose target reaches L4, and therefore the only one whose bar the unrendered blocks actually fail. It is the *same gap class already repaired once on this row*: `rosterCapacity` had zero consumers until V1-45's earlier pass wired it (merged in #1025).

## Repair scope — written, deliberately NOT implemented

This dispatch does not own frontend, so scope is handed over rather than executed:

[...]

PR #1120, opened 2026-08-25
V1-45 — render the canonical finalRosterSimulation on the Trade Calculator

jasonleetucker-code:
Lane 6 (Premium UI / Frontend). Frontend lane released by Integration after #1086 merged; V1-45 is the assigned next target.

**No V1 status column edited. V1-45 is NOT promoted — its bar is L4 and this environment reaches L2.**

## The gap, measured rather than inferred

`POST /api/trade/simulate` has published `finalRosterSimulation` since **V1-42** (`VERIFIED`): the lineup **re-solved** over the post-trade, post-cleanup roster, with Team Strength before/after. On `main` `b857d07c0` it appears **4 times** in `src/api/trade_simulator.py` and **0 times** anywhere in `frontend/`. I fetched `trade-sections.jsx` from `main` directly to confirm rather than trusting the local tree. The backend answered a question the UI never asked.

## It is not a second copy of `rosterCapacity`

| block | answers |
|---|---|
| `rosterCapacity` (V1-45's earlier pass, #1025) | **who must go** — forced releases, cost, upper-bound flag |
| `finalRosterSimulation` (this PR) | **what the roster becomes once they have** — promotions into vacated seats, both need directions, the Team Strength delta |

Releasing the cheapest legal body can vacate a FLEX seat and promote a player the trade never mentioned. That is the fact nothing on the page carried.

Forced releases appear here by **count only**, pointing at the capacity block — one fact, one surface.

## Four states, not three

The row named three. The code emits four, and the fourth is load-bearing.

| state | shape | renders |
|---|---|---|
| **populated** | `available: true` + the full `RosterSimulation` dict | the "Final roster" section |
| **capacity_uncertain** | `{available: false, unavailableReason: "capacity_uncertain"}` | *"…taxi occupancy is unknown… see Roster capacity above"* |
| **unavailable** | **two** shapes — the above with `starter_slots_unresolved`, **and** `{unavailable: "<ExcType>"}`, which carries **no `available` key at all** | *"Not simulated — …"* naming the reason |
| **absent** | key not on the response (no resolved team) | **nothing at all** |

Absent is the ordinary state of an un-teamed what-if; an error banner there would be the defect.

### One deliberate divergence from the backend's own predicate

The gate is `available === true` — stricter than `analyze_trade.py::_roster_impact_dimension`, which uses `.get("available", True)` and so defaults a **missing** key to available. On the exception shape that predicate falls through and only fails later on a missing `strengthBefore`, reporting `team_strength_not_stamped` rather than the real exception. Same end state, imprecise reason. **Observed and recorded, not fixed here** — different file, different already-`VERIFIED` row, unchanged outcome. The frontend fails closed instead.

## Zero recomputation

Every number is `Math.round` / `toLocaleString` / `fmtSigned` over a backend stamp. The Strength change reads the stamped **`strengthDelta`**, never `after − before`. This file's own header says *"NO trade math lives in this file"*; that stays true.

Both need directions render, per `RosterSimulation`'s own docstring: *"A simulation that only showed what improved would be an advocacy tool."*

## Mutation proof — three, each on a different load-bearing property

| mutation | result |
|---|---|
| delete the render block (literal pre-fix state) | **14 of 15 RED** |
| recompute the delta client-side instead of reading `strengthDelta` | **1 RED** — the fixture stamps `before 100`, `after 500`, `strengthDelta 7`, so only a *reader* passes |
| `available === true` → `available !== false` | **1 RED** — the `{unavailable: "ValueError"}` shape |

All restored GREEN. The middle one is the zero-recomputation guard, and that fixture is self-inconsistent on purpose.

## Gates

| gate | result |
|---|---|
| frontend Vitest | **2,378 / 2,378** across 161 files |
| both trade-panel test files | **22 / 22** (the sibling `rosterCapacity` suite unaffected) |
| `tests/api/test_trade_simulator.py` + `tests/trade/` + `tests/roster_intel/` | **1,423 passed** — and **no `.py` file is in the diff** |
| production build + bundle budgets | green, **all 14 under**; `/trade` 62.9 → 65.5 KB against a 92 KB budget |
| planning integrity · decision coercions | clean |

This repo has no frontend lint script, so none was run — stating that rather than implying a gate that does not exist.

## `trade-sections.jsx` overlap, adjudicated

`check_work_claims.py` names three branches on this file. All three are **stale**: every line each adds is already on `main` (checked line-by-line — **0** novel lines from any of them). `--is-ancestor` reports them unmerged only because they were squash-merged, which mints a new SHA. No live competitor on `SimulationPanel`. I ran this check after writing the code; the repo rule says before.

## Boundaries

No backend change, no new endpoint, no trade math, no change to the existing `rosterCapacity` render, no V1 status column edited, no promotion. L4 recipe — including the field-for-field comparison against the raw `/api/trade/simulate` response — in `docs/master-site-audit/evidence/V1-45/L4_PRODUCTION_RECIPE.md`.

---


jasonleetucker-code:
## CI blocker at `1ba7b787a` — base-caused, not this PR

`Validate PR` failed at step 17, **API data contract check (structural lane — blocking)**:

```
[contract] ok=False errors=1 warnings=0 players=1116
[contract] structuralErrors=1 sourceHealthErrors=0
[contract][error] blend_integrity_violation:1 row(s) hold a value outside the range
                  of their own source contributions (2027 Mid 2nd)
```

### Measured, not asserted

| check | result |
|---|---|
| files this PR touches | `trade-sections.jsx`, its new test, `WORK_CLAIMS.md`, the V1-45 recipe — **no `.py`, no `exports/`, no `data/`, no `config/`, no `CSVs/`** |
| the payload the gate validates (`exports/latest/dynasty_data_2026-08-25.json`) | blob `2695ea842` on **both** `HEAD` and `origin/main` — **byte-identical**, untouched by this PR |
| same gate run locally on this branch | reproduces exactly: `structuralErrors=1`, same `2027 Mid 2nd` row |
| anything the validator reads that this diff changed | **none** |

A frontend render change cannot move a blend hull. The gate's input is `main`'s data and its code is `main`'s code.

### It is working as designed

Per `CLAUDE.md`, a blend-integrity violation is a row whose blended value fell outside the range of its own source contributions — structurally impossible under correct operation. The system deliberately **does not coerce it**: the row is quarantined and the build-level validator raises a hard error, because *"coercing an impossible number to a plausible one hides a pipeline fault."* That is exactly what happened. **Nothing here was skipped, relaxed, or allowlisted.**

This is the same defect class that blocked #1086's E2E preflight (then `2028 Late 3rd`). `origin/codex/fix-pick-blend-integrity-2027-late-1st` is already open against it, so it has an owner — it is not this lane's row and not this lane's fix.

### What DID pass

The **hard gate passed** (step 15, 16:57:29 → 17:12:24), including `tests/roster_intel/test_metric_separation.py::test_metrics_genuinely_diverge_on_a_representative_real_board` — the data-coupled assertion that reddened #1086. It passes here on a newer base, which is consistent with the measured diagnosis that it is a function of which archive bundle the merge ref selects, not of any code. Live-data tier, release-candidate freshness and source-health (`contract source-health errors: 0`) all green too.

**Note on coverage:** `Frontend unit tests (vitest)` and `Frontend build + bundle-size budget` carry no `if: always()` in `pr-validation.yml`, so step 17's failure ends the job before this lane's own gates run. They are verified locally at this exact head — Vitest **2,378 / 2,378** across 161 files, production build green with **all 14 bundle budgets under** (`/trade` 62.9 → 65.5 KB against 92 KB) — and CI will exercise them as soon as the contract blocker clears.

**Implementation head: GREEN. Release candidate: BLOCKED on a `main` pipeline defect with an existing owner.**

---

jasonleetucker-code:
## Blocker cleared — and a correction to my diagnosis above

**Correction first.** I wrote that the `blend_integrity_violation` was a genuine pipeline fault and that the gate was "working as designed" catching an impossible value. **That was wrong.** #1122 shows the *detector* was at fault, not the board: the hull check compared two integers quantized by different rules — `rankDerivedValue` is `int()` (truncation, always ≤ the float blend) while each `valueContribution` stamp is `int(round())` (±0.5 either way). For `2027 Mid 2nd` the votes were 3459.692 and 3460.0, the float blend ~3459.85 sat **inside** the true hull, but published as `3459` against stamps of `[3460, 3460]` — flagged as below a hull the blend never left. A **false positive**, not an impossible value.

My *classification* (base-caused, not #1120) and my *action* (report it, weaken nothing) both hold. My characterisation of what it was does not, and it would have misled anyone reading this thread later.

**Verified cleared on this exact tree:**
```
[contract] ok=True errors=0 warnings=0 players=1116
[contract] structuralErrors=0 sourceHealthErrors=0
```

## Independent corroboration of this PR's premise

`#1082` ("V1-45: L4 consumption audit + evidence recipe") landed on `main` while this was in flight, and the Trade Intelligence lane measured **the same gap independently**:

| canonical block | frontend consumers | rendered on `/trade`? |
|---|---|---|
| `rosterCapacity` | 2 | yes |
| `teamImpact` | 1 | yes |
| **`finalRosterSimulation`** | **0** | **no** |

Same conclusion from a different lane against `main` @ `131abf9f`; this PR measured it against `b857d07c0`. **#1082 is the audit; this is the repair.**

## One thing that needed fixing as a result

Two L4 recipes for one row is the drift `ONE CONCEPT, ONE CANONICAL OWNER` forbids. `7c2b71c89` scopes this PR's recipe explicitly to the single repair and **defers to `docs/trade/V1_45_TRADE_CALCULATOR_L4_EVIDENCE_RECIPE.md` as the canonical V1-45 L4 recipe**. Nothing deleted, no second owner created.

Also recorded, deliberately **not** touched: #1082 finds `POST /api/trade/analyze` (V1-43) has zero frontend callers and no Next bridge route. Different row, different owner.

Head is now `7c2b71c89` (Integration's `main` merges + one docs commit). The V1-45 render and its 15 tests survive the rebase intact.

---

PR #1130, opened 2026-08-25
V1 verification: production-auth Playwright specs (V1-27/45/56/111/131 browser halves)

jasonleetucker-code:
# Production-auth verification specs

Adds `tests/e2e/specs/prod-auth/` — five Playwright specs that run against the **deployed production site** with a real session cookie, producing the browser half of the evidence five V1 rows explicitly could not get from offline instruments. Test code only: no workflow files, no auth/minting logic, no server changes.

## What each spec proves (and what it only annotates)

| spec | row | asserts | annotates honestly |
|---|---|---|---|
| `v1-131-nav-gating.spec.js` | V1-131 | `features.consensusEdge.available` is a real boolean AND agrees with the board endpoint (200↔true, 503↔false — the recipe's non-waivable invariant); desktop Market menu, mobile drawer (390×844), command palette, `/more`, and DOM-wide `a[href="/consensus-edge"]` all honour it; negative control that Source Disagreement / Sharp Tracker / Sharp Roster Percentage still render; hard reload of `/rankings` makes **zero** `/api/consensus-edge/*` requests and **exactly one** `/api/auth/status`; direct `/consensus-edge` navigation still renders its own `<h1>` | which branch ran (`branch: available=<bool>`), deployed SHA, menu/palette/drawer contents — both branches are implemented, only the deployed one is exercised |
| `v1-45-trade-surface.spec.js` | V1-45 | drives the real `/trade` UI (team selected via the page's own selector, players added via its search, receiving roster chosen **at the cap** when one exists), captures the page's **own** `POST /api/trade/simulate`, then asserts the rendered Final-roster section equals the response's `finalRosterSimulation` stamps field-for-field under the component's exact formatters (strength before/after/change incl. the U+2212 sign convention, promotions/displacements by count and name, needs closed/opened, `cleanupApplied` forced-release count) plus the `rosterCapacity` banner's release count and named drops; the observed state must not carry the other states' copy (`capacity_uncertain` "taxi occupancy" wording vs populated vs "Not simulated — <reason>") | `states-observed` records exactly which backend states the live league produced — unobserved states are never faked; trade composition and at-cap status recorded |
| `v1-111-premium-rankings.spec.js` | V1-111 | `.psi-editorial` scope present **and active** (its `--surface-0` token remap measurably applied, not just a class string); top row is rank #1 with the backend's rank-1 player name and `rankDerivedValue` under the page's own formatting; a 15-row sample of the visible window matches backend (rank → name, value) exactly — a client-side re-rank or re-value fails with a precise diff; buildRows' zero-rank-stamps fail-fast never fired; runs on both 1366×768 and 390×844 | partial-data states: any visible `—` rank is verified truthful against the contract; when the sampled window exhibits none, that is annotated rather than claimed |
| `v1-56-waivers-faab-strip.spec.js` | V1-56 | the league-FAAB strip's "League average bid" / "League median bid" tiles equal the payload from the page's **own** `faabAnalytics` fetch (formatting-aware `$${Math.round(v)}`), replicating the component's evidence gate so a **measured** $0 median must render as `$0` while a zero with no observations must render `—`; absent analytics ⇒ explicit `—` state or absent strip, and never an invented `$0` | which state the live payload produced (`state-observed`), the analytics request/status observed |
| `v1-27-lineup-render.spec.js` | V1-27 / C2-U1 §10 item 2 | terminal Portfolio starters list equals `sleeper.teams[].optimalLineup` for the displayed team — slot **sequence** identical to the stamp's assignment order, player set equal (normalized; loose display-name joins annotated, real mismatches fail with a diff); split-legend starter count equals `stamp.starters.length`; bench count bounded so stamped-unpriced players cannot be silently benched, and no stamped-unpriced player renders as a starter; truth-ladder note absent iff `slotSource === "sleeper_roster_positions"`; `/rosters` "Starters only" renders all team rows with the `starter-slots-unavailable` note appearing exactly when a stamp is missing | stamp census (available/slotSource per team), name-join quality, and — when the live board has `unpriced: []` — that the positive exclusion case was not exercisable and is not faked |

## Read-only guarantee

Navigation, DOM reads, network observation. The single POST is `/api/trade/simulate` (pure computation). Where driving the UI would write user preferences (team switcher → `PUT /api/user/state`), the spec intercepts the write and answers it locally, so the real user's saved state is never touched. The cookie value is read from the file, attached to the browser context, and never logged or embedded in any assertion message.

## Isolation from the default suite

* New dedicated config **`tests/e2e/prod-auth.config.js`** (projects `prod-desktop` 1366×768 + `prod-mobile` 390×844; no webServer, no global-setup, `retries: 0` so a production finding is never retried away; separate artifact dirs, gitignored).
* `tests/e2e/playwright.config.js` gains `testIgnore: "**/prod-auth/**"`.

Verified:

```
# default config — unchanged before and after this PR
npx playwright test -c tests/e2e/playwright.config.js --list
Total: 516 tests in 22 files            # identical pre/post

# dedicated config
npx playwright test --config tests/e2e/prod-auth.config.js --list
Total: 20 tests in 5 files              # 5 specs × 2 projects

# run WITHOUT credentials (this sandbox has none) — clean skip, exit 0
npx playwright test --config tests/e2e/prod-auth.config.js
  20 skipped
```

`tests/e2e/test_e2e_harness_guards.py`: 14 passed (the new files satisfy the bare-goto and retired-name guards).

## Env contract (supplied by the CI workflow, not this PR)

| var | meaning |
|---|---|
| `PROD_ORIGIN` | e.g. `https://chaseupside.com` |
| `PROD_SESSION_COOKIE_FILE` | file containing ONLY the `jason_session` cookie value |

Unset ⇒ every spec skips with an explicit message. Set but non-authenticating ⇒ loud failure (credentials were supplied and do not work), never a skip.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,10 @@ frontend/.next.old/
 frontend/.next.*/
 tests/e2e/playwright-report/
 tests/e2e/test-results/
+# prod-auth verification suite (tests/e2e/prod-auth.config.js) writes its
+# own artifact dirs so a prod run never clobbers the default suite's.
+tests/e2e/playwright-report-prod-auth/
+tests/e2e/test-results-prod-auth/
 tests/e2e/tests/
 test-results/
 
--- a/tests/e2e/playwright.config.js
+++ b/tests/e2e/playwright.config.js
@@ -103,6 +103,13 @@ const backendEnv = {
 
 module.exports = defineConfig({
   testDir: "./specs",
+  // specs/prod-auth/ is the production-auth verification suite. It runs
+  // against the DEPLOYED site with a real session cookie, exclusively
+  // through tests/e2e/prod-auth.config.js — collecting it here would
+  // add permanently-skipped tests to every local/CI run (its fixture
+  // skips without PROD_ORIGIN + PROD_SESSION_COOKIE_FILE) and shift the
+  // suite-size expectations the reporter and workflows are tuned to.
+  testIgnore: "**/prod-auth/**",
   // Verifies the stack can actually serve the suite (snapshot loaded,
   // test sessions unlocked, authenticated /api/data populated) and
   // fails with a fix-it message instead of letting specs time out or
--- a/tests/e2e/prod-auth.config.js
+++ b/tests/e2e/prod-auth.config.js
@@ -1,0 +1,122 @@
+const fs = require("node:fs");
+const path = require("node:path");
+const { defineConfig } = require("@playwright/test");
+
+/**
+ * Playwright config for the PRODUCTION-AUTH verification specs
+ * (tests/e2e/specs/prod-auth/).
+ *
+ * A deliberately separate config from playwright.config.js, for three
+ * reasons:
+ *
+ *   1. **No stack.** These specs run against the deployed production
+ *      site — no webServer, no global-setup contract priming, no
+ *      test-session minting. Booting the local stack for them would be
+ *      wrong twice over (it is not the thing being verified, and the
+ *      sandbox has no production credentials).
+ *   2. **No collection overlap.** The default config's testDir is
+ *      ./specs, which contains this directory; playwright.config.js
+ *      carries a matching `testIgnore` so the default suite's size is
+ *      unchanged. This config is the ONLY way these specs are collected.
+ *   3. **No retries.** A production verification run must not launder a
+ *      flake into a pass — a spec that fails once against prod is a
+ *      finding to read, not to retry away.
+ *
+ * Runs only from the production-verification CI workflow, which supplies:
+ *   PROD_ORIGIN               e.g. https://chaseupside.com
+ *   PROD_SESSION_COOKIE_FILE  file containing ONLY the jason_session
+ *                             cookie VALUE
+ * Without both, every spec skips with an explicit message (see
+ * specs/prod-auth/helpers.js) — so an accidental local run is loudly
+ * inert, never red and never green-by-vacuity (the skip count says so).
+ */
+
+// Same pre-installed-Chromium fallback as playwright.config.js — needed
+// so `--list` (and dry runs) work in sandboxed agent containers whose
+// browser revision doesn't match this @playwright/test version. Kept as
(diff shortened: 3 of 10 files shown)


jasonleetucker-code:
Production-verifier correction from run 32919648642: V1-56's red is currently NOT valid product evidence. The browser spec reaches /waivers, captures a real faabAnalytics payload (mean 1.37 over 1,097 bids), but never selects a team; the failure screenshot is still in the `Select a team below` state. Current `ManualAddDrop.jsx` renders `<FaabHeaderStat ...>` only in the `selectedTeam` branch, so looking for `League average bid` before team selection is vacuous. The API half has a separate envelope bug: `check_v56()` checks `body.faabAnalytics` or the top-level body, but this endpoint is `{contractVersion, league, section, data}` and current `ManualAddDrop` correctly consumes `data?.data`; that is why V56A reported keys `contractVersion,data,league,section` and falsely said `leagueMedianWinningBid` was absent. Repair verification only: select a real team before asserting the strip, and unwrap `body.data` in V56A. Preserve measured $0 vs missing semantics. Do not change FAAB product methodology based on these two false reds; rerun V1-56 only after the verifier fix.

jasonleetucker-code:
Second production-verifier correction from run 32919648642: V1-45's browser red is also not valid product evidence yet. Current `/trade` still contains the `Simulate impact` button, but `page.jsx` renders it only when the canonical global `useTeam().selectedTeam` is set. The failure screenshot shows the shell still at `Pick your team`; the spec selected the separate Suggestions Desk `Your team` combobox (Brent), which does not establish `useTeam().selectedTeam`. It then waited for a conditionally absent button and declared the consumer missing. API check V45A independently PASSED with a populated canonical `finalRosterSimulation`. Repair verification only: select a real team through the same global team-selection path `useTeam()` consumes, then drive the existing Simulate impact control and compare the rendered `SimulationPanel` to that page-owned `/api/trade/simulate` response. Do not change trade simulation or frontend methodology based on this false red; rerun V1-45 only after the harness setup is corrected.

PR #1132, opened 2026-08-25
V1-76: one canonical failure-attribution vocabulary

jasonleetucker-code:
## V1-76 — Failure attribution compares one vocabulary (audit F-12 / census S-2)

Required level **L2** (RED→GREEN deterministic test + measured board/contract effect). The previously proposed `playerCount: null` design is **refuted** (`docs/VERSION_1_COMPLETION_CONTRACT.md` §8.1) and is **not** resurrected here.

### The real defect (F-12)
The source-health surface compared scraper **run names** (`KTC`, `IDPTradeCalc`, `DLF_LocalCSV`) against ranking-**registry keys** (`ktcSfTep`, `idpTradeCalc`, `dlfSf`). The two vocabularies are disjoint on real data (`"DLF_LocalCSV" != "dlfSf"`), so a run-level failure disposition could not round-trip to the registry rows it concerns — a hard-failed source was misattributed or dropped and rendered green off its stale-but-recent CSV. `frontend/components/SourceHealthStrip.jsx` already documented that its per-row run-state lookups were structurally dead and explicitly punted the mapping to the backend seam (`server._source_health`).

### The fix — one DECLARED mapping, owned by the registry
- `_RANKING_SOURCES` entries gain a `run_source` field, declared **only** where the in-process Dynasty Scraper run governs the source — the only run names that can reach `sourceRunSummary`: `KTC → ktcSfTep`, `IDPTradeCalc → idpTradeCalc`, `DLF_LocalCSV →` the four DLF boards it loads (aligns with V1-80 / F-17, #1126). Keys fetched by their own `scripts/` timer omit it and **fail closed**. Identity metadata only — decides no value.
- `registry_keys_for_run_source()` resolves a run name to its registry keys, **reusing V1-80's `critical_primary_for_run_source` qualifier rule** (exact, or `<primary>_<qualifier>`) rather than inventing a second table.
- `_build_source_health_snapshot` stamps `registryKeys` on each failure and publishes `failed_source_keys` / `partial_source_keys` / `timed_out_source_keys` beside the run-name lists — additive, provenance retained.
- `SourceHealthStrip` consumes the resolved keys: a failure that round-trips to a rendered row lights it up; one that resolves to nothing stays shown verbatim as unattributable.

### Refuted-design avoidance
No `sites[].playerCount` is nulled and the scrape-promoter `site_count = len([s for s in result.get("sites", []) if s.get("playerCount", 0) > 0])` expression is guarded by a dedicated test. The change is additive to `/api/status` only; it never reaches the promotion path.

### L2 measured effect
`exports/latest/dynasty_data_2026-08-25.json` via `build_api_data_contract`: **1116 rows built with and without the change — 0 rows moved value, rank, or tier.** Attribution is a health-surface concern, not a valuation one.

### Tests
- **RED→GREEN**: `tests/api/test_failure_attribution_vocabulary.py` (11) fails to import `registry_keys_for_run_source` at HEAD; green after. Covers the run-name→keys round-trip, fail-closed on unverified names, the snapshot round-trip, and the two refuted-design guards.
- 3 new frontend cases pin row-lighting on the resolved shape and the fail-closed unattributable panel; the 13 existing F-12 cases stay green.
- `tests/api/` (2485 passed), `tests/deploy/` (376 passed), source-registry parity, contract-health lanes, DLF critical gate all green. `check_planning_integrity` + `check_decision_coercions` clean.

---

### DIFF
--- a/frontend/__tests__/components/source-health-strip.test.jsx
+++ b/frontend/__tests__/components/source-health-strip.test.jsx
@@ -349,3 +349,120 @@ describe("SourceHealthStrip · run failures it cannot attribute (F-12)", () => {
     expect(document.body.textContent).toMatch(/matched by key/);
   });
 });
+
+// The server now speaks ONE vocabulary (F-12 / census S-2 / V1-76).
+//
+// ``server._source_health`` resolves each run-named failure to the
+// registry keys it concerns via its single
+// ``registry_keys_for_run_source`` owner, and publishes them as
+// ``source_failures[].registryKeys`` plus the ``*_source_keys``
+// registry-key projections of the run-state lists.  The frontend needed
+// no mapping of its own — it consumes the resolved keys.  A failure that
+// round-trips to a rendered row now LIGHTS THAT ROW (not the
+// unattributable panel); a failure that resolves to nothing (an
+// unverified run name, fail-closed) is still shown verbatim as
+// unattributable.
+describe("SourceHealthStrip · server-resolved attribution (F-12 / V1-76)", () => {
+  /** The resolved shape: run-named failures carry their registry keys,
+      and run-state lists carry a registry-key projection. */
+  const RESOLVED = {
+    source_health: {
+      source_runtime: {
+        overall_status: "partial",
+        enabled_sources: ["KTC", "IDPTradeCalc"],
+        failed_sources: ["KTC"],
+        failed_source_keys: ["ktcSfTep"],
+        partial_sources: ["DLF_LocalCSV"],
+        partial_source_keys: ["dlfSf"],
+        timed_out_sources: [],
+        timed_out_source_keys: [],
+        finished_at: new Date().toISOString(),
+      },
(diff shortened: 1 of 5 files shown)


jasonleetucker-code:
Traffic-control note: this merged unit appears ledger-closeable as V1-76 at its declared L2 bar. Exact feature head `bd3127a1bb7aea534243c319cdbc9d12eac130ce` has PR Validation SUCCESS; the PR records RED→GREEN/non-vacuous attribution tests plus the measured 1,116-row board effect (0 value/rank/tier moves). Current `main` is a descendant via merge `7e73a97b4d494d1cda311dcd92dbebe5b8624058`; later movement is ops/data refresh, not a conflicting product change. The canonical contract still says V1-76 `NOT STARTED`. Integration should reconcile that stale status; do not promote from merge alone if any shipping-tree check contradicts the recorded L2 evidence.

PR #1134, opened 2026-08-25
V1-59: commit per league in the sharp transactions crawl (writer-lock window)

jasonleetucker-code:
## The production defect (found by the V1 on-box checklist harvest, 2026-08-25)

The scheduled sharp chain failed with `sqlite3.OperationalError: database is locked`: `dynasty-sharp-discovery` (ExecMainStatus=1, 06:47), `dynasty-sharp-rosters` (ExecMainStatus=1, 11:28 + Aug 24 13:13), and three lock lines in `dynasty-sharp-records`' journal — six lock events across three units in two days.

## Root cause

Not a missing-PRAGMA problem: every path already routes through the single connection factory `src/intel/ledger.py::connect` (30 s `busy_timeout`, WAL persistent in the production file). The cause is `src/sharp/transactions.py::crawl_transactions` holding one **implicit write transaction across its entire budgeted network crawl**: the first `_save_fetch_state` upsert opens it, and the only commit was the single end-of-crawl `conn.commit()`. In a quiet period most leagues yield zero new events, so `ledger.ingest_events` (which commits internally) never fires and the WAL writer lock is held for the length of the run — minutes up to `TimeoutStartSec=1800` — while every other sharp unit waits out its 30 s `busy_timeout` and raises. Timer overlap is routine, not hypothetical: `Persistent=true` catch-ups and `OnActiveSec=30min` re-fires produce runs far outside the staggered slots (journal: 00:01 / 07:02 / 11:28 / 13:13 / 16:23 / 22:27).

This is the exact defect class already found, fixed and test-pinned for the records crawl (`records.py` per-season commit, `tests/sharp/test_records.py::TestWriterLockWindows`) — the transactions crawl never received the same repair.

## The fix

Commit per league in `crawl_transactions` — one `conn.commit()` on the failure path, one on the success path, mirroring `records.py`. Correctness is unchanged by construction: movements dedupe on `movement_id` (`INSERT OR IGNORE`) and fetch state upserts on `league_id`, so per-league durability changes no outcome — only the lock window. After this change no code path in the repo holds a ledger write transaction across network I/O.

Deliberately NOT done, with reasons: no retry-on-locked wrapper (contention-raises is pinned repo policy — `tests/intel/test_ledger_lock_safety.py`); no new PRAGMAs (the factory is already the one owner); no timer serialization (the journal proves the stagger already fails to serialize, and the write pattern must be safe under overlap).

## Evidence

- **RED→GREEN**: new `tests/sharp/test_transactions.py::TestWriterLockWindows` — a spying HTTP fake probes an independent connection when the crawl reaches the SECOND league and asserts the FIRST league's fetch-state row is already visible (non-vacuous: asserts the probe fired). Fails under the retired single-commit shape (re-verified by mutation), passes with the fix.
- `tests/sharp/` + `tests/intel/` — 607 passed, 0 failed.
- `ruff check` + `ruff format --check` clean.

---

### DIFF
--- a/src/sharp/transactions.py
+++ b/src/sharp/transactions.py
@@ -314,6 +314,7 @@ def crawl_transactions(
                 result.leagues_failed += 1
                 result.errors[lid] = "transactions_fetch_failed"
                 _save_fetch_state(conn, lid, state, error="transactions_fetch_failed")
+                conn.commit()
                 continue
 
             new_boundary = {
@@ -334,6 +335,18 @@ def crawl_transactions(
                 },
                 error=None,
             )
+            # Commit PER LEAGUE, not once for the whole crawl.  Every
+            # iteration does network I/O, so a single end-of-crawl commit
+            # held the SQLite writer lock across the entire budget — the
+            # same contention records.py repaired per season (V1-59): in a
+            # quiet period most leagues yield zero new events, nothing
+            # else commits, and the implicit transaction opened by the
+            # first _save_fetch_state upsert starves every concurrent
+            # sharp unit past its 30 s busy_timeout.  Movements dedupe on
+            # movement_id and fetch state upserts on league_id, so
+            # per-league durability changes no outcome — only the lock
+            # window.
+            conn.commit()
             result.leagues_fetched += 1
 
         conn.commit()
--- a/tests/sharp/test_transactions.py
+++ b/tests/sharp/test_transactions.py
@@ -298,9 +298,9 @@ class TestCoverageHonesty:
         stamps = _fetch_stamps(db)
         assert len(stamps) == 3, "the fixture must crawl three leagues or it proves nothing"
         assert cov["oldestCrawlMs"] == min(stamps)
-        assert cov["oldestCrawlMs"] != max(
-            stamps
-        ), "min and max coincide, so this fixture cannot tell them apart"
+        assert cov["oldestCrawlMs"] != max(stamps), (
+            "min and max coincide, so this fixture cannot tell them apart"
+        )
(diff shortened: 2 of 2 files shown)


jasonleetucker-code:
Traffic-control blocker: exact-head PR Validation on `7807608aab3580f030ae9c63eb8bb38461f328e6` is genuinely RED, not flaky. Run `32905769729` fails at the blocking `ruff format --check` gate, so the PR body's “ruff format --check clean” claim is stale at this exact head. Please make only the formatter-required change on this bounded branch, rerun exact-head CI, and freeze when green; do not broaden scope or chase unrelated main movement. After merge/deploy, V1-59 still requires the contract's L3 clean scheduled discovery→records→rosters production run before promotion.

jasonleetucker-code:
Confirmed and fixed at `e4ed4731e` — formatter-only change, nothing else touched. Root cause: this environment's ruff is 0.15.8 while the repo pins `ruff~=0.6.0`; the previous commit's test file was formatted with the newer version, whose output the pinned CI formatter rejects. Under `python -m ruff` 0.6.9 exactly that one file was flagged; after reformatting, repo-wide `format --check` reports 1346 files clean, `ruff check` clean, and the 22 sharp-transactions tests still pass (including the new writer-lock-window probe). #1135's branch was checked the same way and is already clean under the pinned version. Exact-head CI is re-running; freezing here until it reports.

Agreed on the promotion bar: merge/deploy of this fix is not V1-59 evidence — the row waits for the next clean SCHEDULED discovery→records→rosters chain in the production journal, which the on-box `v59` check will read.

---

PR #1146, opened 2026-08-26
FAAB bid predictor: learn manager tendencies from resolved losing bids

jasonleetucker-code:
## What changed
- keep completed winning claims as the canonical clearing-price distribution
- ingest resolved failed Sleeper waiver bids after processing as owner-tendency evidence
- never ingest pending claims, so the model does not imply blind bids are visible before waivers run
- prefer resolved-attempt aggression when a manager has at least 3 observations, with automatic fallback to legacy winning-bid history
- preserve schema-v1 history compatibility and the legacy `totalAdds` meaning
- report resolved-attempt / failed-bid coverage in the history refresh script
- add regression tests for pending-bid exclusion, failed-bid ingestion, clearing-price isolation, tendency preference, and v1 fallback

## Why
A losing $40 bid is strong evidence that an owner was willing to spend $40, but it is not evidence that the player cleared at $40. Separating those samples lets the existing FAAB engine make better opponent-bid predictions without corrupting its clearing-price priors.

## Safety / truthfulness
This uses only **resolved historical waiver transactions after the run**. It does not expose or claim access to other managers' pending blind bids.

### DIFF
--- a/frontend/components/waivers/FaabRecommendation.jsx
+++ b/frontend/components/waivers/FaabRecommendation.jsx
@@ -22,8 +22,8 @@
  *          the ceiling, because bidding the ceiling captures zero
  *          surplus by construction.
  *   2. Warnings ("likely outbid", pacing, marginal upgrade) as Banners.
- *   3. Contention — the FAAB v2 sealed-auction read: the clearing price,
- *      the projected top rival, and the per-opponent estimate table.
+ *   3. FAAB Bid Predictor — the sealed-auction read: clearing price,
+ *      projected top rival, and the per-opponent estimate table.
  *   4. Everything else behind disclosures: the factor breakdown (the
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
Traffic-control note: this PR is merged on `main`, but the exact PR head `0f1f3b7f4644c6e06e50e32e7a5954dbb79252eb` currently has **0 check runs / 0 commit statuses** in GitHub, and the merge commit `9ff1a394d12bec650bc76b861b7e845c9a8e6231` has no deployment workflow yet. Treat this as `MERGED_NOT_DEPLOYED`, not verified evidence for any V1 row. Before using #1146 for V1 closure, require the normal shipping-tree validation plus deployment and the row-specific production/consumer proof. No methodology objection is asserted here; this is an evidence-state guard only.

PR #1150, opened 2026-08-27
fix(v1): V1-131 auth-tristate bug, V1-56/61 verifier+perf, V1-45 team sync

jasonleetucker-code:
## Summary

Five independent, small, well-tested repairs surfaced by reconciling `docs/VERSION_1_COMPLETION_CONTRACT.md` against live GitHub state. None change methodology, valuation, or lineup invariants.

- **V1-131 (mobile drawer)** — PR #1149 (already deployed) fixed a pathname-close race, but the next authenticated production verification run still failed `v1-131-nav-gating.spec.js`'s mobile drawer check. Root-caused this session directly from that run's own Playwright trace (screencast frames + action log): `MobileTabBar` collapsed `useAuth`'s deliberately three-valued `authenticated` state (`true`/`false`/`null`) into a boolean, so a real signed-in session rendered the **public** tab set ("Sign in" included) for as long as the client-side `/api/auth/status` probe was in flight, then swapped to the real tab set the instant it resolved — and a click can land in that window. Desktop's `TopBar.jsx::visibleGroups` already had the correct three-way handling; `MobileChrome.jsx` did not. Fixed by mirroring that established pattern (new `MOBILE_TABS_UNKNOWN` neutral set), mutation-verified. **Not yet re-verified against production** — needs a fresh deploy + authenticated-verification run before the row can promote.
- **V1-56 (FAAB league context)** — backend and frontend were already correct; the *verifier script* (`scripts/verify_v1_authenticated.py::check_v56`) had the same envelope-unwrap bug as historical finding W11-F006. Fixed + regression-tested.
- **V1-61 (Sharp cohort timeout + false-green smoke gate)** — `cohort_status()` was calling `build_manager_records()` twice per request; `platform_records.coverage()` now accepts the already-computed population so the expensive joins run once. Also added a `SHARP_SMOKE_TOKEN` bearer-auth path (same pattern as `SIGNAL_ALERT_CRON_TOKEN`/`INTEL_REFRESH_TOKEN`) to `/api/sharp/cohort` and `/api/sharp/market` so the "Verify Sharp Production Population" workflow can actually measure instead of permanently reporting `unverifiable_unauthenticated`. This is a no-op until the secret is provisioned (documented in `.env.example`); both endpoints stay exactly as session-only as before until then.
- **V1-45 (/trade "Simulate impact")** — page already fully consumes the canonical `POST /api/trade/simulate` → `finalRosterSimulation` (no client-side computation). Real UX bug: the in-page "Your team" dropdown didn't propagate to the global `useTeam()` context gating the button. Fixed, mutation-verified.
- **V1-109 contract row** — corrected stale prose that contradicted the row's own status column (described only PR #1086's partial state; #1131 had already closed the remaining defects).

## Test plan

- [x] `python3 -m pytest tests/sharp/ tests/ops/ tests/api/test_private_auth.py -q` — 485 passed
- [x] `npx vitest run` (frontend) — 2383 passed
- [x] Mutation-verified each behavioral fix (temporarily reverted, confirmed the new test fails, restored) — V1-131, V1-45, V1-56, V1-61 self-authed auth
- [ ] Fresh production deploy + authenticated verification run (external — not run from this environment)

🤖 Generated with [Claude Code](https://claude.ai/code)

---


jasonleetucker-code:
Authenticated production verification attempt 6 (workflow run 32919648642, attempt 6, completed 2026-08-27T14:42Z) has now been harvested against the deployed #1150 application revision (Deploy Production run 33066619784 deployed b194316f0f1cd76e5eca4e0d73b20031a1d5f6cb successfully).

Row evidence:
- V1-45: PASS at both API and real production browser/L4. API V45A observed populated finalRosterSimulation. Playwright `v1-45-trade-surface.spec.js` passed: `/trade` made its own POST /api/trade/simulate and `SimulationPanel` matched the canonical response field-for-field. This is strong qualifying L4 evidence for contract promotion, subject only to landing the evidence/status update through normal contract convention.
- V1-56: NOT VERIFIED. The rerun checkout used the old verifier and therefore its API `leagueMedianWinningBid absent` result is stale/verifier-invalid after #1150. However the independent live browser test is a real current failure: `/waivers` fetched populated faabAnalytics (mean 1.37, analyzed=1102) but the `League average bid` tile was absent after 30s. Current code places `FaabHeaderStat` only inside the `selectedTeam` branch of `ManualAddDrop`; the ephemeral verification guest has no persisted team, so league-level analytics can load while the strip remains hidden. This is a concrete L4 UI gap, not missing data.
- V1-61: NOT VERIFIED. Authenticated Sharp read still timed out; Lane-4 `/api/sharp/market` also timed out. Do not interpret the dedicated 401/unmeasured Sharp smoke as evidence.
- V1-131: NOT VERIFIED. Desktop capability/menu/reachability checks passed, but the 390x844 mobile assertion still failed: after tabbar Menu click `.shell-drawer-group` never became visible within 15s.
- V1-102: backend configurable-expiry half passed with a real non-default 2h guest pass (expected 7200s, observed 7198s), but this does not satisfy the `/admin` L4 half and must not promote the row.

No status is being inflated from the workflow's overall step conclusions: the final job failed (`api=1 lane4=1`), and only row-specific measured evidence above should be used.

PR #1163, opened 2026-08-28
fix(v1-49): repair BASH_SOURCE unbound-variable crash in the SSH invocation path

jasonleetucker-code:
## Summary

The first live `action: activate` dispatch of the V1-49 controlled-activation workflow (run [33168603289](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/33168603289), 2026-08-28T11:49Z) refused cleanly at the preflight step — no production write happened, exactly as designed. But the actual cause was not a stale `expected_sha`, it was a real bug:

```
bash: line 570: BASH_SOURCE[0]: unbound variable
```

`deploy/diagnostics/v1_49_host_native_activation.sh`'s sourced-vs-executed guard (`if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then main "$@"; fi`) was modeled on `deploy/rollback.sh`'s identical-looking pattern — but that script is always invoked as `bash /path/to/rollback.sh` (a named file). This script is instead piped over SSH via `ssh ... "bash -s" < deploy/diagnostics/v1_49_host_native_activation.sh`, and bash reading a script from stdin with no named file pushes nothing onto `BASH_SOURCE` at all. Under `set -u`, the bare form therefore threw "unbound variable" before `main()` ever ran — **every real dispatch would have failed at that line regardless of ACTION or inputs.**

The bug was invisible to `bash -n` (syntax-valid — referencing an unset variable is not a parse error) and to the existing sourced-file unit tests (`source /real/path/to/script.sh` populates `BASH_SOURCE` correctly, so sourcing never exercised the stdin-piped branch that's actually used in production).

## Fix

`${BASH_SOURCE[0]:-${0}}` — falls back to `$0` when `BASH_SOURCE` has no entry, making the comparison trivially true for the stdin-piped shape (so `main()` always runs there, which is correct — that invocation is never a "sourced for testing" case) while leaving the sourced-for-testing case unchanged (`BASH_SOURCE[0]` is the sourced file's real path there, which does not equal `$0`).

Verified against all three invocation shapes locally before pushing:
1. `bash -s < script` (the real SSH shape) — `main()` now runs ✓
2. `bash script.sh` (direct execution) — `main()` runs ✓ (already worked)
3. `source script.sh` from another script (the test harness shape) — `main()` does NOT auto-run ✓ (already worked)

## Test plan

- [x] Reproduced the original crash locally (`bash -s < script` under `set -u`) before writing the fix
- [x] Added `test_stdin_piped_invocation_runs_main_without_unbound_variable_crash` — pipes the real script via stdin into `bash -s` (the exact production shape) and asserts `main()` reaches its own dispatch logic rather than crashing
- [x] Mutation-verified: reverted to the bare `${BASH_SOURCE[0]}` form and confirmed the new test's underlying reproduction still crashes identically
- [x] `bash -n` syntax check clean
- [x] `pytest tests/deploy/ tests/scripts/` — 712 passed, 4 skipped (no regressions)
- [x] `ruff format --check` / `ruff check` clean
- [x] `python3 scripts/ci_gate_classification.py` — 0 unclassified, 0 stale (no new workflow steps, classification untouched)
- [ ] Exact-head PR CI green (pending)

No production write happened during the failed attempt — preflight refused before any state was captured or mutated. V1-49 remains `IMPLEMENTED_UNVERIFIED`; the controlled activation will be re-dispatched against a freshly-confirmed deployed SHA once this fix lands and deploys.

---

### DIFF
--- a/deploy/diagnostics/v1_49_host_native_activation.sh
+++ b/deploy/diagnostics/v1_49_host_native_activation.sh
@@ -563,10 +563,24 @@ main() {
   esac
 }
 
-# Run main only when EXECUTED, not when sourced — mirrors
-# deploy/rollback.sh's own guard. Sourcing is how a test can call
-# individual functions (e.g. validate_activation_id) in isolation
-# without dispatching a real preflight/activate/rollback.
-if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+# Run main only when EXECUTED, not when sourced. Sourcing is how a
+# test can call individual functions (e.g. validate_activation_id) in
+# isolation without dispatching a real preflight/activate/rollback.
+#
+# ``${BASH_SOURCE[0]:-${0}}`` rather than deploy/rollback.sh's bare
+# ``${BASH_SOURCE[0]}``: that script is always invoked as
+# ``bash /path/to/rollback.sh`` (a named file), which populates
+# BASH_SOURCE. This one is piped over SSH via
+# ``ssh ... "bash -s" < deploy/diagnostics/v1_49_host_native_activation.sh``
+# — bash reading a script from stdin with no named file does not push
+# anything onto BASH_SOURCE at all, so under `set -u` the bare form
+# throws "BASH_SOURCE[0]: unbound variable" before main() ever runs,
+# and every real dispatch fails at this line. The default-expansion
+# form falls back to ``$0`` when BASH_SOURCE has no entry, making the
+# comparison trivially true (never sourced in that shape, so main()
+# is always the right call) while leaving the sourced-for-testing case
+# unchanged (BASH_SOURCE[0] is the sourced file's real path there,
+# which does not equal $0).
+if [[ "${BASH_SOURCE[0]:-${0}}" == "${0}" ]]; then
   main "$@"
 fi
--- a/tests/deploy/test_v1_49_workflow_wiring.py
+++ b/tests/deploy/test_v1_49_workflow_wiring.py
@@ -26,6 +26,7 @@ flag/backup logic lives in ``tests/scripts/test_prod_env_flag_ops.py``,
 
 from __future__ import annotations
 
+import os
 import re
 import subprocess
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


jasonleetucker-code:
ACTIVE V1 traffic-control update (2026-08-28): the repair is now genuinely DEPLOYED. Deploy Production run 33170783655 completed successfully on merge SHA `454fc15074ddb2aebff203d38526f037339784cc`; both `Validate Build Inputs` and the actual `Deploy To Production` job passed, including remote deploy, post-deploy smoke, and live data-contract validation.

This clears the deployment precondition for a fresh V1-49 controlled activation. It does **not** verify V1-49 by itself. The only activation run currently visible remains run 33168603289, which failed in read-only preflight before mutation. Do not re-use that old-head run as evidence.

Next legitimate V1-49 action: dispatch a fresh `V1-49 Host-Native Scoring — Controlled Activation` against the exact deployed SHA required by the workflow, then preserve the literal remaining validation evidence (BDVM challenger rerun, challenger league comparison/live Sleeper input, historical backtests, and rebuilt PBP artifact/join + measured movement). Keep V1-49 `IMPLEMENTED_UNVERIFIED` unless the full recipe passes; rollback on activation/health/validation failure.

jasonleetucker-code:
ACTIVE V1 TRAFFIC-CONTROL — 2026-08-28: controlled activation run 33176834694 reached the real mutation step but FAILED before the flag flip. Preflight passed at deployed SHA 454fc15074ddb2aebff203d38526f037339784cc. Failure root cause is a second production-invocation defect: `env_sourced_python()` shell-sources the systemd EnvironmentFile `.env`; production line 108 is valid/accepted by the service's EnvironmentFile semantics but is not shell-source-safe, producing `command not found` (exit 127). The failure occurred immediately after the PBP backup command, before `prior_flag_state.txt` was captured / before the flag was changed. Automatic rollback then attempted service restart but `/api/status` did not recover within its check and reported ROLLBACK_FAILED / MANUAL INTERVENTION REQUIRED. V1-49 remains IMPLEMENTED_UNVERIFIED; no promotion. I triggered a bounded re-run of the already-green Deploy To Production job from run 33170769974 to restore/reconfirm service health at exact deployed SHA 454fc15. Do not redispatch activation until the EnvironmentFile-vs-shell-sourcing defect is repaired and production health is reconfirmed.

PR #1164, opened 2026-08-28
fix(v1-49): stop sourcing production .env as bash, guard empty env-args

jasonleetucker-code:
## Incident

The second live V1-49 activation attempt (run `33176834694`, 2026-08-28) crashed with exit 127 during `Backup, activate, rebuild, measure`:

```
.env: line 108: c503b850537d737346e1f321b73ed2c9d889e0cde5636bf008985dfa1c17feaf: command not found
[v1-49][ERROR] activation step failed (exit 127)
```

Production's real `.env` contains a line that is not valid `KEY=VALUE` bash syntax (a bare 64-hex-char string, believed to be a wrapped/partial secret). `env_sourced_python()` was `source`-ing that file directly to load flags into the python subshell's environment, and bash tried to execute the malformed line as a command.

The automatic in-script rollback fired as designed and the `.env` flag was never touched (confirmed by the rollback log's own "no captured prior flag state — flag was never changed this attempt" line). Production health was independently confirmed via `/api/status` shortly after — never degraded by this run.

## Root cause and fix

`source`-ing arbitrary, not-fully-controlled `.env` content as bash is unsafe on its own terms, and it was never a faithful match to how production actually loads this file: `EnvironmentFile=` in the systemd unit (`deploy/systemd/dynasty.service.template:29`) is a plain `KEY=VALUE` reader, never a shell interpreter.

`env_sourced_python()` now parses `.env` as plain `KEY=VALUE` text (`env_sourced_python_env_args`) and passes the result to `env`, never to bash `source`. Malformed lines are silently skipped — never executed, never echoed (a value there could be a partial secret).

**Honest caveat**: an isolated local repro of "source this exact malformed line under `set -Eeuo pipefail`" did not reproduce exit 127 (bash tolerated the line and `source` returned 0) — the precise mechanics that turned it into a 127 in the real SSH session are not fully pinned down. The fix removes the risk class regardless of the exact trigger.

## Second bug caught by the new test

The new regression test (`test_env_file_is_never_sourced_as_bash`) caught a second, previously-latent bug on its first run against the fix: `printf '%s\n' "${args[@]}"` on a truly empty array still executes the format once, emitting one blank line rather than none. The caller's `mapfile` turned that into a single empty-string element, so `env ""` failed with `env: '': No such file or directory` whenever `.env` was absent or contributed zero usable lines (`test_env_sourced_python_tolerates_a_missing_env_file` caught this directly). Fixed with an explicit array-length guard before the `printf` call.

## Testing

- `bash -n deploy/diagnostics/v1_49_host_native_activation.sh` — clean
- New tests (`test_env_file_is_never_sourced_as_bash`, `test_env_sourced_python_env_args_skips_malformed_lines_silently`, `test_env_sourced_python_tolerates_a_missing_env_file`) — all pass, and were verified to actually catch both bugs before the fixes (each failed against the pre-fix / partially-fixed code first)
- `tests/deploy/ tests/scripts/` full suite: 715 passed, 4 skipped, 0 regressions
- `ruff format --check` / `ruff check` — clean
- `python3 scripts/check_planning_integrity.py` — OK
- `python3 scripts/ci_gate_classification.py` — 0 unclassified, 0 stale, 0 mismatches (no new workflow steps)

## Scope

Deliberately narrow — the one bounded repair for this incident. Two smaller follow-up items were identified but intentionally NOT bundled in here (tracked in the session's own working notes, not blocking):
- The rollback's post-restart health check (`sleep 3` then one `curl`) raced a real-but-transient connection failure during the second attempt's automatic rollback, even though production was independently confirmed healthy seconds later.
- The second activation run's "Safety-net rollback" GitHub Actions step showed `skipped` rather than a definite outcome — needs a direct read of the `if:` condition against Actions' `outcome` semantics to confirm this is expected (in-script rollback already handled it) rather than a gap.

---

### DIFF
--- a/deploy/diagnostics/v1_49_host_native_activation.sh
+++ b/deploy/diagnostics/v1_49_host_native_activation.sh
@@ -163,11 +163,51 @@ resolve_systemctl_bin() {
 }
 
 # ── small python helpers, run against the SAME .env every production
-#    process would source (EnvironmentFile=-__APP_DIR__/.env) ─────────
+#    process would read (EnvironmentFile=-__APP_DIR__/.env) ───────────
+#
+# Reads .env as plain KEY=VALUE lines (env_sourced_python_env_args),
+# never via bash `source`. First live activation run (2026-08-28,
+# run 33176834694) hit a real, deterministic production incident: the
+# box's actual .env contains a line that is not valid KEY=VALUE bash
+# syntax, and `source`-ing it derailed the run before any write
+# happened (a safe outcome, but not a usable one — every dispatch
+# would fail identically). `EnvironmentFile=` in the systemd unit that
+# actually loads this file is itself a plain KEY=VALUE reader, not a
+# shell interpreter, so a strict line parser here is the FAITHFUL
+# match to how production reads its own .env — bash `source` was
+# always a stricter, riskier stand-in than the real mechanism.
+# Malformed lines are silently skipped (a value there could be a
+# partial secret — never echoed), never executed as a command.
+env_sourced_python_env_args() {
+  local -a args=()
+  local line key value
+  if [[ -f "${ENV_FILE}" ]]; then
+    while IFS= read -r line || [[ -n "${line}" ]]; do
+      [[ "${line}" =~ ^[[:space:]]*(#.*)?$ ]] && continue
+      if [[ "${line}" =~ ^([A-Za-z_][A-Za-z0-9_]*)=(.*)$ ]]; then
+        key="${BASH_REMATCH[1]}"
+        value="${BASH_REMATCH[2]}"
+        args+=("${key}=${value}")
+      fi
+    done < "${ENV_FILE}"
+  fi
+  # `printf '%s\n' "${args[@]}"` on a truly empty array still executes
+  # the format once, substituting "" for the missing operand — one
+  # blank line, not zero. mapfile below then reads that as a single
+  # empty-string element, so `env ""` gets passed a bogus argument it
+  # tries to run as a command ("env: '': No such file or directory").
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
ACTIVE V1 traffic-control audit: this is the current highest-value V1-49 repair and the scope is appropriately bounded to the production `.env` loading defect exposed by activation run 33176834694. The diff preserves the production EnvironmentFile semantics by parsing KEY=VALUE input rather than shell-sourcing it, skips malformed lines rather than executing/echoing them, and adds regression coverage for malformed and empty/missing env input. Integration gate remains exact-head PR Validation on `73fbff572c32f37578eb27acb516c9ba4c1e4e95`; do not merge or redispatch activation until that run is green. V1-49 remains IMPLEMENTED_UNVERIFIED regardless of merge/deploy; a fresh controlled activation and the literal promotion-gate evidence are still required.

jasonleetucker-code:
ACTIVE V1 traffic-control update (2026-08-28): #1164 is merged at `42ce5b2021cad2fb04b623b2f02c8fb16379c459`. Exact PR head `738b9f6c98c9382e3082937612b6c812f2558b35` passed PR Validation run 33187091001. Deploy Production run 33188701111 has both jobs green: `Validate Build Inputs` succeeded and `Deploy To Production` succeeded through remote deploy, post-deploy smoke, and live data-contract validation. This clears the deploy precondition for a FRESH V1-49 controlled activation using the repaired EnvironmentFile-compatible path. It does NOT verify V1-49 by itself; V1-49 remains IMPLEMENTED_UNVERIFIED until a fresh activation and the contract's required production evidence/false-green checks actually pass. Do not reuse/rerun the old failed activation against the pre-fix workflow tree.

PR #1175, opened 2026-08-29
fix(sharp): batch the four holdings_as_of() baseline reads into one pass

jasonleetucker-code:
## Summary

- V1-61's connection-reuse fix (#1169) landed and was confirmed deployed, but a fresh authenticated production run (dispatch after ~30 min of warm-up, no concurrent deploy/scrape) still hit the client's 60s timeout on `/api/sharp/roster-percentage` (`V61A: TimeoutError: The read operation timed out`). Connection reuse alone wasn't enough — `build_board()` still ran four full, independent scans over `sharp_roster_observations` and `sharp_roster_asset_spans`, one per baseline window (7/14/30-day + season-to-date).
- Add `holdings_as_of_multi()`: reads both tables **once** for the full roster population (unfiltered by any `as_of` cutoff), then computes each baseline's holdings in Python against that shared data. Semantically identical to calling `holdings_as_of()` once per timestamp — a roster's earliest observation already answers "observed by X" for every X at once, and a span's open/closed shape doesn't depend on which timestamp is asked.
- `build_board()` now calls `holdings_as_of_multi()` once instead of `holdings_as_of()` four times, still on the shared connection from #1169.

## Test plan

- [x] `test_holdings_as_of_multi_matches_per_call_output` (new `TestHoldingsAsOfMulti` class) — a direct equivalence test: batched output must equal calling `holdings_as_of()` once per timestamp on fixture data covering not-yet-observed rosters (absent, never coerced to empty), an open span with no closing observation, and a span closed by a later contradicting observation.
- [x] Mutation-verified: broke the closed-span exclusion logic, confirmed the equivalence assertion fails for the right reason, restored the fix.
- [x] `tests/sharp/` — 324/324 passing.
- [x] `ruff check` / `ruff format --check` clean on changed files.
- [x] `python3 scripts/check_planning_integrity.py` — same single pre-existing failure present on `origin/main` before this branch (an unrelated PR's doc-governance gap), confirmed via `git stash` diff; not touched by this change.
- [ ] Real production timing recipe against `/api/sharp/roster-percentage` post-deploy (V1-61's row stays `IMPLEMENTED_UNVERIFIED` until this is re-run).

---

### DIFF
--- a/docs/PLANNING_DOCUMENT_STATUS.md
+++ b/docs/PLANNING_DOCUMENT_STATUS.md
@@ -64,6 +64,15 @@ has no destination in the Scope Manifest or the traceability document.
 Detailed owner-approved behaviour. Binding within their subject; they do not authorize implementation.
 
 ### Trade
+- `docs/OWNER_FEATURE_ADDENDUM_2026-08-29_BEST_BALL_ROSTER_UTILITY.md` — **owner-approved scope, issue #1173,
+  owner decision 2026-08-29.** A roster-conditional dynasty best-ball utility layer for Analyze Trade
+  (`F(R) = E[optimal legal weekly lineup score | R, exact league rules]`; Best-Ball Lineup Impact, Lineup Entry
+  Probability, Usable Production/Redundancy, Depth Insurance Value, Roster Spot Shadow Value,
+  Consolidation/Diversification classification) that must reuse canonical exact lineup/assignment machinery and
+  must never overwrite standalone canonical player value. Per the addendum's own §8, reconciliation into
+  `docs/OWNER_REQUESTED_TODO.md`, `docs/OWNER_FEATURE_INVENTORY.md`, the scope manifest and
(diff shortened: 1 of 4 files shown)


jasonleetucker-code:
CI red on the first push was the "Planning integrity gate" — `docs/OWNER_FEATURE_ADDENDUM_2026-08-29_BEST_BALL_ROSTER_UTILITY.md` (added by unrelated PR #1174) has no entry in `docs/PLANNING_DOCUMENT_STATUS.md`, tripping the governance-index check. Not this PR's: confirmed the identical failure reproduces on a clean checkout of `origin/main` via `python3 scripts/check_planning_integrity.py`, with none of this PR's changes applied.

Opened a standalone fix (#1176) and ported the identical change onto this branch (`014fffe`) so this PR doesn't have to wait on it to merge — it will no-op once #1176 lands on `main`.

---

PR #1185, opened 2026-08-30
perf(sharp): hoist the championship shrinkage base out of the loop (V1-61)

jasonleetucker-code:
## Where V1-61 stands, measured

Lane 4 profiler, `dynasty_main`, same box each time:

| | pre-#1183 | post-#1183 | post-#1184 (`afdb0d6`, run **33307244404**) |
|---|---|---|---|
| `build_board` wallMs | 290,184.7 | 128,065.8 | **84,125.7** |
| `cohort_members` | 204,883.9 | 44,271.8 | 47,352.5 |
| ↳ `score_managers` | n/i | 32,025.4 | **34,117.7** |
| row loop (uninstrumented) | n/i | 57,401.8 | **28,593.0** |
| `load_rosters` | 20,284.3 | 14,648.1 | 3,390.6 |

Down **71%** from baseline. #1184 halved the row loop as designed. But
84.1s is still over the bar — the authenticated consumer (run
**33307023491**, deployed `afdb0d6`) still records:

```
V61A       V1-61   unmeasurable
V61A:crash   -     error   TimeoutError: The read operation timed out
```

**V1-61 stays `IMPLEMENTED_UNVERIFIED`. This PR promotes nothing.**
`score_managers` is now the dominant stage.

## Root cause — profiled, not guessed

Local `cProfile` at the production population:

```
63601 calls   25.0s   score.py:199(<listcomp>)      <- _mean
162,016,665           {built-in method builtins.isinstance}
```

**92% of `score_managers`.** And 162,016,665 = 12,720 × 12,720 to within
0.2% — one `_mean` over the whole evaluable population, per manager.

`_performance_component` re-derived `base` (the championship shrinkage
target) on every call. `_mean` rebuilds its input list with an
`isinstance` test per element, and that list holds one entry per
evaluable manager. Same O(N²) shape #1183 fixed for the percentile
lookups, in the same function — missed the first time.

## The fix

`base` is a property of the **population**, not of the manager being
scored. `score_managers` computes it once and passes it in.
`_performance_component` still derives it when `championship_base` is
`None`, so ad-hoc and historical callers keep working; the hot path
always passes it.

**Measured, N=45,000: 11.9s → 0.67s (17.2×)**, with byte-identical
`ManagerScore` output verified against the real pre-fix code via `git
stash` at N=3,000 and N=45,000. Projected on production: ~34.1s → ~2s,
wall **~84.1s → ~52s** — which would be under the 60s bar, though that
is a projection and only a real production run can settle it.

## One subtlety is load-bearing, and the equivalence check caught it

The base must be averaged from **`sorted_population`, not
`population`**. They hold the same multiset, but floating-point addition
is not associative, so the two orders give slightly different means, and
that difference propagates through `_shrunk_rate` into a percentile
lookup.

My first version of this change sourced it from `population`. That moved
manager `u556`'s performance component `0.2433 → 0.2186`, his score
`47.4 → 46.3`, and changed **12,720 of 45,000 rows**. It was caught only
because the before/after comparison runs at production scale — the unit
suite passed it clean.

(`build_population`'s own `observed_base` has always summed the raw list;
it is a separate quantity on a different axis and is untouched.)

## Tests — both mutations verified

* **base is derived from the list the component actually receives** —
  replaces the raw population with a summation-order-sensitive one and
  captures the value actually handed to `_performance_component`. Catches
  **both** sourcing it from the unsorted list **and** dropping the hoist.
  An ordinary test population cannot expose this: at N=150 both orders
  agree to the last bit, which is exactly why my first version of this
  test passed under mutation and had to be rewritten.
* hoisted base matches deriving it per manager, over the whole output.
* an empty population still falls back to the `0.08` prior, not a coerced
  zero.

`tests/sharp/` **338 passed**. `ruff check`/`ruff format` clean.
`check_planning_integrity.py` OK.

## After this

If the projection holds, the next production run is the one that decides
V1-61. If it lands short, the remaining ranked levers are
`build_manager_records` (13.1s) and the residual row loop (28.6s).

---

### DIFF
--- a/src/sharp/score.py
+++ b/src/sharp/score.py
@@ -309,12 +309,28 @@ def _performance_component(
     rec: ManagerRecord,
     population: dict[str, list[float]],
     cfg: dict[str, Any],
+    championship_base: float | None = None,
 ) -> tuple[float, list[str]]:
     """``population`` must be pre-sorted, pre-cleaned float arrays (see
     ``score_managers``'s ``sorted_population`` — every lookup here goes
     through ``_percentile_rank_sorted``, not ``percentile_rank``, because
     this runs once per manager and a fresh O(N) scan per call was the
-    dominant cost of ``score_managers`` in production (V1-61)."""
+    dominant cost of ``score_managers`` in production (V1-61).
+
+    ``championship_base`` is the population's mean championship rate — the
+    shrinkage target. It is a property of the POPULATION, not of ``rec``,
+    so ``score_managers`` computes it once and passes it in. Left as
+    ``None`` it is derived here exactly as before, which keeps ad-hoc and
+    historical callers working; the hot path must always pass it.
+
+    Deriving it per manager was the dominant cost of ``score_managers``
+    even after the percentile repair: ``_mean`` rebuilds its input list
+    with an ``isinstance`` test per element, and the list is one entry per
+    evaluable manager, so calling it once per manager is O(N^2). Measured
+    at the production population (12,720 evaluable of 45,286 evidence
+    managers): 162,016,665 ``isinstance`` calls — 92% of the function's
+    runtime, and 12,720 x 12,720 to within 0.2%.
+    """
     block = cfg.get("performance") or {}
     sub = block.get("subWeights") or {}
     notes: list[str] = []
@@ -338,7 +354,11 @@ def _performance_component(
             )
 
     # Championships: shrunk, because one title in one season is luck.
-    base = _mean(population.get("championshipRate") or []) or 0.08
+    base = (
+        championship_base
+        if championship_base is not None
+        else (_mean(population.get("championshipRate") or []) or 0.08)
+    )
     prior_n = float((block.get("championshipShrinkage") or {}).get("priorN", 6.0))
     shrunk = _shrunk_rate(rec.championships, rec.completed_seasons, base, prior_n)
     p = _percentile_rank_sorted(shrunk, population.get("championshipRateShrunk") or [])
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
ACTIVE V1 traffic-control: exact-head PR Validation run 33307971897 is RED, so this PR must not merge yet. The blocking non-livedata failure is `TestChampionshipBaseHoist::test_base_is_derived_from_the_list_the_component_actually_receives`: its fixture `[1e16] + [1.0] * 200` no longer separates raw vs sorted `_mean()` under the CI Python 3.12 runtime (`from_raw == from_sorted == 49751243781095.52`). This is a test-fixture portability failure, not evidence against the optimization. Repair the test without weakening what it proves: it must still fail if championship_base is sourced from raw `population` instead of the exact `sorted_population` passed to `_performance_component`, and it must fail if the hoist is removed. Prefer a deterministic sentinel/spy approach rather than relying on floating-point addition order. Re-run exact-head PR Validation and do not merge until the hard unit-test gate is green. The livedata KTC drift and stale `flockFantasySfRookies` source-health messages in the same run are advisory/non-blocking and unrelated; do not 'fix' them in this V1-61 PR.

jasonleetucker-code (commented):
Traffic-control note: head 3769587 fixes the Python 3.12 portability failure by replacing the floating-point-order fixture with deterministic _mean sentinels. Product code is unchanged by this follow-up. The file replacement also stripped explanatory comments/docstrings elsewhere in tests/sharp/test_score.py; those are non-functional, but this should be minimized/restored before merge so #1185 stays a bounded V1-61 change. Do not merge solely on CI green until that test-file diff is cleaned back to the intended guard-only edit.

PR #1187, opened 2026-08-30
docs(test): restore the explanatory prose stripped from test_score.py

jasonleetucker-code:
## Why

This closes your own review note on #1185:

> The file replacement also stripped explanatory comments/docstrings
> elsewhere in `tests/sharp/test_score.py`; those are non-functional, but
> this should be minimized/restored before merge so #1185 stays a bounded
> V1-61 change.

#1185 merged before that happened. This does it.

The strip came from commit `3769587` — the fixup that **correctly**
replaced my non-portable float-ordering fixture with deterministic `_mean`
sentinels. (CPython 3.12 gave `sum()` Neumaier compensated summation, so
`[1e16] + [1.0] * 200` separated raw from sorted `_mean` on 3.11 but not on
CI's 3.12. My fixture, my bug.) **That repair is kept exactly as written** —
only the collateral prose loss is undone.

## What was actually lost — measured, not assumed

* `test_score.py` on `main` carries **0 comments**. Before #1185 it carried **37**.
* **6 docstrings** present before #1185 were gone from `main`.
* The losses were **not confined to this session's work**. Also gone were
  pre-existing notes that predate it:
  * why the `methodologyVersion` assertion reads the config rather than a
    literal — *"hardcoding `sharp-v2` made every legitimate version bump
    fail here, which turns the config's own rule into an obstacle instead
    of a guard"*
  * why the v2 cohort-tier test asserts an **ordering** rather than a
    brittle exact count
  * the keeper-league and win-rate-floor annotations

## How it was reconstructed

Mechanically, not rewritten from memory: the pre-#1185 file, with #1185's
`TestChampionshipBaseHoist` class appended **verbatim** from `main`.

## Proven safe by AST comparison, not by reading

* **Executable code is identical to current `main`** with docstrings
  stripped — so no test body, assertion, or fixture changed, and the only
  difference between the two files was ever prose plus that one new class.
* **0** docstrings present on `main` are missing here.
* **6** docstrings that `main` had lost are back; **37** comments restored.

`tests/sharp/` **338 passed**. `ruff check` and `ruff format` clean.
`check_planning_integrity.py` OK. **No production code touched**, no V1 row
promoted.

---


jasonleetucker-code:
**Merge-order constraint — do not merge this before #1189.**

#1189 repairs a live scoring regression in `src/sharp/score.py` (introduced by #1183: the championship shrinkage base was sourced from `sorted_population`, desynchronizing it from `build_population`'s `observed_base` by one ULP and breaking the `championshipRateShrunk` tie — measured 8 `qualified` flips, i.e. sharp cohort membership moved).

Part of that repair is flipping the assertion in `TestChampionshipBaseHoist::test_base_is_derived_from_the_list_the_component_actually_receives` from `{sorted_sentinel}` to `{raw_sentinel}` — that test currently pins the regression as intended behaviour, which is why nothing caught it.

This PR reconstructs `tests/sharp/test_score.py` by appending `TestChampionshipBaseHoist` **verbatim from `main`**, i.e. the pre-#1189 version. Merging it after #1189 would revert that assertion and silently re-pin the regression.

So: land #1189 first, then I'll rebuild this branch on the new `main` and re-verify by AST that executable code is unchanged. Nothing else about this PR changes — it stays docs-only, no production code, no row promoted.

---

jasonleetucker-code:
**Rebuilt on the new `main` — the ordering constraint above is now discharged.** #1189 landed (`b6559ac`), so this branch has merged current `main` and no longer carries the pre-#1189 `TestChampionshipBaseHoist`.

Re-verified after the merge, same AST method as the original:

- **executable code identical to current `main`** with docstrings stripped → no test body, assertion, or fixture changed
- **0** docstrings present on `main` are missing here
- **6** docstrings restored; **38** comments (`main` carried 0)
- #1189's repaired assertion is intact — `assert set(seen) == {raw_sentinel}` — and its new `test_a_managers_shrunk_rate_lands_on_its_own_population_entry` is present
- diff vs `main` is `tests/sharp/test_score.py` only, **+91 / −1**

`tests/sharp/` **342 passed**. `ruff format --check` and `ruff check` clean.

Still docs-only: no production code, no V1 row promoted.

---

PR #1199, opened 2026-09-01
V1-62: fix the W15-F017 cohort memo's self-poisoning write

jasonleetucker-code:
## Summary

V1-62 (Sharp Tracker, required L4) has repeatedly failed authenticated verification because `/api/sharp/market` times out — the same symptom class V1-61's `/api/sharp/roster-percentage` had before its own fix. Before implementing anything, I profiled the real endpoint (`scripts/profile_sharp_market.py`, mirroring V1-61's `profile_sharp_roster_percentage.py`) per the standing "profile before implementing" directive, starting from a documented candidate hypothesis (an `ffpc_config` cache-key mismatch between `market.py` and `roster_percentage.py`). Profiling surfaced something more fundamental than that hypothesis.

## Root cause

The W15-F017 cohort memo (`src/sharp/cohort.py::cohort_members`) keys its cache on `(qualification, ledger_path, ffpc_config_signal)` and validates freshness against the ledger sqlite file's `(mtime_ns, size)`. But `_compute_cohort_members` calls `curated_industry_members` unconditionally on **every** cohort build, which reaches `curated.ensure_schema` — and that function ran an **unconditional** `INSERT OR REPLACE INTO meta(...)` + `commit()` on every single call, writing to the *exact* ledger file whose mtime is the memo's own freshness signal.

So building a cohort invalidated the cache entry it was about to store, before the next caller — even the *same request's* second `cohort_members` call, or a call microseconds later with an identical key — could ever see a hit. Verified directly with a counting wrapper around `_compute_cohort_members`: with the old code, every single `cohort_members` call triggered a full rebuild; with the fix, repeat calls against an unchanged ledger hit the cache.

This means the memo added by W15-F017 (#1129) has been contributing close to nothing in production since it landed — every cohort build was silently poisoning its own cache.

## Fixes

1. **`src/sharp/curated.py::ensure_schema`** — the version-stamp write is now conditional on the stored value actually differing from `SCHEMA_VERSION`. The steady-state case (every call after the first-ever migration) becomes a pure read with zero writes. Migration semantics are unchanged — the version marker ends at `SCHEMA_VERSION` either way.
2. **`src/sharp/market.py::market_payload`** — with the memo now actually working, its two `cohort_members` call sites are fixed to pass the *original* `ffpc_config` argument (`None` by default) instead of the pre-resolved config dict. `roster_percentage.py` never passes `ffpc_config` either, so both now key the memo identically (`"file"`) and can share one cohort build across requests/endpoints — previously `market.py`'s resolved-dict argument produced a different key (`"cfg:<sha1>"`) that never matched, even though `_compute_cohort_members` resolves `None` the same way internally. `audit_payload` already passed the original argument and needed no change.

## Evidence

- `scripts/profile_sharp_market.py` (new, read-only diagnostic, no product behavior change) — before the fix, `cohort_compute` fired on every `cohort_members` call; after, repeat calls in one process share a single build.
- Two new regression tests in `tests/sharp/test_cohort_memo.py`, both **mutation-verified** (confirmed RED against the reverted code, GREEN on the fix):
  - `test_ensure_schema_does_not_touch_ledger_once_version_is_current`
  - `test_curated_industry_lookup_does_not_poison_the_cohort_memo` (end-to-end, through the real `cohort_members` → `curated_industry_members` path)
  - Plus a mutation-control test reproducing the retired unconditional write directly, showing it breaks the fingerprint on its own.
- `tests/sharp/` — 345/345 pass (up from 307 pinned at W15-F017's original close).
- `ruff check` / `ruff format --check` clean on all changed files.
- `python3 scripts/check_planning_integrity.py` — clean (this PR does not touch the V1 contract).

## Out of scope, noted honestly

While investigating, I found `curated_industry_members` doesn't accept or forward a `ledger_path` argument — it always resolves the default ledger internally. This is a separate, narrower, pre-existing characteristic (every current caller uses the default ledger anyway, so it isn't causing any observed defect) and is not touched by this PR.

## What this PR does NOT do

This does not promote V1-62 to `VERIFIED`. The row's required L4 evidence is an authenticated production pass against the deployed fix — that's the next step after this merges and deploys.

## Test plan

- [x] `tests/sharp/` 345/345 pass
- [x] Both new tests mutation-verified (RED on revert, GREEN on fix)
- [x] `ruff check` / `ruff format --check` clean
- [x] `check_planning_integrity.py` clean
- [ ] CI green on this PR
- [ ] Post-merge: authenticated verification pass against `/api/sharp/market` on the deployed fix

---

### DIFF
--- a/scripts/profile_sharp_market.py
+++ b/scripts/profile_sharp_market.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python3
+"""Read-only stage profiler for V1-62 Sharp Tracker (``/api/sharp/market``).
+
+This script does not implement a second market engine and does not change
+any product result. It calls the canonical ``src.sharp.market.market_payload``
+function exactly as the HTTP route (``src/sharp/service.py::get_market``)
+calls it -- default ``ffpc_config=None`` -- while temporarily wrapping its
+existing stage owners with timers, then prints a JSON timing report.
+
+Purpose: V1-62's required L4 evidence has repeatedly shown
+``/api/sharp/market`` timing out on the remote authenticated-verification
+path while passing on-box, the same symptom class V1-61 had before its
+``/api/sharp/roster-percentage`` fix.
+
+Profiling with this script surfaced something more fundamental than the
+``ffpc_config`` cache-key mismatch it was written to test (that mismatch is
+real too, and both are now fixed): the W15-F017 cohort memo
+(``src/sharp/cohort.py``) was silently self-poisoning on EVERY call,
+including two calls made microseconds apart with an identical key.
+``_compute_cohort_members`` calls ``curated_industry_members`` on every
+build, which reaches ``curated.ensure_schema`` -- and that function
+performed an UNCONDITIONAL ``INSERT OR REPLACE INTO meta(...)`` + commit
+on every single call, writing to the exact ledger sqlite file whose
+``(mtime_ns, size)`` is the memo's only freshness signal. So the very act
+of building a cohort invalidated the cache entry it was about to store,
+before the next caller could ever see a hit -- inside ``market_payload``
+itself (two ``cohort_members`` calls per request, line 394-ish and
+517-ish, never shared) as well as across requests and across endpoints.
+Fixed in ``src/sharp/curated.py::ensure_schema`` by making the version
+stamp write conditional on the value actually needing to change. With
+that fixed, the ``ffpc_config`` key-mismatch fix in ``market.py`` (passing
+the ORIGINAL argument to ``cohort_members`` instead of the pre-resolved
+dict) lets ``market.py`` and ``roster_percentage.py`` share one cohort
+build across endpoints, as W15-F017 originally intended.
+
+``--simulate-shared-memo`` demonstrates both effects together: after the
+fix, repeat ``market_payload`` calls in one process do ONE cohort compute
+(previously N), and warming the memo under ``roster_percentage``'s own
+call shape (no explicit ``ffpc_config``) is now visible to a subsequent
+``market_payload`` call too.
+
+Truthfulness rules (same as ``profile_sharp_roster_percentage.py``):
+- missing data remains missing; no values are substituted;
+- no auth or feature flags are changed;
+- no database writes are performed by this script itself;
+- the canonical implementation is executed exactly as production calls it;
+- a timeout/error is reported as such, never as an empty or healthy result.
+"""
+
+from __future__ import annotations
+
+import argparse
+import functools
+import json
+import signal
+import sys
+import time
+from collections import OrderedDict
+from pathlib import Path
+from typing import Any, Callable
+
+# Same sys.path fix as profile_sharp_roster_percentage.py -- this script is
+# invoked as ``python scripts/profile_sharp_market.py``, which puts
+# ``scripts/`` rather than the repo root at sys.path[0].
+_REPO_ROOT = Path(__file__).resolve().parents[1]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+from src.intel import platform_ledger  # noqa: E402
+from src.sharp import cohort as sharp_cohort  # noqa: E402
+from src.sharp import consensus as sharp_consensus  # noqa: E402
+from src.sharp import market as sharp_market  # noqa: E402
+from src.sharp import platform_records as sharp_platform_records  # noqa: E402
+from src.sharp import score as sharp_score  # noqa: E402
+
+
+class ProfileTimeout(RuntimeError):
+    """Raised when the bounded profiler wall-clock budget is exhausted."""
+
+
+class StageRecorder:
+    def __init__(self) -> None:
+        self._stats: OrderedDict[str, dict[str, float | int]] = OrderedDict()
+        self._restores: list[tuple[Any, str, Callable[..., Any]]] = []
+
+    def wrap(self, owner: Any, attr: str, label: str) -> None:
+        original = getattr(owner, attr)
+        stats = self._stats.setdefault(
+            label,
+            {"calls": 0, "totalMs": 0.0, "maxMs": 0.0},
+        )
+
+        @functools.wraps(original)
+        def timed(*args: Any, **kwargs: Any) -> Any:
+            started = time.perf_counter()
+            try:
+                return original(*args, **kwargs)
+            finally:
+                elapsed_ms = (time.perf_counter() - started) * 1000.0
+                stats["calls"] = int(stats["calls"]) + 1
+                stats["totalMs"] = float(stats["totalMs"]) + elapsed_ms
+                stats["maxMs"] = max(float(stats["maxMs"]), elapsed_ms)
+
+        self._restores.append((owner, attr, original))
+        setattr(owner, attr, timed)
+
+    def restore(self) -> None:
+        while self._restores:
+            owner, attr, original = self._restores.pop()
+            setattr(owner, attr, original)
+
+    def report(self) -> dict[str, dict[str, float | int]]:
+        return {
+            label: {
+                "calls": int(stats["calls"]),
+                "totalMs": round(float(stats["totalMs"]), 3),
+                "maxMs": round(float(stats["maxMs"]), 3),
+            }
+            for label, stats in self._stats.items()
+        }
+
+
+def _install_stage_wrappers(recorder: StageRecorder) -> None:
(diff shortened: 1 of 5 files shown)


jasonleetucker-code:
CI run [33454889816](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/33454889816) failed, but none of the three failures are in code this PR touches (`src/sharp/curated.py`, `src/sharp/market.py`, `tests/sharp/test_cohort_memo.py`, `scripts/profile_sharp_market.py`). All three reproduce identically on plain `main` at `0553df8` (checked out clean, no PR changes applied):

1. **`tests/canonical/test_ktc_reconciliation.py::TestKTCReconciliation::test_rank_delta_within_tolerance[12-8561-12.3-3.0]`** — `assert 3.7813559322033896 <= 3.0`. This is a pinned-tolerance test comparing today's live-scraped KTC board against a historical pinned percentage; it drifts as KTC's own board moves day to day and is independent of any code here.
2. **`tests/deploy/test_backup_root_resolution.py::test_an_interleaved_non_generation_entry_costs_no_real_generation`** — `assert 13 == 14`. A backup-rotation test whose expected date window depends on the current date; unrelated to sharp/cohort code.
3. **`scripts/check_source_health.py`** — `flockFantasySfRookies last fetched 2026-08-26T16:07:55+00:00 (128.5h > 24.0h)`. A source-freshness check on an external scraper feed, also independent of this PR's diff.

No fix exists yet for any of the three (they're data-drift/date-window conditions, not a breaking commit to revert) — I'm not widening this PR to try to re-pin the KTC tolerance or backup-window test, since that's unrelated methodology work with its own blast radius. Since all three reproduced identically and deterministically on unmodified `main` (not flaky), I'm not spending the one allowed re-run on them — re-running against the same data snapshot would reproduce the same result.

---

---

jasonleetucker-code:
Correction/addendum to my previous comment, for accuracy:

- **`check_source_health.py`'s `flockFantasySfRookies` staleness did NOT fail the job** — that step correctly carries `continue-on-error: true` in `pr-validation.yml` (line 313), matching the documented "advisory, loud always" intent. I mis-listed it as one of the failure causes; it's loud but non-blocking, exactly as designed. (Separately, real: that source has been stale 127h+ across every scheduled refresh since 2026-08-26, tracked in #1161 — unrelated to this PR, not something I'm touching.)
- **`test_an_interleaved_non_generation_entry_costs_no_real_generation` is NOT a live-date-drift test** — I described it that way without fully checking; it actually uses fixed literal dates (`2026-08-01` through `2026-08-14`) and calls the real `deploy/backup/riskit-state-backup.sh` with a fixed `KEEP_DAILY=14`. It's deterministic, not date-sensitive. It still reproduces identically and deterministically on unmodified `main` (verified), so it's still not caused by this PR — I just had the wrong reason for why.

The only two things actually gating this job's conclusion are the two pytest failures (`test_ktc_reconciliation` and `test_backup_root_resolution`), both confirmed pre-existing on `main`. Standing-down rationale from my previous comment otherwise holds.

---

---

jasonleetucker-code:
ACTIVE V1 TRAFFIC CONTROL — exact-head CI blocker is independently confirmed deterministic and unrelated to the Sharp change. Do **not** rerun unchanged and do **not** weaken `KEEP_DAILY=14`.

Blocking test: `tests/deploy/test_backup_root_resolution.py::test_an_interleaved_non_generation_entry_costs_no_real_generation`.

The fixture seeds 14 legitimate Aug 1–14 generations, then the shipped writer creates the pinned current `TODAY` generation. Production policy retains the newest **14 legitimate dated generations total**, so the correct post-run result is 13 seeded survivors + the newly promoted generation = 14. The current assertion `len(survivors) == 14` counts only the seeded August set and therefore demands 15 legitimate generations total after the writer runs.

Required repair is test-only and must preserve production semantics. The corrected invariant should prove all three facts:
1. exactly 14 legitimate dated generations remain after the writer runs;
2. the oldest seeded generation is the one pruned while the other 13 seeded generations survive (so the scratch entry did not consume an extra slot);
3. `2026-08-07-restore-scratch` still exists and is excluded rather than deleted.

After that repair: rerun the focused deploy test, then require fresh exact-head PR Validation + E2E Safety Net green before merge. V1-62 remains `IMPLEMENTED_UNVERIFIED (L4)` until the fix is merged, deployed, and `/market/sharp-tracker` is proven consuming the canonical Sharp path in authenticated production evidence.

jasonleetucker-code:
Pushed `9f74f81` implementing the repair exactly as specified.

`DATE_STAMP=TODAY` pins the script to this module's single clock read (matching `_run_proof`'s existing pattern), and the corrected assertion proves all three facts:
1. exactly 14 legitimate dated generations survive (`survivors == expected`, where `expected` = 13 surviving seeded dates + `TODAY`)
2. `2026-08-01` (the oldest) is specifically the one pruned — asserted explicitly, not just implied by a count
3. `2026-08-07-restore-scratch` still exists — unchanged

**Mutation-verified**, not just made to pass: I temporarily loosened `riskit-state-backup.sh`'s stray-entry recognition (dropped the shape-regex anchor and the date round-trip check — the two independent guards that currently correctly exclude the scratch entry) and reran. The rewritten test correctly went RED (`prune: .../2026-08-02` — an extra real generation evicted, the exact damage the docstring describes), then GREEN again with the script restored untouched. `git diff --stat` on the script is empty; only the test changed.

`tests/deploy/` — 402 passed, 4 skipped. `ruff check` clean on the changed file (skipped `ruff format` — it would reformat ~15 unrelated pre-existing lines elsewhere in the file under this environment's ruff version, matching the known local/CI ruff-version mismatch noted in prior commits; not touching unrelated code).

**Still outstanding**: `tests/canonical/test_ktc_reconciliation.py::TestKTCReconciliation::test_rank_delta_within_tolerance[12-8561-12.3-3.0]` remains red (pinned +12.3% vs actual +16.1% at rank 12, ±3.0pp band) — this wasn't addressed by your comment and I haven't touched it; it's canonical-valuation methodology (re-pin the tolerance or re-fit the curve) outside what I'd unilaterally decide. Let me know if you'd like me to investigate that one too, or if it's being handled elsewhere.

---

---

jasonleetucker-code (commented):
Traffic-control review: the retention invariant repair is conceptually correct, but exact-head PR Validation 33467554887 now fails deterministically at `ruff format --check` on `tests/deploy/test_backup_root_resolution.py` before the hard unit gate. Please format only that test file on this branch (the visible long `assert "2026-08-01" not in survivors, ...` line is the obvious formatter delta), then require a fresh exact-head PR Validation run. Do not change KEEP_DAILY, product retention behavior, Sharp logic, or the repaired test invariant to satisfy this gate. The livedata KTC drift and flockFantasy source-health findings in that run are advisory and are not the blocking verdict.

jasonleetucker-code:
Pushed `c3ef23c` — `ruff format` applied to `tests/deploy/test_backup_root_resolution.py` only, formatting-only (assert-message line-wrapping convention). Confirmed via diff review: no changes to `KEEP_DAILY`, production retention behavior, Sharp logic, or the repaired test invariant — only how the existing assert messages wrap. `ruff check` / `ruff format --check` now clean on all 5 files this PR touches; `tests/deploy/test_backup_root_resolution.py` still 49 passed / 4 skipped.

Ready for a fresh exact-head PR Validation run on `c3ef23c`.

---

---

jasonleetucker-code:
My previous push (`c3ef23c`) didn't actually fix it — found and fixed the real cause now.

This environment has **two** `ruff` installs: the bare `ruff` on PATH resolves to `/root/.local/bin/ruff` at **0.15.8** (newer), while this repo pins `ruff~=0.6.0` and CI installs/runs **0.6.9** via `python -m ruff format --check .`. The two versions disagree on this file's assert-message wrapping style, so my previous commit reformatted to 0.15.8's opinion — which is why the exact-head run on `c3ef23c` still failed with `Would reformat: tests/deploy/test_backup_root_resolution.py`, identically to before.

Pushed `d8fcb46`, reformatted with `python3 -m ruff` (confirmed `--version` reports 0.6.9, resolving to the pip-installed package, matching CI). Verified: `python3 -m ruff format --check` on all 5 files this PR touches now reports "already formatted"; `python3 -m ruff check` clean; `tests/deploy/test_backup_root_resolution.py` 49 passed / 4 skipped. Still formatting-only — confirmed via diff review, no change to `KEEP_DAILY`, retention behavior, Sharp logic, or the repaired invariant.

Noting for future PRs on this branch/session: I'll use `python3 -m ruff` (not the bare `ruff` command) for all lint/format checks going forward to avoid repeating this.

---

---

PR #1225, opened 2026-09-03
feat(v1-onbox): add v20 check — C1-SRC-02 fails-closed, against the deployed tree

jasonleetucker-code:
## Summary

V1-20's row is honest that it sits at L2, not the required L3, for one specific reason: production offers no live `UNKNOWN` feed to refuse (the real registry is all-`DYNASTY`), so the fails-closed half of `C1-SRC-02`'s gate has never been observed as a live negative case — and the row correctly refuses to round that up to L3.

That's a real limit on *observing* a negative case in production traffic. It is not a limit on running the deployed *code* against an injected one. `tests/sources/test_game_type_gate_red.py` already pins this property in this dev/CI tree (`_validate_source_game_types_invariant` refuses `UNKNOWN`/`REDRAFT`/absent, accepts verified-`DYNASTY`) via its injectable `sources` parameter — built for exactly this. What was missing is running that same injectable-rogue check against the code actually running in **production**, closing the gap between "the fails-closed behavior exists somewhere in this repo" and "the fails-closed behavior is live on the deployed SHA" — the L3 bar, without fabricating a bad feed anywhere in production.

## Changes

- `scripts/verify_v1_onbox.py`: new `check_v20()`, following the existing `check_v11` pattern (imports the deployed `src.api.data_contract` module over SSH, exercises the gate read-only via the module's own injectable `sources` param — no mutation of the live registry).
- Wired into `CHECK_FNS` and `.github/workflows/v1-onbox-checklists.yml`'s `checks` input description.

## Test plan

- [x] `python3 -m pytest tests/ops/ tests/deploy/ tests/sources/test_game_type_gate_red.py` — 606 passed, 4 skipped
- [x] Mutation-verified: monkeypatched a permissive (never-raises) validator and confirmed the check correctly flips to `fail` rather than a false pass on a broken gate
- [x] `ruff check` / `ruff format --check` — clean
- [x] `python3 scripts/verify_v1_onbox.py --checks v20` run locally against this tree's own (real, all-`DYNASTY`) registry — `pass`, exit 0

Once merged and deployed, dispatching `v1-onbox-checklists.yml` with `checks=v20` gives V1-20 the L3 evidence its row is currently missing.

---

### DIFF
--- a/.github/workflows/v1-onbox-checklists.yml
+++ b/.github/workflows/v1-onbox-checklists.yml
@@ -3,7 +3,8 @@ name: V1 On-Box Checklists (read-only by default)
 # Executable form of V1 production checklists that exist as prose and whose
 # inputs are prod-only:
 #   V1-89  (DraftSharks §6),  V1-15/16/17 (C1-U8 §8),  V1-59 (journal chain),
-#   V1-83  (F-20 delivery-keyed cooldown),  V1-11 item 1,  V1-124 (job matrix).
+#   V1-83  (F-20 delivery-keyed cooldown),  V1-11 item 1,  V1-124 (job matrix),
+#   V1-20  (C1-SRC-02 fails-closed half, against the deployed tree).
 #
 # Same SSH mechanism, permissions, concurrency and buffering posture as
 # lane4-onbox-verification.yml (see its header for why -u and tee-without-
@@ -20,7 +21,7 @@ on:
   workflow_dispatch:
     inputs:
       checks:
-        description: "Comma list of check groups (v89,c1u8,v59,v83,v11,v124) or 'all'"
+        description: "Comma list of check groups (v89,c1u8,v59,v83,v11,v124,v20) or 'all'"
         required: false
         default: "all"
         type: string
--- a/scripts/verify_v1_onbox.py
+++ b/scripts/verify_v1_onbox.py
@@ -1,4 +1,4 @@
-"""On-box V1 production checklists — V1-89, V1-15/16/17, V1-59, V1-83, V1-11, V1-124.
+"""On-box V1 production checklists — V1-89, V1-15/16/17, V1-59, V1-83, V1-11, V1-124, V1-20.
 
 Runs ON the production box (dispatched by
 ``.github/workflows/v1-onbox-checklists.yml`` over the existing SSH lane),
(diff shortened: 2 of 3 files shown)


jasonleetucker-code:
CI red is not this PR's — reproduced identically against `main` with zero local changes:

- `tests/canonical/test_ktc_reconciliation.py::TestKTCReconciliation::test_rank_delta_within_tolerance[12-8561-12.3-3.0]` and `[24-7242-13.5-3.0]` — KTC's own live-scraped values have drifted outside the pinned tolerance band (our curve's output is unchanged/pinned exactly; only KTC's board moved: rank 12 pinned +12.3% vs actual +16.1%, rank 24 pinned +13.5% vs actual +9.8%). This test's own comment says as much: "KTC's curve wiggles with their daily scrape" — periodic re-pinning is expected maintenance, not a defect this PR introduced. This PR touches only `scripts/verify_v1_onbox.py`, its test file, and the on-box workflow YAML — nothing in the valuation pipeline.
- `check_source_health.py`: `flockFantasySfRookies last fetched ... (179.0h > 24.0h)` — a source-health staleness advisory, also unrelated to this diff.

No fix exists yet for the KTC drift (it needs a deliberate re-pin against a fresh KTC scrape, which is a separate, unrelated change I don't want to fold into this PR). Re-running once now in case a newer scheduled data refresh has already narrowed the drift; if it's still red after that, the block is external to this PR and I'll keep it watched rather than force anything through.

---

jasonleetucker-code:
Correction to my last comment — I misread the log the first time. The KTC reconciliation failures I flagged are real (reproduce on `main`), but they run under the `pytest -m livedata` step, which carries `continue-on-error: true` in `pr-validation.yml` — that step does **not** actually gate the job, by design (it's exactly why `test_ktc_reconciliation.py` is listed in `tests/conftest.py`'s `_LIVEDATA_MODULES`). Not this PR's problem, and not actually the blocker either.

The real blocker was mine: the **`Python format gate (ruff format --check)`** step flagged `tests/ops/test_v1_verification_primitives.py`. Root cause — this environment had `ruff 0.15.8` shadowing the repo-pinned `ruff~=0.6.0` (`requirements-dev.txt`) at `~/.local/bin/ruff`. My earlier `ruff format` calls used the wrong (much newer) version, which reformatted the *whole file* to 0.15.8's opinion — including two pre-existing, untouched functions whose multi-line `assert` wrapping 0.6.9 renders differently — not just the new V1-20 tests this PR adds.

Fixed by installing the pinned version (`pip install ruff~=0.6.0`, invoked as `python3 -m ruff` since the standalone binary still shadowed the pip one) and reformatting with that. The diff against `origin/main` is now exactly the two new V1-20 test functions, nothing else. Pushed as `8382127`.

---

PR #1296, opened 2026-09-08
Harness: make Ruff formatting deterministic for every coding agent

jasonleetucker-code:
## Why

Ruff formatting has repeatedly become a false-progress loop: an agent hand-edits wrapping based on intuition, CI says Ruff would format it differently, and the next prompt has to rediscover the exact style.

The fix is to make formatter behavior executable shared harness state rather than prompt folklore.

## Changes

- pin the exact CI formatter version: `ruff==0.6.9`
- add the model-neutral pre-push command:
  `python scripts/format_changed_python.py`
- that helper auto-detects branch/staged/unstaged/**untracked** Python files, runs `ruff format`, then `ruff format --check`, `ruff check`, and `git diff --check`
- add the explicit universal rule to `AI_INSTRUCTIONS.md`: **run Ruff; never imitate Ruff**
- surface the command in `scripts/agent_session_start.sh` so Claude, Codex, ChatGPT, Copilot, Gemini, and future agents see it at startup
- add a harness regression test pinning the whole contract

The actual formatting rules remain owned by `pyproject.toml`; prompts do not duplicate line-wrapping conventions.

Agent-OS-Receipt: `1d065ab778c1671c0c43dff8f088149fb3843944`

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -107,6 +107,14 @@ jobs:
           python -m py_compile server.py "Dynasty Scraper.py"
           python -m py_compile src/api/data_contract.py
 
+      - name: TEMP exact Ruff diff for Agent OS formatting test
+        shell: bash
+        run: |
+          set -Eeuo pipefail
+          python -m ruff format tests/docs/test_agent_operating_system.py
+          git diff -- tests/docs/test_agent_operating_system.py
+          exit 1
+
       - name: Python format gate (ruff format --check)
         shell: bash
         run: |
--- a/AI_INSTRUCTIONS.md
+++ b/AI_INSTRUCTIONS.md
@@ -35,6 +35,18 @@ Carry:
 
 into the first material checkpoint, material work-claim/PR handoff, and final handoff.
 
+## Python formatting is executable, not prose
+
+For any material Python edit, **do not hand-format or guess what Ruff will do**. The repository's formatting contract is the formatter itself:
+
+- exact tool version: `ruff==0.6.9` from `requirements-dev.txt`;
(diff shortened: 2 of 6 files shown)


jasonleetucker-code:
Closing as **absorbed** — every capability this PR proposes is already live on `main`, shipped independently under a different filename:

- Pinned Ruff version (`ruff==0.6.9`) — `requirements-dev.txt`.
- The model-neutral pre-push formatter command — `scripts/format_python_changes.sh` (this PR proposed `scripts/format_changed_python.py`; same job, different name — verified this PR's own base commit `50c0d75` does not contain `format_python_changes.sh`, so `main` grew it independently after this PR branched).
- The explicit "run Ruff, never imitate Ruff" rule in `AI_INSTRUCTIONS.md` — present under "## Python formatting contract".
- Surfaced at agent-session start — `scripts/agent_session_start.sh` prints the exact command.
- A harness regression test referencing the contract — `tests/docs/test_cross_model_agent_entrypoints.py`.

Verified directly this session by running `scripts/format_python_changes.sh` repeatedly against another PR's changes — it works exactly as this PR's description specifies (auto-detects changed files, runs `ruff format` + `ruff format --check` + `ruff check`, fails loud on drift).

Merging this PR now would create a second, competing formatter-contract script — the "ONE CONCEPT, ONE CANONICAL OWNER" invariant this repo already enforces elsewhere. Not a rejection of the idea; the idea already shipped, just not from this branch. If anything here still adds value on top of the current contract (e.g. a check this PR's test covers that `test_cross_model_agent_entrypoints.py` doesn't), reopen with a diff scoped to just that gap.

---

PR #1307, opened 2026-09-09
Week 1: run existing production proof instruments after deploy

jasonleetucker-code:
Temporary operational bridge for the final Week 1 evidence rows. No new verification logic is introduced.

- Adds `workflow_run` on successful `Deploy Production` to the existing V1 authenticated production verification workflow so W1-30 can be proven against the freshly deployed tree.
- Adds the same successful-deploy trigger to the existing retention backup+restore proof so merged #1302 can emit the required `C5-GD-02 game_day/: restored N file(s)` evidence for W1-03.
- Manual `workflow_dispatch` remains unchanged/canonical.
- On workflow-run events, V1 inputs use the existing defaults (`dynasty_main`, all halves, 2h guest pass); retention forces `RUN_BACKUP=1`.
- Both job guards refuse to run on failed deploys or non-main upstream runs.

This trigger is intentionally temporary and should be removed after W1-03 and W1-30 evidence is recorded. The tests/scripts themselves are unchanged.

PR #1310, opened 2026-09-09
Week 1: production proof triggers on current main

jasonleetucker-code:
Fresh current-main re-derivation of #1307 after the deploy-timeout repair changed the upstream `Deploy Production` workflow while #1307 was validating. The #1307 implementation itself was green (PR Validation run 34336261828); this replacement ports the identical two workflow changes onto current main so the final merge candidate includes the now-30-minute deploy workflow. No verification/test logic changed. Supersedes #1307 when green and merged.

### DIFF
--- a/.github/workflows/retention-backup-proof.yml
+++ b/.github/workflows/retention-backup-proof.yml
@@ -25,6 +25,14 @@ name: Retention backup + restore proof
 # manager name or roster reaches this log.
 
 on:
+  # TEMP Week-1 proof bridge: run this existing restore/verify instrument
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Integration sequencing note: once this PR is green and merged, freeze unrelated merges until its `Deploy Production` run completes and the new successful-deploy `workflow_run` triggers have emitted the retention and authenticated-production proof. This prevents a later push from cancelling the exact deployment needed for W1-03/W1-30 evidence.

PR #1311, opened 2026-09-09
chore(deps): current-main frontend refresh with Next 16.3.4

jasonleetucker-code:
Fresh current-main re-derivation of stale Dependabot #1264. The frontend package files have not changed since #1264's base, so this ports its exact dependency/lock refresh onto current main rather than merging a 190+ commit-old branch. Includes Next 16.3.4 plus the grouped testing-library/user-event/vite-plugin minor/patch updates. This supersedes #1264 and #1300 if green. Vitest 5 (#1299) is intentionally not folded into this PR because it is a major-version migration. Nanoid #1298 remains a separate lock-only patch until this lock refresh is adjudicated.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -9,15 +9,15 @@
       "version": "1.0.0",
       "dependencies": {
         "html2canvas": "^1.4.1",
-        "next": "16.2.12",
+        "next": "16.3.4",
         "react": "19.2.8",
         "react-dom": "19.2.8"
       },
       "devDependencies": {
-        "@testing-library/jest-dom": "^7.0.0",
-        "@testing-library/react": "^16.1.0",
-        "@testing-library/user-event": "^14.5.2",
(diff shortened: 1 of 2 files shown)


jasonleetucker-code:
Final base-movement triage before merge:

- validated head: `55e52dec8c5bec6883a0d15129cf2c143a2c47ed`
- validation base: `ba2b4e782836f1a58088d3b15d2092b70287fc9f`
- current main: `b33ed1cb3df8901147a9feca9a82c00ad7384dbd`
- intervening main files: README/docs/FAAB feature-flag files plus `data/ops/sharp-production-smoke.json`
- overlap with this PR's `frontend/package.json` / `frontend/package-lock.json`: none
- classification: `BENIGN_AUTOMATION_MOVE` / non-overlapping independent feature movement for this dependency refresh.

Exact-head PR Validation and E2E Safety Net are both green, so no freshness-only CI restart is required.

PR #1373, opened 2026-09-15
fix(w1-28): make a completed week's FINAL Game Day reachable

jasonleetucker-code:
## Summary

W1-28's acceptance is a **FINAL** Game Day: *"FINAL state is production-usable and preserves final optimal lineup/results plus clean transition/linkage to the canonical recap system."*

The backend defects behind it are fixed and merged (#1367, refined by #1371). But the row still could not be proven, and this PR is about why.

**`/game-day` could only ever render whichever week the HOST clock is on.** `/api/matchup/intel` has always accepted `season` and `week` — both documented optional, falling back to the host — while `GameDayPanel` forwarded neither. So the moment the host rolls forward, the completed week's FINAL state (final score, final optimal lineup, result, recap link) becomes unreachable through the UI for **everyone**, owner and verifier alike. It is not a stale view: the host itself says Week 1 is over, and Week 1's result is a fact that does not expire.

## Measured, not hypothesised

Verification run [76](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/34991789313) ran against `410fc5a` — the tree carrying both W1-28 backend fixes — and its `w1-16-game-day.spec.js` annotations read:

```
w1-16-state  = SCHEDULED/pregame
w1-27-mode   = pregame
w1-16-week   = 2026 week 2
```

There is **no `w1-28-branch` annotation at all**, because the spec's FINAL branch never executed. Sleeper's own state endpoint agrees: `season 2026, week 2`. The specs passed (77 passed / 2 failed / 25 skipped) — they simply proved a pregame, which is not what W1-28 asks for.

Worth stating plainly: the `/api/matchup/intel` HTTP 500 is **gone** in that run. Game Day rendered on both desktop and mobile, `w1-16-endpoint-status = 200`, `w1-16-anon-api = HTTP 401`. The two failures in that run are the already-tracked #1157 and #1340, neither of which is Game Day.

## Change

**`frontend/components/GameDayPanel.jsx`** forwards `week`/`season` exactly the way it already forwards `team`, and for the reason that field's own comment gives — the endpoint supports it, the page did not, and without it the surface cannot be verified for anyone.

Values are passed **verbatim** rather than validated client-side: the endpoint's `_int_param` already treats an unparseable value as *absent* and falls back to the host clock, so re-deciding that in the client is how the page and the endpoint would start disagreeing about which week is on screen.

[...]

Issue #1340, opened 2026-09-11
Team Strength page stuck on "Loading league data..." past 60s in production


jasonleetucker-code:
Observed as an incidental failure in a W1-27 LIVE-evidence verification run (workflow_dispatch, `v1-authenticated-verification.yml`, run [34548330201](https://github.com/jasonleetucker-code/riskittogetthebrisket/actions/runs/34548330201), 2026-09-11 ~01:10 UTC), unrelated to that run's actual purpose — filing so it isn't lost.

**Failure:** `tests/e2e/specs/prod-auth/v1-123-team-strength.spec.js` — "V1-123 Phase 2: Team Strength/Weakness tab (production) > ROS roster strength renders a real per-team table or an explicit not-ready state on production" (prod-mobile project).

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('table').or(getByText(/ROS data not ready/i)).or(getByText(/ROS roster strength unavailable/i)).first()
Expected: visible
Timeout: 60000ms
Error: element(s) not found
```

Page snapshot at timeout shows the page still rendering `paragraph: Loading league data...` — neither the real table nor either of the two explicit not-ready states ever appeared within 60s. This is a genuine production behavior gap: the test's own locator already accepts a truthful "not ready" state as a pass, so this isn't a test being too strict — the page never reached *any* of the three expected terminal states.

Not diagnosed further (out of scope for the run that surfaced it). Worth a trace-level look at whether this is a slow backend response, a stuck client-side fetch, or a genuinely hung request — same category of investigation as #1157's roster-percentage timeout, but a different endpoint/page.

Full trace/video/screenshot are in the workflow run's `v1-authenticated-report` artifact if still available (90-day retention).

jasonleetucker-code:
Automation follow-up diagnosis from current `main` (`6054b4a650a0272c23a579cabbec06e41d95c6ff`), using repository source only; no paid services and no code change yet.

The observed indefinite `Loading league data...` state has a concrete source-level path:

- `tabs.js::SECTION_FOR_TAB` intentionally omits `rosTeamStrength` because that tab is self-fetching, so a deep link to `?tab=rosTeamStrength` makes `page.jsx::sectionForTab()` return `null` and SSR requests only `overview`.
- If that bounded 3s server fetch yields no usable league payload, `page.jsx` passes `initialContract=null`.
[...]

Issue #1157, opened 2026-08-27
V1-61: roster-percentage production timeout is the current engineering blocker


jasonleetucker-code:
Live V1 traffic-control finding, 2026-08-27.

The current contract no longer has an authority/session blocker for V1-61. Fresh authenticated production verification run 33098047781 minted a legitimate guest pass, logged in, and completed the API/browser suites, but the row still failed because `/api/sharp/roster-percentage` itself times out (>60s). PR #1155 already corrected the contract to reflect this current state.

Important separation:
- #1150 optimized `src/sharp/service.py::cohort_status()` / `/api/sharp/cohort`.
- The remaining failing path is `src/sharp/roster_percentage.py::build_board()`.
- Do NOT merely increase the timeout or treat timeout as empty data.

Bounded next engineering action:
1. Profile `build_board()` and its direct expensive dependencies on a production-like dataset.
2. Identify duplicate SQLite scans / repeated cohort or roster reconstruction / repeated market or contract joins.
3. Reuse existing canonical cohort/roster owners and established memoization/materialization patterns; do not create a second Sharp methodology.
4. Preserve missing != zero, stale != current, truthful degraded states, and auth.
5. Add targeted timing/regression coverage that catches the >60s path without brittle micro-benchmarks.
6. Merge/deploy only after exact-head CI.
7. Re-run the authenticated V1 production recipe; promote only on a genuinely measured, non-timeout result.

The dedicated anonymous Sharp smoke's 401/unmeasured green is not V1 evidence and should not be used to close this row.

jasonleetucker-code:
[...]

PR #1374, opened 2026-09-15
fix(power): separate canonical history from the results-only diagnostic

jasonleetucker-code:
## Summary

Three defects in how past Power weeks are **displayed** and how movement arrows are **derived**. The canonical 40/20/15/15/10 blend and its early-season ramp are untouched — no weighting, ROS, all-play, form or VORP calculation changes here.

### 1. The reconstruction wore the canonical label

Historical weeks render `trend.weeks`, which the backend builds with `results_only=True`. That renormalizes onto all-play + recent + record, and at week 1 the result is close to a single-week points sort — served under a bare **"Power Rankings"** heading. That is why it read as a standings table.

Now: the diagnostic is explicitly titled (`Power Rankings — diagnostic (results-only), 2026 Wk 1`), its week-selector option is suffixed `· diagnostic`, the canonical blend summary is suppressed while it is on screen, and the formula line finally reads **that week's own** `effectiveWeights` instead of the current week's. Relabelled and isolated — not removed.

### 2. Movement crossed season boundaries

`trend.weeks` chains every tracked season onto one sequential list (`week_states` loops over `seasons_sorted`), so `weekDelta()` walking backward from a season's first week landed in the **previous season's** results-only ranking — a different league state, and a standings-derived one. This is the Eric ▲8 / Jason ▲7 / Kich ▼6 class of arrow.

It now stops at the season boundary and returns `null`. A season's first week has no in-season predecessor, and "no movement" is the honest answer, not a number borrowed from last year.

### 3. Week 0 was structurally unpublishable

`snapshot_path()` raised on `week < 1`, so Week 1 could never have a legitimate baseline. Week 0 is now the **preseason publication**, stamped `preseason: true`, and `movement_against_previous()` short-circuits at `<= 0` instead of `<= 1`. The lookup stays scoped to `(same season, week-1)`, so the canonical path still cannot reach across a season.

## Ranking composition

Each row now shows `ROS #2 · All-play #1 · Last 4 #4 · Record #6` plus the week's effective weights. A raw percentile doesn't answer "which of these five put me here?" — #1 of 12 and #9 of 12 can sit a few points apart.

Derived in `power_v2._attach_component_ranks`, **not** the page: these are ordinals over a population, which `CLAUDE.md` forbids the frontend to compute. The page renders them verbatim. A `None` component is **unranked**, never sorted last — a team nobody measured is not the worst team at it. They're frozen into the published snapshot with everything else, so a past week's sub-ranks can't be silently restated against a changed roster.

## VORP/PAR — TODO only, no behavior change

`team_vorp` stays unavailable with its 15% redistributed. That is correct interim behavior (scoring every team 0 for a quantity nobody measures would punish them equally and still be fabricated), now recorded as the blend's one open dependency with a named `TODO` rather than left to be rediscovered.

## Deliberately NOT done

**No 2026 Week 0 snapshot is back-filled.** Week 1 is already published with `rankDelta: null`; minting a Week 0 now would restate what an immutable publication means after the fact — the exact thing this store exists to prevent. Whether to publish one is your call, and it would be stamped with today's `finalizedAt`, not a preseason date.

Worth noting: the live canonical 2026 Wk 1 ranking is **Brent 1, Joey 2, Eric 3, Jason 4, Collin 5, MaKayla 6, Kich 7, Blaine 8, Ed 9, Ty 10, jstuedle 11, Roy 12** — exactly the "last official publication" list in the task. So the baseline described there is the real published one.

## Test plan

- [x] **Owner acceptance fixture passes exactly** — 12-team Wk0→Wk1: Eric ▲2, Joey —, Jason ▲1, MaKayla ▲2, Brent ▼4, Ed ▲3, Collin ▼2, Blaine —, jstuedle ▲2, Kich ▼3, Ty ▼1, Roy —. Plus a companion test proving the 10-team results-only series (Brent 1, Joey 2, Ed 3, Kich 4, …) can **never** drive those arrows — Eric is the discriminator at +2 vs +8.
- [x] Season-boundary regression: a new season's first week reports no movement rather than comparing to last season's finale.
- [x] `tests/ros` — **449 passing + 9 new** (4 snapshot/Week-0, 5 componentRanks).
- [x] `npx vitest run` — 170 files, **2473 passed**.
- [x] `npm run build` — clean, all 14 budgets under (`/league` 41.8 / 50 KB).
- [x] Ruff 0.6.9 format + lint green, **no rewrites**.
- [x] Decision-path coercion gate: `clean: no new coercions`.
- [x] End-to-end against the live `dynasty_main` snapshot: all 12 owners carry `componentRanks`, `team_vorp` correctly absent, effective weights show the ramp (ROS 75.1% at one scored game).

Agent-OS-Receipt: af1d50a577c96fd9eed9f934a902a9469f8b69bc

https://claude.ai/code/session_01DvuXzRycU7b6dShZF7iNiU

---
_Generated by [Claude Code](https://claude.ai/code/session_01DvuXzRycU7b6dShZF7iNiU)_


jasonleetucker-code:
Week-1 coordination note for this branch before it moves again: production evidence for the still-active W1-28 claim is now harvested from authenticated verification run 35015317637 attempt 2 (source SHA `36f8dd5a6baf9722ec1ea41e86d5c7f5ba90793a`, production carrying #1373 lineage). The dedicated historical completed-week test passed on desktop + mobile with `2026 W1`, `mode=final`, `result=WIN`, recap `/league/articles/2026/1 -> HTTP 200`, and the spec asserts rendered FINAL score + final optimal lineup + no remaining probabilities. W1-28 is therefore evidence-ready for `VERIFIED`.

The same one-pass result was 79 passed / 25 skipped / 2 unexpected; the only unexpected specs were exactly the two pre-existing `v1-123-*` defects already named in the Week-1 contract/runbook: sharp roster percentage (#1157) and Team Strength (#1340). Every W1 production spec passed. In this attempt #1157 occurred on desktop and #1340 on mobile, so preserve that observed project split rather than repeating the older descriptive “mobile-only” wording; the failing spec identities / underlying tracked defects are unchanged and neither maps to a Week-1 row.

Because this PR uses the same branch and edits `docs/WORK_CLAIMS.md`, I am deliberately not pushing the Week-1 contract/claim closeout into this unrelated Power PR. After #1374 is resolved, the next bounded unit should release/complete the W1-28 claim and promote W1-28 plus W1-30 in a dedicated closeout change if the canonical owner applies §7's named-exception clause as written, mechanically recounting to 30/30. Only after that literal merge should the 2026-09-08 Site Steward Phase-1 authorization activate.

Evidence was also recorded on merged PR #1373 in comment 5688454381.

